PR #865, opened 2022-07-06
Executor tests

ToniRamirezM:
### What does this PR do?

Add executor tests:

- TestExecuteTransaction: Still WIP, as Executor does not yet provide all fields needed
- TestExecutor: Implements a test vector against the executor to check for the expected new state root

### Reviewers

Main reviewers:

- @arnaubennassar  
- @tclemos


ToniRamirezM:
I am closing this PR as the executor is not yet able to properly pass all the tests and the branch is getting hard to maintain. It would be faster re-adding the tests from this branch (that I will keep) into the develop branch once the system is able to pass these tests.

PR #892, opened 2022-07-13
Prometheus endpoint with metrics

xavier-romero:
### What does this PR do?

This PR introduces a new `metrics` package that provides utilities to expose metrics to [Prometheus](https://prometheus.io/).

If metrics are enabled in configuration, a new `/metrics` endpoint is exposed through a separate http server started at the application level and listening on a dedicated port.

Each component that wants to expose metrics will need to:
- set up a `metrics` sub-package defining its own metric and the utility functions to update each metrics.
- register the metrics at component startup.

You can refer to these branches to see the implementation in [`jsonrpc`](https://github.com/0xPolygonHermez/zkevm-node/tree/prometheus-json-rpc) and [`sequencer`](https://github.com/0xPolygonHermez/zkevm-node/tree/prometheus-sequencer).

### Reviewers

Main reviewers:

- @arnaubennassar 
- @tclemos 

Codeowner reviewers:

- @arnaubennassar 
- @tclemos


jleebee:
@cla-bot check

docker-compose.yml
```diff
@@ -252,7 +256,7 @@ services:
 
   zkprover-mock:
     container_name: zkprover-mock
-    image: hermeznetwork/zkprover-mock:latest
```
KonradIT, line 255: Remember to switch to this image once you have access to dockerhub login

KonradIT (changes requested):

docker-compose.yml
kind84: good point @KonradIT, done here 2dab38e5f631e394800f1d48c8215d2539d4da8f

KonradIT (approved):
Excellent job.

cmd/run.go
```diff
@@ -102,6 +110,10 @@ func start(cliCtx *cli.Context) error {
 		}
 	}
 
+	if c.Metrics.Enabled {
+		go startMetricsHttpServer(c)
```
ARR552, line 114: I see that this will expose an endpoint where metric can be checked. Am I right? I thought metric would be sent to a central server where all the metrics would be collected. Why did you choose this polling strategy??

config/config.local.toml
```diff
@@ -109,3 +109,8 @@ URI = "zkevm-prover:50071"
 [BroadcastServer]
 Host = "0.0.0.0"
 Port = 61090
+
+[Metrics]
+Host = "0.0.0.0"
+Port = 9091
+Enabled = false
```
ARR552, line 116: Instead of having this enable flag, Why don't you just check if host and port cointain valid info?? In that case if fields are empty the metrics would be disabled

cmd/run.go
kind84: @ARR552 there's been quite a discussion about whether to push or pull metrics at the beginning of the implementation. The main concern was about the jsonrpc service that can scale horizontally.
Finally we decided together with @xavier-romero to keep the pulling strategy and have the scaling problem resolved on the metrics server side (through service discovery or other custom mechanism).

Prometheus scraper collects metrics by pulling that endpoint on services it is aware of. The other way around would have required to set up a metrics push-gateway and [it seems to not being recommended as a generic solution](https://prometheus.io/docs/instrumenting/pushing/).

config/config.local.toml
Psykepro: Combined all three fields as `URI`.

config/default.go
```diff
@@ -117,4 +117,9 @@ URI = "127.0.0.1:50071"
 [BroadcastServer]
 Host = "0.0.0.0"
 Port = 61090
+
+[Metrics]
+Host = "0.0.0.0"
+Port = 9091
+Enabled = false
```
ARR552, line 124:
```suggestion

[Metrics]
URI = ""
```

ARR552 (changes requested):

ToniRamirezM (approved):

tclemos (approved):

arnaubennassar (approved):
I'd like to test in on a "production" env before merging

config/default.go
kind84: @ARR552 we finally decided to stick with `Host`, `Port` and `Enabled` because we find it more explicit. Also it allows to turn metrics on/off without having to type the `URI` every time. This is helpful especially in this first stage where we are still testing the metrics. WDYT?

PR #918, opened 2022-07-20
Trusted state synchronization via Broadcast service

### LINKED ISSUE #891: Synchronize trusted state on permissionless sequencers
The trusted sequencer will broadcast all the changes in the trusted state, the Permissionless sequencers must synchronize this data in order to update their own Trusted State "copy".

### LINKED ISSUE #933: Update namespace for the RPC `hez` ==> `zkevm`

tclemos:
Closes #891.
Closes #933.

### What does this PR do?

adds `permissionless` containers to `docker-compose` to work as a `permissionless node`
adds new `Make` commands: `make run-permissionless` and `make stop-permissionless`
adds `Trusted State` synchronization for `Permissionless nodes` via `Broadcast` service
replaces the current JSON RPC Server `hez` namespace by `zkevm` namespace.
adds a new API to expose the broadcast URI via JSON RPC Server

Tested using `Metamask` manually and running `Test1000EthTransfer`.

### Reviewers

Main reviewers:

- @ARR552 - please check the synchronization loop and the synchronizer code
- @ToniRamirezM - please check the data synchronized to make sure the state is correct
- @fgimenez - please check the `Broadcast` integration
- @arnaubennassar - please check it 🙃


synchronizer/synchronizer.go
```diff
@@ -110,45 +130,39 @@ func (s *ClientSynchronizer) Sync() error {
...
+				log.Info("Trusted state fully synchronized")
+				waitDuration = s.cfg.SyncInterval.Duration
 			}
-			// Sync L2Blocks
-			// TODO
 		}
 	}
 }
 
+func initConn(serverAddress string) (*grpc.ClientConn, context.CancelFunc, error) {
```
fgimenez, line 157: This is very similar to what we have in the e2e test, could we refactor to a `broadcast.NewClient` method similar to what we have for executor and smt?

synchronizer/synchronizer.go
```diff
@@ -41,8 +47,21 @@ func NewSynchronizer(
 	genesis state.Genesis,
 	reorgBlockNumChan chan struct{},
 	cfg Config) (Synchronizer, error) {
-	ctx, cancel := context.WithCancel(context.Background())
+	var broadcastClient pb.BroadcastServiceClient
+
+	if cfg.TrustedSequencerURI != "" {
+		log.Infof("connecting to trusted sequencer broadcast service: %v", cfg.TrustedSequencerURI)
+		// connects to broadcast service to get updates from trusted sequencer
```
fgimenez, line 54: I think this comment is redundant given the above log, wdyt?

ToniRamirezM (approved):

synchronizer/synchronizer.go
```diff
@@ -682,3 +748,113 @@ func (s *ClientSynchronizer) processVerifiedBatch(verifiedBatch etherman.Verifie
...
+func (s *ClientSynchronizer) processTrustedBatch(trustedBatch *pb.GetBatchResponse, dbTx pgx.Tx) error {
+	log.Debugf("processing trusted batch: %v", trustedBatch.BatchNumber)
+	processCtx := state.ProcessingContext{
+		BatchNumber:    trustedBatch.BatchNumber,
+		Coinbase:       common.HexToAddress(trustedBatch.Sequencer),
+		Timestamp:      time.Unix(int64(trustedBatch.Timestamp), 0),
+		GlobalExitRoot: common.HexToHash(trustedBatch.GlobalExitRoot),
+	}
+	txs := []types.Transaction{}
+	for i := 0; i < len(trustedBatch.Transactions); i++ {
```
fgimenez, line 761: Could we use a `range` here?

fgimenez (approved):

config/config.debug.toml
```diff
@@ -25,6 +25,7 @@ SequencerNodeURI = ""
 [Synchronizer]
 SyncInterval = "5s"
 SyncChunkSize = 100
+TrustedSequencerURI = ""
```
ARR552, line 28: this param is not required because the trusted sequencer url must be read from the smc. I suggest to include a method in the etherman called ethMan.GetTrustedSequencerURL().
Here in the config file we can use a boolean to allow sync from the ts.

synchronizer/synchronizer.go
```diff
@@ -110,41 +121,26 @@ func (s *ClientSynchronizer) Sync() error {
...
-					log.Warn("error getting latest batch synced. Error: ", err)
+					log.Warn("error syncing trusted state. Error: ", err)
 					continue
 				}
-				if latestSyncedBatch == latestsequencedBatchNumber {
-					waitDuration = s.cfg.SyncInterval.Duration
-				}
-				if latestSyncedBatch > latestsequencedBatchNumber {
-					log.Fatal("error: latest Synced BatchNumber is higher than the latest Proposed BatchNumber in the rollup")
-				}
```
ARR552, line 144: Why do yu delete this check?? this situation will lead to a wrong synchronizarion and I think we should detected

synchronizer/synchronizer.go
```diff
@@ -110,41 +121,26 @@ func (s *ClientSynchronizer) Sync() error {
...
+			if err != nil {
+				log.Warn("error getting latest sequenced batch in the rollup. Error: ", err)
+				continue
+			}
+			latestSyncedBatch, err := s.state.GetLastBatchNumber(s.ctx, nil)
+			if err != nil {
+				log.Warn("error getting latest batch synced. Error: ", err)
+				continue
+			}
+			if latestSyncedBatch >= latestSequencedBatchNumber {
```
ARR552, line 134: I don't thing doing these operations all the time even if they are not necessary, is worth it. I suggest to keep the if  `			if waitDuration != s.cfg.SyncInterval.Duration {`

synchronizer/synchronizer.go
```diff
@@ -110,41 +121,26 @@ func (s *ClientSynchronizer) Sync() error {
...
+				log.Info("L1 state fully synchronized")
+				err = s.syncTrustedState(latestSyncedBatch)
```
ARR552, line 136: Maybe including this into the else, you could replicate the same logic.
```
if waitDuration != s.cfg.SyncInterval.Duration {

} else {
      err = s.syncTrustedState(latestSyncedBatch)
}
```

synchronizer/synchronizer.go
```diff
@@ -682,3 +730,113 @@ func (s *ClientSynchronizer) processVerifiedBatch(verifiedBatch etherman.Verifie
...
+			GlobalExitRoot:    common.HexToHash(trustedBatch.GlobalExitRoot),
+			RawTxsData:        trustedBatchL2Data,
+			// ForcedAt:          trustedBatch.,
+		}
+
+		log.Debugf("adding forced batch %v for batch %v", trustedBatch.ForcedBatchNumber, trustedBatch.BatchNumber)
+		if err := s.state.AddForcedBatch(s.ctx, forcedBatch, dbTx); err != nil {
+			log.Errorf("error adding forced batch for batch %v", trustedBatch.BatchNumber)
+			return err
+		}
```
ARR552, line 837: I guess that here you are adding the batchNumber field to the forcedBatch to remove it from the queue and leave it in a stated "used". Am I right?

ARR552 (changes requested):

synchronizer/synchronizer.go
```diff
@@ -236,6 +232,58 @@ func (s *ClientSynchronizer) syncBlocks(lastEthBlockSynced *state.Block) (*state
...
+
+		dbTx, err := s.state.BeginStateTransaction(s.ctx)
+		if err != nil {
+			log.Fatalf("error creating db transaction to sync trusted batch %v: %v", batchNumberToSync, err)
+		}
+
+		if err := s.processTrustedBatch(batchToSync, dbTx); err != nil {
+			log.Fatalf("error processing trusted batch %v: %v", batchNumberToSync, err)
+			err := dbTx.Rollback(s.ctx)
```
arnaubennassar, line 271: This line is unreachable, as the previous line will throw a panic

synchronizer/synchronizer.go
```diff
@@ -682,3 +730,113 @@ func (s *ClientSynchronizer) processVerifiedBatch(verifiedBatch etherman.Verifie
...
+	if err := s.state.ResetTrustedState(s.ctx, trustedBatch.BatchNumber-1, dbTx); err != nil {
+		log.Errorf("failed to reset trusted state", trustedBatch.BatchNumber)
+		return err
+	}
+
+	log.Debugf("opening batch %v", trustedBatch.BatchNumber)
+	if err := s.state.OpenBatch(s.ctx, processCtx, dbTx); err != nil {
+		log.Errorf("error opening batch %d", trustedBatch.BatchNumber)
+		return err
+	}
```
arnaubennassar, line 791: `ProcessSequencerBatch` which you are calling just bellow already takes care of opening the batch

synchronizer/synchronizer.go
```diff
@@ -682,3 +730,113 @@ func (s *ClientSynchronizer) processVerifiedBatch(verifiedBatch etherman.Verifie
...
+	if err != nil {
+		log.Errorf("error processing sequencer batch for batch: %d", trustedBatch.BatchNumber)
+		return err
+	}
+
+	log.Debugf("storing transactions for batch %v", trustedBatch.BatchNumber)
+	if err = s.state.StoreTransactions(s.ctx, trustedBatch.BatchNumber, processBatchResp.Responses, dbTx); err != nil {
+		log.Errorf("failed to store transactions for batch: %d", trustedBatch.BatchNumber)
+		return err
+	}
```
arnaubennassar, line 804: `ProcessSequencerBatch` which you are calling above already takes care of strong transactions

synchronizer/synchronizer.go
```diff
@@ -682,3 +730,113 @@ func (s *ClientSynchronizer) processVerifiedBatch(verifiedBatch etherman.Verifie
...
+			BatchNumber:   trustedBatch.BatchNumber,
+			StateRoot:     common.HexToHash(trustedBatch.StateRoot),
+			LocalExitRoot: common.HexToHash(trustedBatch.LocalExitRoot),
+		}
+		log.Debugf("closing batch %v", trustedBatch.BatchNumber)
+		if err := s.state.CloseBatch(s.ctx, receipt, dbTx); err != nil {
+			log.Errorf("error closing batch %d", trustedBatch.BatchNumber)
+			return err
+		}
+	}
```
arnaubennassar, line 819: `ProcessSequencerBatch` also closes the batch. Probably it's a bad idea to use `ProcessSequencerBatch` in this context, and a combo of [OpenBatch, ProcessSequencerBatch, StoreTransactions, CloseBatch]. I think it's worth to discuss this on a huddle. Please ping em when you're available

synchronizer/synchronizer.go
arnaubennassar: No, `AddForcedBatch` is inserting a new batch. What should happen here is what @ARR552 suggests:

- The forced batch should already exist on the DB
- We need to mark this batch as sequenced (the forced batch has moved from the forced queue into the state). This is done by updating the `batch_num` from `NULL` to `trustedBatch.BatchNumber`

arnaubennassar (changes requested):

synchronizer/synchronizer.go
ARR552: We should wait until the l1 synchronization is done.The waitDuration is 0 at the beginning. So if it is equal to the value in the config file, l1 is synced and we could start syncing the trusted state. We could have problems with s.cfg.SyncInterval.Duration = 0s
We will syncing l1 every waitDuration but we want check if it is updated because it is.

synchronizer/synchronizer.go
```diff
@@ -682,3 +730,113 @@ func (s *ClientSynchronizer) processVerifiedBatch(verifiedBatch etherman.Verifie
...
+		log.Infof("batch %v needs to be updated", trustedBatch.BatchNumber)
+	} else {
+		log.Infof("batch %v needs to be synchronized", trustedBatch.BatchNumber)
+	}
+
+	log.Debugf("resetting trusted state from batch %v", trustedBatch.BatchNumber)
+	if err := s.state.ResetTrustedState(s.ctx, trustedBatch.BatchNumber-1, dbTx); err != nil {
+		log.Errorf("failed to reset trusted state", trustedBatch.BatchNumber)
+		return err
+	}
```
ARR552, line 785: Why do you reset the trusted state here?? Maybe to remove the txs synced until this moment and proccess them again?? If it is, seems that it is not optimal. If not, I think we shouldn't reset the state

arnaubennassar: As I've mentioned above, we need to discuss the `sync <==> state` interaction for this scenario. This PR uses a mix of the interface for sequencer and synchronizer. This is normal because when we designed the state interface we didn't have this scenario in mind.

Let's do a huddle when you're online

config/config.debug.toml
tclemos: This config is not the `TrustedSequencerURL` which is supposed to be the `URL` of the `JSON RPC Server`. 
This is the `IP:PORT` of the `Broadcast Service` provided by the `Trusted Sequencer`

I agree to get this info from the `SMC`, but looks like the current information provided by the `SMC` is not enough to fulfill this feature needs.

An option would be to create a new custom `JSON RPC Server` endpoint like `hez_getBroadcastURI`, so we can query this from the `Trusted Sequencer` after getting its `URL` from the `SMC`.

wdyt @arnaubennassar?

synchronizer/synchronizer.go
tclemos: This check is not valid anymore since the trusted stated will be ahead of the latest sequenced batch in the L1.

docker-compose.yml
```diff
@@ -37,6 +37,24 @@ services:
       - "-c"
       - "/app/zkevm-node run --network local --cfg /app/config.toml --components rpc"
 
+  zkevm-non-sequencer-json-rpc:
```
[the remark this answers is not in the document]
tclemos: `zkevm-non-sequencer-json-rpc`: relay tx to the trusted node
`zkevm-permissionless-node`: sync trusted state

probably we need to merge them and also review how we are configuring them because we have the `URL` of the `Trusted Sequencer` provided by the `SMC`

synchronizer/synchronizer.go
tclemos: where?

synchronizer/synchronizer.go
tclemos: I couldn't find the code handling the transactions, can you show me where it's happening?

synchronizer/synchronizer.go
tclemos: I couldn't find the code closing the batch, can you show me where it's happening?

synchronizer/synchronizer.go
tclemos: The behavior of the synchronizer now is different from before because the old implementation was considering data coming only from L1, but now we have data coming from L1(virtual and verified) and L2(trusted).

The method `s.syncBlocks` is the one that syncs the L1(virtual and verified) data, but this method doesn't know if all the L1 data was synced, so the sync process after syncing part of the L1 needs to check the last batch in the L2 state and, the last batch in the rollup to decide if its time to start syncing data from L2(trusted).

synchronizer/synchronizer.go
tclemos: `waitDuration != s.cfg.SyncInterval.Duration` means the variable used to control the interval between sync cycles and the parameter configured to define this interval are different, this doesn't mean `L1 was fully synchronized`

What defines if the L1 was fully synced is checking the L2 last batch against the last batch in the rollup.

synchronizer/synchronizer.go
tclemos: When you say `it is not optimal` is It based on what? performance? maintenance? safeness? code readability? code complexity?

A previous trusted state synced from the trusted sequencer can also be affected by an L1 reorg or a trusted state reorg due to a forced batch. These scenarios would lead us to have processed `"wrong"/"invalid"` transactions from previously trusted state synchronization.

When the synchronization understands it's time to sync a batch from the trusted state, sure this is not the most performative solution, but it is the safest and simple thing to do until we have a good reason to change this behavior.

I recommend looking for two topics:
- Pareto Principle 80 20
- Early optimization is bad

synchronizer/synchronizer.go
```diff
@@ -682,3 +730,113 @@ func (s *ClientSynchronizer) processVerifiedBatch(verifiedBatch etherman.Verifie
...
+
+	log.Debug("trustedBatch.StateRoot ", trustedBatch.StateRoot)
+	isBatchClosed := trustedBatch.StateRoot != state.ZeroHash.String()
+	if isBatchClosed {
+		receipt := state.ProcessingReceipt{
+			BatchNumber:   trustedBatch.BatchNumber,
+			StateRoot:     common.HexToHash(trustedBatch.StateRoot),
+			LocalExitRoot: common.HexToHash(trustedBatch.LocalExitRoot),
```
arnaubennassar, line 812: Use whatever it comes from exxecutor instead

arnaubennassar (changes requested):

config/config.debug.toml
tclemos: We end up going with the solution to create a new endpoint to provide the broadcast URI.

synchronizer/config.go
```diff
@@ -11,4 +11,8 @@ type Config struct {
 
 	// SyncChunkSize is the number of blocks to sync on each chunk
 	SyncChunkSize uint64 `mapstructure:"SyncChunkSize"`
+
+	// TrustedSequencerURI is used to allow permission less nodes
+	// to synchronize the trusted state from the trusted sequencer.
```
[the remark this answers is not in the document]
tclemos: added the `Trusted` config to the root of the configuration, the default is `false` as most of the nodes are going to run as a `permissionless node`

synchronizer/synchronizer.go
tclemos: replaced this `Fatal` by `Error`

synchronizer/synchronizer.go
tclemos: done

etherman/etherman.go
```diff
@@ -582,6 +582,11 @@ func (etherMan *Client) ApproveMatic(maticAmount *big.Int, to common.Address) (*
 	return tx, nil
 }
 
+// GetTrustedSequencerURL Gets the trusted sequencer url from rollup smc
+func (etherMan *Client) GetTrustedSequencerURL() (string, error) {
+	return etherMan.PoE.TrustedSequencerURL(nil)
```
ARR552, line 587:
```suggestion
	return etherMan.PoE.TrustedSequencerURL(&bind.CallOpts{Pending: false})
```

synchronizer/synchronizer.go
```diff
@@ -23,18 +29,20 @@ type Synchronizer interface {
...
 type ClientSynchronizer struct {
-	etherMan          ethermanInterface
-	state             stateInterface
-	ctx               context.Context
-	cancelCtx         context.CancelFunc
-	genBlockNumber    uint64
-	genesis           state.Genesis
-	reorgBlockNumChan chan struct{}
-	cfg               Config
+	IsTrustedSequencer bool
```
ARR552, line 32: is it necessary to make it public?? or could it be internal?

synchronizer/synchronizer.go
```diff
@@ -236,6 +230,102 @@ func (s *ClientSynchronizer) syncBlocks(lastEthBlockSynced *state.Block) (*state
...
+		log.Errorf("error getting broadcast URI. Error: %v", err)
+		return err
+	}
+	log.Debug("broadcastURI ", broadcastURI)
+
+	if broadcastURI != "zkevm-broadcast:61090" {
+		log.Error("URI IS DIFFERENT")
+		log.Error([]byte(broadcastURI))
+		log.Error([]byte("zkevm-broadcast:61090"))
+	}
```
ARR552, line 253: I don't understand why we need this. We have a hardcoded url and as we discussed it is a bad idea

ARR552 (changes requested):

synchronizer/synchronizer.go
arnaubennassar: yeah, looks like you left this while debugging :sweat_smile:

arnaubennassar (changes requested):

arnaubennassar (approved):

ARR552 (approved):

PR #956, opened 2022-07-28
Fix updatedeps

### LINKED ISSUE #954: Fix updatedeps GHA

fgimenez:
Closes #954

### What does this PR do?
* Use `PERSONAL_ACCESS_TOKEN` instead of `UPDATE_DEPS_SSH_PK` to authenticate against github API
* Update list of docker images checked by updatedeps.

### Reviewers
@arnaubennassar

### DIFF
--- a/scripts/cmd/dependencies.go
+++ b/scripts/cmd/dependencies.go
@@ -8,7 +8,7 @@ import (
 func updateDeps(ctx *cli.Context) error {
 	cfg := &dependencies.Config{
 		Images: &dependencies.ImagesConfig{
-			Names:          []string{"hermeznetwork/geth-zkevm-contracts", "hermeznetwork/zkevm-mock-prover"},
+			Names:          []string{"hermeznetwork/geth-zkevm-contracts", "hermeznetwork/zkprover-local"},
 			TargetFilePath: "../../../docker-compose.yml",
 		},
 		PB: &dependencies.PBConfig{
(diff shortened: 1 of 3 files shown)


arnaubennassar (approved):

fgimenez:
All green and execution of updatedeps successful from this branch https://github.com/0xPolygonHermez/zkevm-node/runs/7558377134?check_suite_focus=true, merging

scripts/cmd/dependencies.go
```diff
@@ -8,16 +8,16 @@ import (
 func updateDeps(ctx *cli.Context) error {
 	cfg := &dependencies.Config{
 		Images: &dependencies.ImagesConfig{
-			Names:          []string{"hermeznetwork/geth-zkevm-contracts", "hermeznetwork/zkevm-mock-prover"},
+			Names:          []string{"hermeznetwork/geth-zkevm-contracts", "hermeznetwork/zkprover-local"},
```
ARR552, line 11: Maybe we need to use the new 0xPolygonHermez instead fo hermeznetwork

ARR552 (approved):

scripts/cmd/dependencies.go
fgimenez: Well, these are docker image names as found in docker-compose.yml. Not sure if there's a new `0xPolygonHermez` in Docker Hub, in any case, the image names here should match what we have in docker-compose.yml (as they do now), and both should be updated at the same time.

PR #961, opened 2022-07-28
Move merkletree table o prover database

### LINKED ISSUE #951: Remove state.merkletree
In order to move on with the implementation, we added the merkletree table to the state schema. This is super dirty since this is meant to be used by the merkletree service.

Instead we should create a separated schema or even database for the merkletree

xavier-romero:
Closes #951 

### What does this PR do?

Moves merkletree table to isolated prover database, thus removing table creation from app migration file.

### Reviewers

Main reviewers:

- @arnaubennassar 
- @fgimenez


config/init.sql
```diff
@@ -0,0 +1,14 @@
+CREATE DATABASE prover_db;
```
arnaubennassar, line 1: Can you rename the file `init.sql` ==> `initproverdb.sql`?

arnaubennassar (changes requested):

fgimenez (approved):

config/init.sql
xavier-romero: There is only one SQL file that can be attached to initialize the postgres container with whatever we need, and it's common practice to name it **init.sql**. In example, if tomorrow we want to create additional users for each component (sequencer, agregator, etc), that would be done in this file as well, so I would prefer to avoid tying the filename to the specific action it is doing right now, since in fact it could change at any moment. Are you ok with this?

arnaubennassar (approved):

Issue #941, opened 2022-07-26
Aggregator unit tests


Mikelle:
Add unit tests for the aggregator

PR #964, opened 2022-07-28
Aggregator refactoring and tests addition

Mikelle:
Closes https://github.com/0xPolygonHermez/zkevm-node/issues/941

### What does this PR do?

Aggregator refactoring and tests addition

### Reviewers

Main reviewers:

- @fgimenez 
- @arnaubennassar


aggregator/aggregator_internal_test.go
```diff
@@ -0,0 +1,272 @@
...
+	proverClient := new(aggrMocks.ProverClientMock)
+	a := Aggregator{
+		State:                st,
+		EthTxManager:         ethTxManager,
+		Ethman:               etherman,
+		ProverClient:         proverClient,
+		lastVerifiedBatchNum: 1,
+	}
+	var (
+		oldStateRoot     = common.HexToHash("oldroot")
```
fgimenez, line 87: "oldroot" is not actually a Hex string, out of curiosity, what is the output of `common.HexToHash("oldroot")`? Looks like `common.HexToHash` calls at some point `common.Hex2Bytes`, and the potential error from `hex.DecodeString` is ignored here https://github.com/ethereum/go-ethereum/blob/master/common/bytes.go#L80

aggregator/aggregator.go
```diff
@@ -43,275 +40,238 @@ func NewAggregator(
...
+	if lastVerifiedBatch != nil {
+		a.lastVerifiedBatchNum = lastVerifiedBatch.BatchNumber
+	}
+	lastConsolidatedEthBatchNum, err := a.Ethman.GetLatestVerifiedBatchNum()
+	if err != nil {
+		log.Warnf("failed to get last eth batch, err: %v", err)
+		return false
+	}
+	if a.lastVerifiedBatchNum < lastConsolidatedEthBatchNum {
+		log.Infof("waiting for the state to be synced, lastConsolidatedBatchNum: %d, lastEthConsolidatedBatchNum: %d",
```
fgimenez, line 148:
```suggestion
		log.Infof("waiting for the state to be synced, lastVerifiedBatchNum: %d, lastEthConsolidatedBatchNum: %d",
```

aggregator/aggregator.go
```diff
@@ -43,275 +40,238 @@ func NewAggregator(
...
-			batchesSent[batchToVerify.BatchNumber] = true
-		case <-a.ctx.Done():
-			getProofCtxCancel()
-			return
+		a.tryVerifyBatch(ctx, ticker)
+	}
+}
+
+func (a *Aggregator) tryVerifyBatch(ctx context.Context, ticker *time.Ticker) {
+	// 1. check, if state is synced
```
fgimenez, line 77: Could we have `log.Info` instead of these comments?

fgimenez (approved):
Looks good, some nitpicks.

aggregator/aggregator.go
```diff
@@ -43,275 +40,238 @@ func NewAggregator(
...
+		return
+	}
+
+	genProofID, err := a.ProverClient.GetGenProofID(ctx, inputProver)
+	if err != nil {
+		log.Warnf("failed to get gen proof id, err: %v", err)
+		return
+	}
+
+	resGetProof, err := a.ProverClient.GetResGetProof(ctx, genProofID, batchToVerify.BatchNumber)
```
arnaubennassar, line 115: shouldn't this be in a loop waiting for the proof generation to either complete (which takes a considerable amount of time) or fail?

aggregator/aggregator.go
```diff
@@ -43,275 +40,238 @@ func NewAggregator(
...
+		log.Infof("batch with number %d was already sent, but not yet consolidated by synchronizer",
+			batchToVerify.BatchNumber)
+		return nil, nil
+	}
+
+	return batchToVerify, nil
+}
+
+func (a *Aggregator) buildInputProver(ctx context.Context, batchToVerify *state.Batch) (*pb.InputProver, error) {
+	oldStateRoot, err := a.State.GetStateRootByBatchNumber(ctx, a.lastVerifiedBatchNum, nil)
```
arnaubennassar, line 179:
```suggestion
	oldStateRoot, err := a.State.GetStateRootByBatchNumber(ctx, batchToVerify.BatchNumber -1, nil)
```

aggregator/aggregator.go
```diff
@@ -43,275 +40,238 @@ func NewAggregator(
...
+	if err != nil && err != state.ErrNotFound {
+		return nil, fmt.Errorf("failed to get current state root, err: %v", err)
+	}
+
+	newStateRoot, err := a.State.GetStateRootByBatchNumber(ctx, batchToVerify.BatchNumber, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get state root to consolidate, err: %v", err)
+	}
```
arnaubennassar, line 187:
```suggestion
	newStateRoot := batchToVerify.StateRoot
```

aggregator/aggregator.go
```diff
@@ -43,275 +40,238 @@ func NewAggregator(
...
+
+	rawTxs, err := state.EncodeTransactions(batchToVerify.Transactions)
+	if err != nil {
+		return nil, fmt.Errorf("failed to encode transactions, err: %v", err)
+	}
```
arnaubennassar, line 192:
```suggestion
	rawTxs := batchToVerify.BatchL2Data
```

aggregator/aggregator.go
arnaubennassar: Actually, it would be better if you get the full batch, so you avoid extra DB calls later on, for instance calling `GetLocalExitRootByBatchNumber`

aggregator/aggregator.go
```diff
@@ -43,275 +40,238 @@ func NewAggregator(
...
+	globalExitRoot := batchToVerify.GlobalExitRoot
+
+	oldLocalExitRoot, err := a.State.GetLocalExitRootByBatchNumber(ctx, a.lastVerifiedBatchNum, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get local exit root for batch %d, err: %v", a.lastVerifiedBatchNum, err)
+	}
```
arnaubennassar, line 198: Get full batch in previous step so you can do something like:

```suggestion
	oldLocalExitRoot := previousBatch.LocalExitRoot
```

aggregator/aggregator.go
Mikelle: Yes, it's in the loop, it's here - https://github.com/0xPolygonHermez/zkevm-node/pull/964/files#diff-1a6a23e9ac05b8b9b16f35d7459375093c07b45bfb35da0fe9799514d8e3cc97R54

tclemos (approved):

ToniRamirezM (approved):

arnaubennassar (approved):

PR #967, opened 2022-07-29
Close Synchronized Batch

### LINKED ISSUE #859: Differentiate CloseBatch when it's called by sequencer / sync
- In the case of sequencer, the `L2BatchData` should be built from stored txs (already implemented)
- In the case of sync the`L2BatchData` should be stored as seen on L1

ToniRamirezM:
Closes #859 

### What does this PR do?

Implements a closeBatch method for synchronized batches.

### Reviewers

Main reviewers:

- @arnaubennassar 
- @tclemos

### DIFF
--- a/state/state.go
+++ b/state/state.go
@@ -445,8 +445,8 @@ func (s *State) isBatchClosable(ctx context.Context, receipt ProcessingReceipt, 
 	return nil
 }
 
-// CloseSynchronizedBatch is used by Synchronizer to close the current batch.
-func (s *State) CloseSynchronizedBatch(ctx context.Context, receipt ProcessingReceipt, txs []types.Transaction, dbTx pgx.Tx) error {
+// closeSynchronizedBatch is used by Synchronizer to close the current batch.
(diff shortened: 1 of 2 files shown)


tclemos (approved):

state/state.go
```diff
@@ -448,6 +441,47 @@ func (s *State) CloseBatch(ctx context.Context, receipt ProcessingReceipt, dbTx
 	if isLastBatchClosed {
 		return ErrBatchAlreadyClosed
 	}
+
+	return nil
+}
+
+// CloseSynchronizedBatch is used by Synchronizer to close the current batch.
+func (s *State) CloseSynchronizedBatch(ctx context.Context, receipt ProcessingReceipt, txs []types.Transaction, dbTx pgx.Tx) error {
```
arnaubennassar, line 449: I understand that this function should be call only by `ProcessAndStoreClosedBatch`. If this is the case, please make the function private.

Please keep in mind that the synchronizer also calls the Open / Close batch methods, when synchronizing from the trusted sequencer broadcast. In this scenario, whoever is implementing this could be very tempted to call `CloseSynchronizedBatch` instead of `CloseBatch`

arnaubennassar (changes requested):

arnaubennassar (approved):

Issue #968, opened 2022-07-29
Sequencer is trying to close a batch without transactions


arnaubennassar:
Note that this could be legit for instance to update the GER, but this is not the case as no one is using the bridge

```
2022-07-29T10:51:53Z    INFO    sequencer/sequencer.go:191      synchronizer has synced last batch, checking if current sequence should be closed
2022-07-29T10:51:53Z    INFO    sequencer/sequencer.go:318      current sequence should be closed
2022-07-29T10:51:53Z    ERROR   sequencer/sequencer.go:322      failed to create new sequence, err: failed to close batch, err: can not close a batch without transactions%!(EXTRA string=
/src/log/log.go:106 github.com/0xPolygonHermez/zkevm-node/log.appendStackTraceMaybeArgs()
/src/log/log.go:164 github.com/0xPolygonHermez/zkevm-node/log.Errorf()
/src/sequencer/sequencer.go:322 github.com/0xPolygonHermez/zkevm-node/sequencer.(*Sequencer).closeSequence()
/src/sequencer/sequencer.go:192 github.com/0xPolygonHermez/zkevm-node/sequencer.(*Sequencer).tryToProcessTx()
/src/sequencer/sequencer.go:137 github.com/0xPolygonHermez/zkevm-node/sequencer.(*Sequencer).Start()
)
2022-07-29T10:51:53Z    INFO    sequencer/sequencer.go:191      synchronizer has synced last batch, checking if current sequence should be closed
2022-07-29T10:51:53Z    INFO    sequencer/sequencer.go:318      current sequence should be closed
2022-07-29T10:51:53Z    ERROR   sequencer/sequencer.go:322      failed to create new sequence, err: failed to close batch, err: can not close a batch without transactions%!(EXTRA string=
/src/log/log.go:106 github.com/0xPolygonHermez/zkevm-node/log.appendStackTraceMaybeArgs()
/src/log/log.go:164 github.com/0xPolygonHermez/zkevm-node/log.Errorf()
/src/sequencer/sequencer.go:322 github.com/0xPolygonHermez/zkevm-node/sequencer.(*Sequencer).closeSequence()
/src/sequencer/sequencer.go:192 github.com/0xPolygonHermez/zkevm-node/sequencer.(*Sequencer).tryToProcessTx()
```

ToniRamirezM:
Take into account that there is an open PR that may not solve this, but affects the code.

PR #971, opened 2022-07-29
fixing close empty batch and failure to send tx to the ethereum

Mikelle:
Closes https://github.com/0xPolygonHermez/zkevm-node/issues/968
Closes https://github.com/0xPolygonHermez/zkevm-node/issues/978

### What does this PR do?

- Simplifies the tx manager to be sequential and panic if fails to send tx to L1, no retries. In the future we should improve this, right now the code is set up in a way to make it obvious and debuggable if :hankey: happens. This includes waiting for the tx to be mined
- Unify `shouldCloseSequence` under a single function
- Remove SC compilation from `make run`
- Setup correct values for ZK counters in config file and defaults
- Added a bunch of logs to understand what's happening
- Added reason of revert on `WaitTxToBeMined`: it detects both failed reason (for instance, `out of gas`) and revert messages

### Reviewers

Main reviewers:

- @ToniRamirezM 
- @KonradIT


config/default.go
```diff
@@ -48,13 +48,13 @@ LastTimeBatchMaxWaitPeriod = "15s"
...
-MaxArithmetics = 100
-MaxBinaries = 100
-MaxSteps = 100
+MaxKeccakHashes = 1000
+MaxPoseidonHashes = 1000
+MaxPoseidonPaddings = 1000
+MaxMemAligns = 1000
+MaxArithmetics = 1000
+MaxBinaries = 1000
+MaxSteps = 10000
```
Mikelle, line 57: Increased those variables, bcs previous one was not suitable even for sending a simple transfer

etherman/etherman.go
```diff
@@ -198,7 +198,7 @@ func (etherMan *Client) EstimateGasSequenceBatches(sequences []ethmanTypes.Seque
 	if err != nil {
 		return 0, err
 	}
-	return tx.Cost().Uint64(), nil
+	return tx.Gas(), nil
```
Mikelle, line 201: Returned back, bcs it was impossible to send a tx with tx.Cost instead of tx.Gas. A request was failing with "exceeds block gas limit" even after 10 times increase

ethtxmanager/ethtxmanager.go
```diff
@@ -63,13 +63,25 @@ func (c *Client) TrackSequenceBatchesSending(ctx context.Context) {
...
+
+			gasLimit := uint64(float64(gas) * gasLimitIncrease)
+			attempts = 0
+			err = c.sequenceBatches(sequences, gasLimit)
+			for err != nil && attempts < c.cfg.MaxSendBatchTxRetries {
+				log.Errorf("failed to sequence batches, trying once again, retry #%d, gasLimit: %d, err: %v",
+					attempts, gasLimit, err)
+				time.Sleep(c.cfg.FrequencyForResendingFailedSendBatches.Duration)
+				attempts++
+				gasLimit = uint64(float64(gasLimit) * gasLimitIncrease)
```
ARR552, line 83: You should increase the GasPrice instead of the gas. 
As you may know gas is the effort unit to process the tx and gasprice is the amount of wei you are willing to pay for each gas unit.
You should be able to tweak the value here `auth.GasPrice`

ARR552 (changes requested):

ethtxmanager/ethtxmanager.go
arnaubennassar: Resolving conversation as the affected code by the review has been completely removed

Makefile
```diff
@@ -195,12 +195,11 @@ stop-explorer-db: ## Stops the explorer database
 	$(STOPEXPLORERDB)
 
 .PHONY: run
-run: compile-scs ## Runs all the services
```
KonradIT, line 198: Why is this step removed?

KonradIT (approved):

ethtxmanager/ethtxmanager.go
```diff
@@ -63,14 +54,30 @@ func (c *Client) TrackSequenceBatchesSending(ctx context.Context) {
 	for {
 		select {
 		case sequences := <-c.sequencesToSendChan:
-			err := c.sequenceBatches(sequences)
 			var attempts uint32
+			// attempts = 0
```
ToniRamirezM, line 58: Remove comment

ethtxmanager/ethtxmanager.go
```diff
@@ -63,14 +54,30 @@ func (c *Client) TrackSequenceBatchesSending(ctx context.Context) {
...
+			log.Info("sending sequence to L1")
+			tx, err := c.ethMan.SequenceBatches(sequences, 0)
 			for err != nil && attempts < c.cfg.MaxSendBatchTxRetries {
-				log.Errorf("failed to sequence batches, trying once again, retry #%d, err: %v", attempts, err)
+				log.Errorf("failed to sequence batches, trying once again, retry #%d, gasLimit: %d, err: %v",
+					attempts, 0, err)
 				time.Sleep(c.cfg.FrequencyForResendingFailedSendBatches.Duration)
 				attempts++
-				err = c.sequenceBatches(sequences)
+				// gasLimit = uint64(float64(gasLimit) * gasLimitIncrease)
```
ToniRamirezM, line 66: Remove comment

test/operations/wait.go
```diff
@@ -189,28 +193,56 @@ func grpcHealthyCondition(address string) (bool, error) {
...
+		}
+		return false, fmt.Errorf("transaction has failed, reason: %s, receipt: %+v. tx: %+v", reason, receipt, tx)
+	}
+	return true, nil
+}
 
-	var done bool
-	if !isPending {
-		r, err := client.TransactionReceipt(ctx, hash)
+func revertReason(ctx context.Context, c ethClienter, tx *types.Transaction, blockNumber *big.Int) (string, error) {
```
ToniRamirezM, line 221: In state/errors.go there is a similar method to this: constructErrorFromRevert

Makefile
arnaubennassar: because it's not needed to run the node and it takes time

test/operations/wait.go
arnaubennassar: I think it's not similar enough to reuse it

ToniRamirezM (approved):

arnaubennassar (approved):

PR #1004, opened 2022-08-04
Executor - Nonce Test

### LINKED ISSUE #1002: State Test: Nonces Match in Executor [state]

tclemos:
Closes #1002.

### What does this PR do?

Adds a test to make sure Executor is taking into consideration the transaction nonce to consider a tx as a valid tx.

### Reviewers

Main reviewers:

@ToniRamirezM 
@KonradIT 
@Mikelle 
@arnaubennassar

### DIFF
--- a/state/state_test.go
+++ b/state/state_test.go
@@ -5,11 +5,13 @@ import (
 	"fmt"
 	"math/big"
 	"os"
+	"strconv"
 	"strings"
 	"testing"
 	"time"
 
 	"github.com/0xPolygonHermez/zkevm-node/db"
+	"github.com/0xPolygonHermez/zkevm-node/encoding"
 	"github.com/0xPolygonHermez/zkevm-node/hex"
 	"github.com/0xPolygonHermez/zkevm-node/log"
 	"github.com/0xPolygonHermez/zkevm-node/merkletree"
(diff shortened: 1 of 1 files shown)


KonradIT (changes requested):
Maybe add a log for `testCase.Name` so its clear which error it belongs to

Right now this

```
Error:      	Not equal: 
        	            	expected: 1
        	            	actual  : 14
        	Test:       	TestExecutorInvalidNonce
        	Messages:   	invalid tx Error, it is expected to be INVALID TX
```

Is repeated 3x and not very clear which TestCase it belongs to. Otherwise, it's good.

state/state_test.go
```diff
@@ -1226,3 +1228,106 @@ func TestExecutorTxHash(t *testing.T) {
...
+			EthTimestamp:     uint64(0),
+			UpdateMerkleTree: 1,
+		}
+
+		// Process batch
+		processBatchResponse, err := executorClient.ProcessBatch(ctx, processBatchRequest)
+		require.NoError(t, err)
+
+		transactionResponses := processBatchResponse.GetResponses()
+		assert.Equal(t, transactionResponses[0].Error, executorclientpb.Error_ERROR_INVALID_TX, "invalid tx Error, it is expected to be INVALID TX")
```
arnaubennassar, line 1331: Switch order: `actual, expected` ==> `expected, actual`

```suggestion
		assert.Equal(t,  executorclientpb.Error_ERROR_INVALID_TX, transactionResponses[0].Error, "invalid tx Error, it is expected to be INVALID TX")
```

arnaubennassar (changes requested):
If you update the zk prover image to `sha256:bbece5756c2c9ebc84a8564d7c7dacb4137f6b9600cedce45546fdd5159bd198`, you'll be closer to make this test work! It looks like the C++ version of the executor is returning the enums with a +1:

- With the version used in this PR it returns 1, when it should be 0 as errors are not implemented
- Using the version I've just suggested it returns 15 instead of 14

state/state.go
```diff
@@ -542,7 +549,9 @@ func (s *State) ProcessAndStoreClosedBatch(ctx context.Context, processingCtx Pr
 	// Filter unprocessed txs and decode txs to store metadata
 	// note that if the batch is not well encoded it will result in an empty batch (with no txs)
 	for i := 0; i < len(processed.Responses); i++ {
-		if !isTransactionProcessed(processed.Responses[i].UnprocessedTransaction) {
+
+		//TODO: Also check this
+		if false { // if !isTransactionProcessed(processed.Responses[i]) {
```
KonradIT, line 554: Bunch of `// TODO` and nasty fixes on the code. Keep in this mind.

state/state.go
```diff
@@ -509,8 +513,11 @@ func (s *State) CloseBatch(ctx context.Context, receipt ProcessingReceipt, dbTx
 
 // isTransactionProcessed determines if the given process transaction response
 // represents a processed transaction.
-func isTransactionProcessed(unprocessedTransaction uint32) bool {
-	return unprocessedTransaction == cFalse
+func isTransactionProcessed(input bool) bool {
+
+	//TODO: Finish this function
+
+	return false
```
arnaubennassar, line 520: remove this function, as this logic is actually implemented in `func isUnprocessed(error pb.Error) bool`

arnaubennassar (changes requested):

KonradIT (approved):

arnaubennassar (approved):

PR #1005, opened 2022-08-04
Update merkletree behaviour

### LINKED ISSUE #979: Update merkletree behaviour
- Update hash of the bytecode:
  - spec: https://hackmd.io/bbknowuiQxm3xrgWOt890g#SC-Code
  - impl: https://github.com/0xPolygonHermez/zkevm-commonjs/blob/main/src/smt-utils.js#L403
- Add bytecode length leaf
  - spec: https://hackmd.io/bbknowuiQxm3xrgWOt890g#SC-length

Note that this is important for the genesis, as other smart contracts will be deployed through transactions and the executor will handle this

ToniRamirezM:
Closes #979 

### What does this PR do?

- Implements new leaf type (code lenght) and new byte code hash function.
- Adds tests for this.

### Reviewers

Main reviewers:

- @arnaubennassar 
- @fgimenez 

Codeowner reviewers:

- @tclemos


state/state_test.go
```diff
@@ -59,7 +59,7 @@ func TestMain(m *testing.M) {
 	}
 	defer stateDb.Close()
 
-	zkProverURI := testutils.GetEnv("ZKPROVER_URI", "localhost")
+	zkProverURI := testutils.GetEnv("ZKPROVER_URI", "54.170.178.97")
```
fgimenez, line 62: We should keep localhost here
```suggestion
	zkProverURI := testutils.GetEnv("ZKPROVER_URI", "localhost")
```

state/state_test.go
```diff
@@ -1226,3 +1226,85 @@ func TestExecutorTxHash(t *testing.T) {
...
+				Type:            int(merkletree.LeafTypeStorage),
+				StoragePosition: "0x0000000000000000000000000000000000000000000000000000000000005BBF",
+				Value:           "0x0B6E",
+			},
+		},
+	}
+
+	if err := dbutils.InitOrReset(cfg); err != nil {
+		panic(err)
+	}
```
fgimenez, line 1300:
```suggestion
	require.NoError(t, dbutils.InitOrReset(cfg))
```

fgimenez (approved):
:rocket:

state/state_test.go
ToniRamirezM: Yes, I should have not pushed that

tclemos (approved):

PR #1040, opened 2022-08-16
Fix nonce order

### LINKED ISSUE #1003: Take nonce into consideration when selecting a tx [sequencer]
Currently it seems the nonce can be arbitrary and not follow the x+1 rule.

KonradIT:
Closes #1003.

### What does this PR do?

Taken from the issue:

- [X] Add `from` column in `pool.txs`
- [X] Modify function to insert `from` value in db
- [X] Modify select query to get the smallest nonce grouped by `from` row 
- [X] E2E 1000 Test should pass

### Reviewers

Main reviewers:

- @arnaubennassar 
- @tclemos 

Codeowner reviewers:


test/e2e/ethtransfer_test.go
```diff
@@ -31,15 +31,15 @@ func Test1000EthTransfer(t *testing.T) {
...
 	// Load account with balance on local genesis
-	auth, err := operations.GetAuth("0x59c6995e998f97a5a0044966f0945389dc9e86dae88c7a8412f4603b6b78690d", big.NewInt(1001))
+	auth, err := operations.GetAuth("0x59c6995e998f97a5a0044966f0945389dc9e86dae88c7a8412f4603b6b78690d", big.NewInt(1000))
 	require.NoError(t, err)
 	// Load eth client
 	client, err := ethclient.Dial("http://localhost:8123")
 	require.NoError(t, err)
 	// Send txs
-	nTxs := 100
+	nTxs := 1000
```
arnaubennassar, line 40: Now that the test is running on GHA and it waits for batches to be consolidated, the `1000` should be reduced. Otherwise the test timeout (timeout currently set to 10min). 

It's possible to increase the timeout, but IMO it's better to make the test faster, I'd try to change  `nTxs` to 100 or 200.

Also if you introduce this change, it would make sense to change the name of the test to `Test1000EthTransfer => TestEthTransfer`

arnaubennassar (changes requested):

arnaubennassar (approved):

tclemos (approved):

Mikelle (approved):

Issue #1008, opened 2022-08-05
Handle errors when processing transactions


ToniRamirezM:
Instead of validating this field the check should be done with the ProcessTransactionResponse.error field, it will be equals to ERROR_OUT_OF_COUNTERS.

Note that when a transactions makes the executor run out of counters, every TX in the batch will not get executed and all of them will have the ERROR_OUT_OF_COUNTERS error, even if they where processed in the same batch before adding the transaction that makes the executor run out of counters.

In this situation a new batch should be created. If the first transaction of a batch makes the executor run out of counters, that transaction will never be processable and should be discarded.

arnaubennassar:
Renamed issue, as it's a bit more complex. Badically the sequencer should take into account the following scenarios when receiving txs with errors:

A) The tx has an error but has affected the state. In this case, the tx should be included into the state. Example: a reverted tx
B) Tx has not affected the state and it will not be processable in the future. In this case, the tx shouldn't be added into the state and it should get deleted from the pool. Example: nonce smaller than current value
C) Tx has not affected the state but it could be processable in the future. In this case, the tx shouldn't be added into the state but it shouldn't get deleted from the pool. Example: out of counters in a batch with many other txs, nonce greater than current nonce +1

PR #1042, opened 2022-08-16
Feature/err process txs

Mikelle:
Closes https://github.com/0xPolygonHermez/zkevm-node/issues/1008

### Reviewers

Main reviewers:

- @arnaubennassar


arnaubennassar (approved):
Before merging, please open an issue explaining that we should cover the following case:

- If a tx with nonce > current account nonce is executed as a first tx in a batch it will get discarded as it fails to process within an empty batch. We shouldn't discard such a tx as it may be processable in the future

PR #1044, opened 2022-08-17
l2 chain id is now a network config

### LINKED ISSUE #1023: Fix executor input for unsigned transactions
After talking with @0xPolygonHermez/protocol we've decided that the executor will require correct nonce and chainID when using unsigned transactions. Since the jRPC method behind this functionality doesn't receive this data, we need to artificially inject this, probably in this method `func EncodeUnsignedTransaction(tx types.Transaction) ([]byte, error)`:

- [ ] ChainID it's already being added, but it's a hardcoded constant, instead of something that depends on network config
- [x] The nonce should be fetched from DB. Handled by the jRPC, implemented in this PR #1034

tclemos:
Closes #1023.

### What does this PR do?

replace the chain ID constant variables by a network config
Improve the `make check` command to check the essential steps instead of only unit tests and local build

### Reviewers

Main reviewers:

@arnaubennassar 
@Mikelle 
@KonradIT

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -143,7 +143,7 @@ lint: ## Runs the linter
 	$(LINT)
 
 .PHONY: check
-check: lint build test ## lint, build and unit tests
+check: stop lint build build-docker test-full-non-e2e test-e2e-group-2 ## lint, build and essential tests
 
 .PHONY: validate
 validate: lint build test-full ## lint, build, unit and e2e tests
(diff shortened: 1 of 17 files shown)


KonradIT (approved):
Absolutely needed, especially yesterday!

arnaubennassar (approved):

PR #1053, opened 2022-08-18
Fix GER latest l2 block for unsigned txs

### LINKED ISSUE #1048: Handle latest L2 block case for unsigned txs and estimate gas
While debugging the bridge service, @cool-develope faced a situation, that turned out to be an interesting edge case related to our protocol. The situation:

1. Do deposit on L1
2. Sequencer opens new batch with the updated GER
3. Estimate gas for claim
4. Tx reverted because the wrong GER

The problem:

- The state gets the execution context for the executor (timestamp, GER, ...) based on the relation of an L2 block number
- There is no L2 block number yet related to the latest (open) batch
- Therefore the executor fails as it doesn't have the right GER

tclemos:
Closes #1048.

### What does this PR do?

Adds the concept to differentiate `latest` requests from the ones strictly pointing to the `last` block in the chain for unsigned transactions requests(aka `eth_estimateGas` and `eth_call`)

When the `latest` L2 block is provided as a parameter, the node is going to consider the open batch in order to get the most updated GER instead of getting only the batches with l2 block associated.

This was needed because the process to update the GER made by the Synchronizer from L1, updates the GER for the open batch when there is no transaction for the current open batch, and this process does not generate a new l2 transaction forcing a new l2 block to be created.

This is the sequence of events:
- A deposit is made on L1 so the GER is updated on L1
- Synchronizer detects the GER updated on L1 and updates the GER on open batch on L2
- Bridge tries to estimate the gas for the L2 Claim tx
- The node uses the GER from the open batch to allow the Claim to be estimated
- Claim is estimated properly
- Claim is sent to L2
- Sequencer selects the Claim TX from the Pool
- Claim TX is added to the state
- L2 Block is created with the new State Root
- Batch is updated with the new roots

### Reviewers

@arnaubennassar 
@cool-develope 
@KonradIT 
@Mikelle


arnaubennassar (approved):

state/pgstatestorage.go
```diff
@@ -421,19 +421,43 @@ func (p *PostgresStorage) GetLastNBatches(ctx context.Context, numBatches uint,
 }
 
 // GetLastNBatchesByL2BlockNumber returns the last numBatches batches along with the l2 block state root by l2BlockNumber
-func (p *PostgresStorage) GetLastNBatchesByL2BlockNumber(ctx context.Context, l2BlockNumber uint64, numBatches uint, dbTx pgx.Tx) ([]*Batch, common.Hash, error) {
+// if the l2BlockNumber parameter is nil, it means we want to get the most recent last N batches
```
arnaubennassar, line 424: IMO since we already have a function that `get the most recent last N batches` I'd keep `GetLastNBatchesByL2BlockNumber` "clean" (without the particular behavior when `l2BlockNumber == nil`) and let the consumer choose which function has to use. Although this is just an opinion, here is my rationale:

- More predictable: what the function does is closer to its name
- Easier: 2 easy SQL queries vs 1 complex query. IMO it's easier to debug and maintain 2 easy

arnaubennassar (approved):

Issue #995, opened 2022-08-03
Closing batches shouldn't rely on last time a batch was virtualized


arnaubennassar:
Right now the sequencer closes batches if the time since a batch has been virtualized exceeds a config param.
This is a problem because there is a huge delay between a batch being closed and a sequences being sent to L1, therefore when this condition is evaluated to true for a batch, it will also be evaluated to true for the following batches causing many consecutive batches to have a single tx.

Instead we should close a batch if :

- it has transactions and
- has been open for more than X time and
- previous batch has been virtualized

PR #1058, opened 2022-08-18
changed closing batch condition

Mikelle:
Closes https://github.com/0xPolygonHermez/zkevm-node/issues/995

### Reviewers

Main reviewers:

- @arnaubennassar


sequencer/batchbuilder.go
```diff
@@ -333,20 +333,17 @@ func (s *Sequencer) shouldCloseSequenceInProgress(ctx context.Context) bool {
...
 	}
-	if lastBatchTime.Before(time.Now().Add(-s.cfg.LastTimeBatchMaxWaitPeriod.Duration)) && len(s.sequenceInProgress.Txs) > 0 {
-		isProfitable := s.isSequenceProfitable(ctx)
-		if isProfitable {
-			log.Info(
-				"current sequence should be closed because LastTimeBatchMaxWaitPeriod has been exceeded, " +
-					"there are pending sequences to be sent and they are profitable")
-			return true
-		}
+	if time.Unix(s.sequenceInProgress.Timestamp, 0).Add(s.cfg.MaxTimeForBatchToBeOpen.Duration).After(time.Now()) &&
```
arnaubennassar, line 341:
```suggestion
	if time.Unix(s.sequenceInProgress.Timestamp, 0).Add(s.cfg.MaxTimeForBatchToBeOpen.Duration).Before(time.Now()) &&
```

If I understand it well, it should be before: `time to close = sequence in progress timestamp + max open time`.
Let's say `time to close = 10:00`, then:

- if the current time is 10:30, we should close, because we should have closed 30 mins ago already! Note that in that case 10:00 is before 10:30
- If the current time is 9:30, we shouldn't close as we have 30 more minutes to reach the time to close. In this case 10:00 is after 9:30, or what is the same is not before

arnaubennassar (changes requested):

arnaubennassar (approved):

Issue #1061, opened 2022-08-19
jRPC is not taking the pool into account for pending block requests


tclemos:
`ethclient.PendingNonceAt` is not returning the correct nonce when the account has pending txs in the pool.

We need to check all the jRPC methods that allow the `pending` block to be specified to make sure we are also consuming the `pool` to give the proper answer, at the moment we are handling the `pending` as the `latest`

tclemos:
These are the methods that need to be reviewed:

- [ ] eth_call
- [ ] eth_estimateGas
- [ ] eth_getBalance
- [ ] eth_getCode
- [ ] eth_getStorageAt
- [ ] eth_getTransactionCount
- [ ] eth_getBlockTransactionCountByNumber

tclemos:
Since the methods `eth_call`, `eth_estimateGas`, `eth_getBalance`, `eth_getCode`, `eth_getStorageAt` should work directly with a `pending` state, we understand in our consensus protocol it means consuming the `trusted state`.

The current implementation of the node already does this for these methods, so we are closing this issue due to this reason.

This is the issue to track the tests: #1106

PR #1084, opened 2022-08-24
fix eth_transactionCount for pending block

tclemos:
The issue this PR is involved is the #1061

### What does this PR do?

fix the jRPC `eth_getTransactionCount` method to handle `pending` block properly and consider the transactions into the pool.

### Reviewers

Main reviewers:

@arnaubennassar 
@KonradIT 
@Mikelle

### DIFF
--- a/jsonrpc/eth.go
+++ b/jsonrpc/eth.go
@@ -449,17 +449,30 @@ func (e *Eth) GetTransactionByHash(hash common.Hash) (interface{}, rpcError) {
 // GetTransactionCount returns account nonce
 func (e *Eth) GetTransactionCount(address common.Address, number *BlockNumber) (interface{}, rpcError) {
 	return e.txMan.NewDbTxScope(e.state, func(ctx context.Context, dbTx pgx.Tx) (interface{}, rpcError) {
+		var pendingNonce uint64
+		var nonce uint64
 		var err error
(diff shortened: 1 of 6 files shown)


arnaubennassar (approved):
Just wondering: what happens if the greatest nonce in the pool for a given account is 10, but there is no nonce 6, 7, 8, 9. And current nonce is 5.

As per your code I understand that we would return 10, but is it the same behaviour as `go-ethereum`?

tclemos:
> Just wondering: what happens if the greatest nonce in the pool for a given account is 10, but there is no nonce 6, 7, 8, 9. And current nonce is 5.
> 
> As per your code I understand that we would return 10, but is it the same behaviour as `go-ethereum`?

This was my first approach when creating the E2E test, but geth looks like has a different way to add txs to the pool, it uses a queue and as far as I know, does not add the tx to the pool until it is ready to be in the pool. So if you send a tx with a future nonce, it stays in the queue and the pending nonce is not affected until the nonce sequence is fulfilled.

I tried testing geth in different ways, but this was the way I've found to be able to test both L1 and L2 affecting the nonce related to the pool with the exact same code.

Theoretically I assumed that the way we do to add txs to the pool is different to the way geth does, so I thought the current implementation should be fair enough until we have a real case where it doesn't fit.

PR #1114, opened 2022-09-02
Reprocess discarded batch by OOC error in batchbuilder

### LINKED ISSUE #1102: Properly manage executor run out of counters error

ToniRamirezM:
Closes #1102 

### What does this PR do?

Reprocessed processedTx in a batch when an OOC error happens during batch building.

### Reviewers

Main reviewers:

- @arnaubennassar 
- @Mikelle

### DIFF
--- a/sequencer/batchbuilder.go
+++ b/sequencer/batchbuilder.go
@@ -75,21 +75,19 @@ func (s *Sequencer) tryToProcessTx(ctx context.Context, ticker *time.Ticker) {
 		return
 	}
 
-	if unprocessedBatch {
-		// An Out Of Counter error has happened
+	closeBatch := unprocessedBatch
+
+	for unprocessedBatch {
+		// The entire batch hasn't been processed
 		// processedTxs must be reprocessed as the batch was discarded by the executor
 		reprocessTxs := make([]*pool.Transaction, len(processedTxs))
-		for _, pTxHash := range processedTxsHashes {
-			tx := s.getPendingTxByHash(s.pendingTxs, pTxHash)
+		for _, pTx := range processedTxs {
+			tx := s.getPendingTxByHash(s.pendingTxs, pTx.TxHash)
 			reprocessTxs = append(reprocessTxs, tx)
 		}
-		reprocessedTxs, _, unreprocessedTxs, unreprocessedBatch, err := s.processTxs(ctx, reprocessTxs)
+		processedTxs, _, _, unprocessedBatch, err = s.processTxs(ctx, reprocessTxs)
 		if err != nil {
 			log.Errorf("failed to reprocess txs, err: %w", err)
-			return
-		}
-		if len(reprocessedTxs) != len(reprocessTxs) || len(unreprocessedTxs) > 0 || unreprocessedBatch {
-			log.Error("failed to reprocess txs")
 			return
 		}
 	}
@@ -120,6 +118,22 @@ func (s *Sequencer) tryToProcessTx(ctx context.Context, ticker *time.Ticker) {
 	s.pendingTxs = []*pool.Transaction{}
 	s.pendingTxsHashes = []string{}
 	s.sumZkCounters = pool.ZkCounters{}
+
+	if closeBatch {
+		err := s.closeSequence(ctx)
+		if errors.As(err, &state.ErrClosingBatchWithoutTxs) {
+			log.Info("current sequence can't be closed without transactions")
+			waitTick(ctx, ticker)
+			return
+		} else if err != nil {
+			log.Errorf("error closing sequence: %v", err)
+			log.Info("resetting sequence in progress")
+			if err = s.loadSequenceFromState(ctx); err != nil {
+				log.Error("error loading sequence from state: %v", err)
+			}
+			return
+		}
+	}
 }
 
 func (s *Sequencer) newSequence(ctx context.Context) (types.Sequence, error) {
--- a/sequencer/batchbuilder.go
+++ b/sequencer/batchbuilder.go
@@ -16,6 +16,13 @@ import (
 	"github.com/jackc/pgx/v4"
 )
 
+type processTxResponse struct {
+	processedTxs       []*state.ProcessTransactionResponse
+	processedTxsHashes []common.Hash
+	unprocessedTxs     map[string]*state.ProcessTransactionResponse
+	isUnprocessedBatch bool
+}
+
 func (s *Sequencer) tryToProcessTx(ctx context.Context, ticker *time.Ticker) {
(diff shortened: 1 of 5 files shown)


sequencer/batchbuilder.go
```diff
@@ -69,11 +69,31 @@ func (s *Sequencer) tryToProcessTx(ctx context.Context, ticker *time.Ticker) {
...
 	}
-	processedTxs, processedTxsHashes, unprocessedTxs, err := s.processTxs(ctx, s.pendingTxs)
+	processedTxs, processedTxsHashes, unprocessedTxs, unprocessedBatch, err := s.processTxs(ctx, s.pendingTxs)
 	if err != nil {
 		log.Errorf("failed to process txs, err: %w", err)
 		return
 	}
+
+	if unprocessedBatch {
+		// An Out Of Counter error has happened
```
arnaubennassar, line 79:
```suggestion
		// The entire batch hasn't been processed
```

sequencer/batchbuilder.go
```diff
@@ -69,11 +69,31 @@ func (s *Sequencer) tryToProcessTx(ctx context.Context, ticker *time.Ticker) {
...
 		log.Infof("processing tx: %s", tx.Hash())
 	}
-	processedTxs, processedTxsHashes, unprocessedTxs, err := s.processTxs(ctx, s.pendingTxs)
+	processedTxs, processedTxsHashes, unprocessedTxs, unprocessedBatch, err := s.processTxs(ctx, s.pendingTxs)
 	if err != nil {
 		log.Errorf("failed to process txs, err: %w", err)
 		return
 	}
+
+	if unprocessedBatch {
```
arnaubennassar, line 78:
```suggestion
	for unprocessedBatch {
```

sequencer/batchbuilder.go
```diff
@@ -69,11 +69,31 @@ func (s *Sequencer) tryToProcessTx(ctx context.Context, ticker *time.Ticker) {
...
+		// An Out Of Counter error has happened
+		// processedTxs must be reprocessed as the batch was discarded by the executor
+		reprocessTxs := make([]*pool.Transaction, len(processedTxs))
+		for _, pTxHash := range processedTxsHashes {
+			tx := s.getPendingTxByHash(s.pendingTxs, pTxHash)
+			reprocessTxs = append(reprocessTxs, tx)
+		}
+		reprocessedTxs, _, unreprocessedTxs, unreprocessedBatch, err := s.processTxs(ctx, reprocessTxs)
```
arnaubennassar, line 86: Use the same variables name so you can do `=` instead of `:=` and use them across iterations until the batch is processed successfully

sequencer/batchbuilder.go
```diff
@@ -69,11 +69,31 @@ func (s *Sequencer) tryToProcessTx(ctx context.Context, ticker *time.Ticker) {
...
+		if err != nil {
+			log.Errorf("failed to reprocess txs, err: %w", err)
+			return
+		}
+		if len(reprocessedTxs) != len(reprocessTxs) || len(unreprocessedTxs) > 0 || unreprocessedBatch {
+			log.Error("failed to reprocess txs")
+			return
+		}
+	}
+
```
arnaubennassar, line 96: Store the `unprocessedBatch` result of the first call in a separated var. Then close the batch if we've entered the for loop

arnaubennassar (changes requested):

arnaubennassar (approved):

sequencer/batchbuilder.go
```diff
@@ -69,11 +69,29 @@ func (s *Sequencer) tryToProcessTx(ctx context.Context, ticker *time.Ticker) {
...
 		log.Errorf("failed to process txs, err: %w", err)
 		return
 	}
+
+	closeBatch := unprocessedBatch
+
+	for unprocessedBatch {
+		// The entire batch hasn't been processed
+		// processedTxs must be reprocessed as the batch was discarded by the executor
+		reprocessTxs := make([]*pool.Transaction, len(processedTxs))
```
Mikelle, line 83: In this case slice looks like this: `len processedTxs = 2 [*pool.Transaction{}, *pool.Transaction{}, appendedTxs...]`. You should init slice with capacity, not with length.
```suggestion
		reprocessTxs := make([]*pool.Transaction, 0, len(processedTxs))
```

sequencer/batchbuilder.go
```diff
@@ -171,11 +205,11 @@ func (s *Sequencer) isSequenceProfitable(ctx context.Context) bool {
 }
 
 func (s *Sequencer) processTxs(ctx context.Context, pendingTxs []*pool.Transaction) (
-	[]*state.ProcessTransactionResponse, []common.Hash, map[string]*state.ProcessTransactionResponse, error) {
+	[]*state.ProcessTransactionResponse, []common.Hash, map[string]*state.ProcessTransactionResponse, bool, error) {
```
Mikelle, line 208: Amount of variables to return has become too large. I think, it will be better to return struct here
```
type processTxsResponse struct {
  processedTxs []*state.ProcessTransactionResponse
  processedTxsHashes []common.Hash
  unprocessedTxs map[string]*state.ProcessTransactionResponse
  isBatchNotProcessed bool
}  
```

sequencer/batchbuilder.go
ToniRamirezM: Hi!

I did this change but I don't really understand what the problem is not initializaing the capacity in this case, as we are sure of the final length of the slice. In fact I would init it to a len(processTxs) value instead of 0. 

Can you give a more in deep explanation for this?

Thanks!

Mikelle: You could see the behavior here - https://go.dev/play/p/wSoQEzIrEhf

sequencer/batchbuilder.go
```diff
@@ -171,11 +216,17 @@ func (s *Sequencer) isSequenceProfitable(ctx context.Context) bool {
 }
 
 func (s *Sequencer) processTxs(ctx context.Context, pendingTxs []*pool.Transaction) (
-	[]*state.ProcessTransactionResponse, []common.Hash, map[string]*state.ProcessTransactionResponse, error) {
+	processTxResponse, error) {
+	response := processTxResponse{
+		processedTxs:       nil,
+		processedTxsHashes: nil,
+		unprocessedTxs:     nil,
+		isUnprocessedBatch: true,
```
Mikelle, line 224: if you change `isUnprocessedBatch -> isBatchProcessed`, you will not need to init this struct and could just return `&processTxResponse`

Mikelle (approved):

PR #1127, opened 2022-09-06
Properly check tx hash in CheckSupersetBatchTransactions

### LINKED ISSUE #902: state.CheckSupersetBatchTransactions should compare txHash field and not tx.Hash() method
It looks like the value in txHash is not correct, so for the moment the code will compare against tx.Hash()

The corresponding test has been commented so it does not get executed. This should be uncommented once the tx hash is properly calculated.

https://github.com/0xPolygonHermez/zkevm-node/pull/901/files#diff-ad95c17c31b539a3e385a23d86bbecf4d8a925ba6679aec2e66b5202ba4c4f81R797

ToniRamirezM:
Closes #902 

### What does this PR do?

Checks TX returned from the executor instead of reusing the one provided by the node to the executor in state.CheckSupersetBatchTransactions

### Reviewers

Main reviewers:

- @arnaubennassar 
- @tclemos 
- @Mikelle


Mikelle (approved):

ToniRamirezM:
Setting this PR to draft until we ensure the executor properly generates the Tx Hash in all cases.

tclemos (approved):

PR #1141, opened 2022-09-12
check genesis root

ARR552:
### What does this PR do?

This PR adds the check for genesis root even if it is loaded from a file.

### Reviewers

Main reviewers:

- @ToniRamirezM 
- @tclemos 
- @cool-develope


arnaubennassar:
I guess that this works using the merge option. Instead, it would be amazing if:

- Addresses are loaded from the config file
- Genesis is loaded from a separated config file (mandatory parameter)
- Remove `config/genesis.go`
- Remove `tools/genesis`

cool-develope (approved):
I agree with @arnaubennassar, the genesis mechanism is a little messy right now, we need to discuss it in team-wide meeting.

arnaubennassar (approved):

PR #1155, opened 2022-09-13
Database Split

### LINKED ISSUE #649: RPC should use the state in "read only" mode
This may be already the case, but we have to ensure this in order to use a setup where RPCs are connected to read only replicas of the state and MT DBs.

Of course they will still need to be connected to write replicas for the pool.

This will bring the ability to have multiple RPC instances with a single synchronizer

tclemos:
Closes #649.

### What does this PR do?

Split the single DB into 3 new ones: State, Pool and RPC.

### Reviewers

Main reviewers:

@arnaubennassar
@ToniRamirezM 
@Mikelle 
@KonradIT 
@ARR552 
@cool-develope


ARR552 (approved):

KonradIT (approved):

docker-compose.yml
```diff
@@ -7,10 +7,14 @@ services:
     container_name: zkevm-sequencer
     image: zkevm-node
     environment:
-      - ZKEVM_NODE_DATABASE_USER=test_user
-      - ZKEVM_NODE_DATABASE_PASSWORD=test_password
-      - ZKEVM_NODE_DATABASE_NAME=test_db
-      - ZKEVM_NODE_DATABASE_HOST=zkevm-db
+      - ZKEVM_NODE_STATEDB_USER=state_user
```
arnaubennassar, line 10: Suggestion: this config is already defined in config files. We can rely on it instead of having it on the docker-compose

arnaubennassar (changes requested):
Great job :rocket:

Mikelle (approved):

ToniRamirezM (approved):

arnaubennassar (approved):

Issue #1166, opened 2022-09-27
Sequencer seems not to be marking transactions as selected


ToniRamirezM:
It looks like the sequencer keeps selecting already processed transactions that were properly executed.

ToniRamirezM:
It was due to the executor not properly calculating tx hashes.

PR #1172, opened 2022-09-27
fix not updated status for the tx

Mikelle:
Closes https://github.com/0xPolygonHermez/zkevm-node/issues/1166

### Reviewers

Main reviewers:

- @ToniRamirezM


ToniRamirezM (approved):

tclemos (approved):

ToniRamirezM:
It was due to the executor not properly calculating tx hashes. This change is not needed as the executor is now properly working.

Issue #1146, opened 2022-09-13
Improve selection strategy


arnaubennassar:
- [x] RPC should execute txs before adding into the pool, and store counter usage
- [x] Sequencer should select txs based on counter usage and gas price

PR #1190, opened 2022-09-29
Feature/selection improvement

Mikelle:
Closes https://github.com/0xPolygonHermez/zkevm-node/issues/1146

### Reviewers

Main reviewers:

- @ToniRamirezM 
- @arnaubennassar


ToniRamirezM:
ChainId is mandatory for executor and prover now. 

When #1201 is merged this PR should be updated reflecting this, for example en the method state.ProcessTx.

ToniRamirezM (approved):

sequencer/batchbuilder.go
```diff
@@ -497,26 +504,45 @@ func (s *Sequencer) appendPendingTxs(ctx context.Context, isClaims bool, minGasP
...
+			UsedSteps:            s.cfg.MaxSteps,
+		}) {
+			log.Info("reached max zkCounters, need to close a batch")
+			s.sequenceInProgress.ZkCounters = zkCountersBeforeAddition
+			s.sequenceInProgress.IsZkCountersReachedMax = true
+			break
+		}
+		zkCountersBeforeAddition = s.sequenceInProgress.ZkCounters
+		s.sequenceInProgress.ZkCounters.SumUpZkCounters(pendTxs[i].ZkCounters)
+		pendTxsPreSelected = append(pendTxsPreSelected, pendTxs[i].Transaction)
```
arnaubennassar, line 542: ZK counters should be used before executing (to estimate a realistic batch at first shot). Checking after executing should be redundant: if any counter is exceeded, executor shall return an OOC

sequencer/closingchecker.go
```diff
@@ -11,6 +11,11 @@ import (
 // shouldCloseSequenceInProgress checks if sequence should be closed or not
 // in case it's enough blocks since last GER update, long time since last batch and sequence is profitable
 func (s *Sequencer) shouldCloseSequenceInProgress(ctx context.Context) bool {
+	if s.sequenceInProgress.IsZkCountersReachedMax {
+		log.Info("current sequence should be closed because it has reached max zk counters")
+		s.sequenceInProgress.IsZkCountersReachedMax = false
+		return true
+	}
```
arnaubennassar, line 18: Similar to previous coment we should never see any counte being exceeded, as before this happens the executor stops and returns an OOC. On the other hand it may be interesting to have some loic to close a batch if it's close to exceed some counters (example counter usage over 80%), but I wouldn't implement something like this yet

state/state.go
```diff
@@ -502,6 +502,50 @@ func (s *State) processBatch(ctx context.Context, batchNumber uint64, batchL2Dat
 	return res, err
 }
 
+// ProcessTx send batch with one tx to the executor to get zk counters
+func (s *State) ProcessTx(ctx context.Context, tx types.Transaction, dbTx pgx.Tx) (*ProcessBatchResponse, error) {
```
arnaubennassar, line 506: @ToniRamirezM WDYT about this function? I've the feeling that this logic it's already implemented somewhere else

arnaubennassar (changes requested):

etherman/types/sequence.go
```diff
@@ -14,6 +15,8 @@ type Sequence struct {
 	Timestamp                                int64
 	Txs                                      []types.Transaction
 	IsSequenceTooBig                         bool
+	ZkCounters                               pool.ZkCounters
+	IsZkCountersReachedMax                   bool
```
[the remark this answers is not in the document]
Mikelle: This type fixed by Thiago - https://github.com/0xPolygonHermez/zkevm-node/pull/1330/files#diff-3dfedcd1320099edf07cfdbdcdddf7c76942e7400397abfaf410cf286b081edfL12

jsonrpc/eth.go
```diff
@@ -624,14 +627,30 @@ func (e *Eth) relayTxToSequencerNode(input string) (interface{}, rpcError) {
...
 
-func (e *Eth) tryToAddTxToPool(input string) (interface{}, rpcError) {
+func (e *Eth) tryToAddTxToPool(input string, dbTx pgx.Tx) (interface{}, rpcError) {
 	tx, err := hexToTx(input)
 	if err != nil {
 		return rpcErrorResponse(invalidParamsErrorCode, "invalid tx input", err)
 	}
+	ctx := context.Background()
+	procResp, err := e.state.ProcessTx(ctx, *tx, dbTx)
+	if err != nil {
```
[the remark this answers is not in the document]
Mikelle: > We should return error for txs that won't be processed ever (nonce too low, ...)

For nonce is too low we have validateTx check, which returns an error in that case

> It could be interesting to add the tx as failed to the pool if the execution fails

Added

PR #1249, opened 2022-10-10
Dont reduce precision when estimating gas for eth transfers

### LINKED ISSUE #1247: Estimate gas for ETH transfer should be 21000
After reducing the precision of the estimate gas function, simple ETH transfers are suggested to consume 23625 gas units.
In general, it's OK to overestimate by ~2k gas units, however, most users know that a simple ETH transfer costs 21k, and they could be confused if they see this number

ToniRamirezM:
Closes #1247 

### What does this PR do?

Does not reduce precision when estimating gas for eth transfers

### Reviewers

Main reviewers:

- @arnaubennassar 
- @tclemos

### DIFF
--- a/state/state.go
+++ b/state/state.go
@@ -148,6 +148,15 @@ func (s *State) EstimateGas(transaction *types.Transaction, senderAddress common
 
 	if lowEnd != ethTransferGas {
 		reducePrecision = true
+
+		if transaction.To() != nil {
+			code, err := s.tree.GetCode(ctx, *transaction.To(), l2BlockStateRoot.Bytes())
+			if err != nil {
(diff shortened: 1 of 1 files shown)


state/state.go
```diff
@@ -143,6 +146,10 @@ func (s *State) EstimateGas(transaction *types.Transaction, senderAddress common
 		return 0, err
 	}
 
+	if lowEnd != ethTransferGas {
```
arnaubennassar, line 149:
```suggestion
	if lowEnd == ethTransferGas {
```

arnaubennassar (changes requested):

arnaubennassar (approved):

PR #1255, opened 2022-10-10
Check added to checkTrustedState

### LINKED ISSUE #1250: Wrong root from trusted state
After some time running, the prover detects that the roots don't match.

ARR552:
Closes #1250 

### What does this PR do?

This PR adds a check for the state root calculated by the sequencer.
If there is a problem with the state root, it will be detected and the synchronizer will reset the trusted state.

### Reviewers

- @cool-develope 
- @arnaubennassar 
- @ToniRamirezM


cool-develope:
`ExecuteBatch` is missing in the stateMock

cool-develope (approved):
I approved but we need to investigate the test failures

ToniRamirezM (approved):

PR #1265, opened 2022-10-11
adding ooc handling

Mikelle:
Adding handling of the case, when batch consists of the failed intrinsic value transaction and meets OOC error. Here batchbuilder is trying to reprocess only tx with OOC error. If it fails to be processed again, tx is marked as invalid.
### Reviewers

Main reviewers:

- @arnaubennassar


ToniRamirezM:
This login should be revised once this issue is closed:

https://github.com/0xPolygonHermez/zkevm-node/issues/1304

PR #1276, opened 2022-10-13
Fix non e2e tests

### LINKED ISSUE #1274: non e2e tests are not passing

ToniRamirezM:
Closes #1274

### What does this PR do?

Fixes non e2e tests

### Reviewers

Main reviewers:

- @arnaubennassar 
- @KonradIT

### DIFF
--- a/pool/pool_test.go
+++ b/pool/pool_test.go
@@ -383,7 +383,7 @@ func Test_GetTopFailedTxsByProfitabilityAndZkCounters(t *testing.T) {
 	txs, err := p.GetTxs(ctx, pool.TxStatusFailed, false, 1, 10)
 	require.NoError(t, err)
 	// bcs it's sorted by nonce, tx with the lowest nonce is expected here
-	assert.Equal(t, txsCount/2, len(txs))
+	assert.Equal(t, txsCount, len(txs))
 }
 
 func Test_UpdateTxsStatus(t *testing.T) {
--- a/pool/transaction.go
+++ b/pool/transaction.go
@@ -1,8 +1,8 @@
 package pool
 
 import (
-	"time"
 	"strings"
+	"time"
 
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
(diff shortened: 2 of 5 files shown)


pool/pool_test.go
```diff
@@ -383,7 +383,7 @@ func Test_GetTopFailedTxsByProfitabilityAndZkCounters(t *testing.T) {
 	txs, err := p.GetTxs(ctx, pool.TxStatusFailed, false, 1, 10)
 	require.NoError(t, err)
 	// bcs it's sorted by nonce, tx with the lowest nonce is expected here
-	assert.Equal(t, txsCount/2, len(txs))
+	assert.Equal(t, txsCount, len(txs))
```
ARR552, line 386: @Mikelle  let me know what you think

Mikelle: No, it shouldn't be that way. In the previous step half of the txs were updated to Invalid status. So here it should be 5 txs.

pool/transaction.go
```diff
@@ -77,8 +78,8 @@ func (tx *Transaction) IsClaimTx(l2BridgeAddr common.Address) bool {
 		return false
 	}
 
-	if *tx.To() == l2BridgeAddr { // &&
-		// strings.HasPrefix("0x"+common.Bytes2Hex(tx.Data()), bridgeClaimMethodSignature)  {
+	if *tx.To() == l2BridgeAddr &&
+		strings.HasPrefix("0x"+common.Bytes2Hex(tx.Data()), bridgeClaimMethodSignature) {
```
Mikelle, line 82: Not sure about this change, bcs it was commented in this commit - https://github.com/0xPolygonHermez/zkevm-node/commit/4e03c05d6900b7937cef9a6b40ae2b99bb4f8aa9#diff-b87b703deb53896d58cb942d1585ea5a40f705af35f3e12e92c8b3e2548a092bR81. @arnaubennassar what do you thing?

KonradIT (approved):
🔥🔥🔥 🚀

pool/transaction.go
ARR552: I've already commented to him. A claimable tx goes to the l2bridge smc and to the method claim (it is filtered by the signature). If we remove that check we will consider deposits as claimables (gas 0)

Mikelle (approved):

arnaubennassar (approved):

PR #1321, opened 2022-10-21
Etherman optionally readonly.

### LINKED ISSUE #1227: Remove account requirement for nodes that are not sending sequences or proofs
The current implementation requires an account to be configured(aka requires a .keystore file) regardless of the node type.

This will cause unnecessary friction for users.

We could have a Read Only Etherman for nodes that are only synchronizing data and redirecting TXs to the trusted sequencer.

In order to achieve this, we basically need to add a check at the beginning of the methods that really depend on the account and make the account to be optional.

wdyt @arnaubennassar?

kind84:
Closes #1227.

### What does this PR do?

This PR implements the option to set Etherman to be readonly (a.k.a. without an account). This is useful for nodes that don't send transactions.

The behaavior is controlled by the new `ReadOnly` field in the Etherman config (`false` by default).

Methods accessing the account are now safeguarded by a check on the `readOnly` field in the `Etherman` struct.

Bonus: added the Etherman test cases in `config/config_test.go`

### Reviewers

Main reviewers:

- @tclemos 
- @Psykepro

### DIFF
--- a/etherman/etherman.go
+++ b/etherman/etherman.go
@@ -49,7 +49,8 @@ var (
 	// ErrNotFound is used when the object is not found
 	ErrNotFound = errors.New("Not found")
 	// ErrIsReadOnlyMode is used when the EtherMan client is in read-only mode.
-	ErrIsReadOnlyMode = errors.New("etherman client in read-only mode")
+	ErrIsReadOnlyMode = errors.New("Etherman client in read-only mode: no account configured to send transactions to L1. " +
+		"Please check the [Etherman] PrivateKeyPath and PrivateKeyPassword configuration.")
 )
 
 // EventOrder is the the type used to identify the events order
--- a/etherman/etherman.go
+++ b/etherman/etherman.go
@@ -92,11 +92,7 @@ type Client struct {
 
 	GasProviders externalGasProviders
 
-	// readOnly if `true` the client won't have an account.
-	// Check this field before trying to access the `auth` field.
-	readOnly bool
-
-	auth *bind.TransactOpts
+	auth *bind.TransactOpts // nil in case of read-only client
 }
 
 // NewClient creates a new etherman.
@@ -144,8 +140,13 @@ func NewClient(cfg Config, auth *bind.TransactOpts) (*Client, error) {
 			MultiGasProvider: cfg.MultiGasProvider,
 			Providers:        gProviders,
 		},
-		auth: auth}, nil
+		auth: auth,
+	}, nil
 }
+
+// IsReadOnly returns whether the EtherMan client is in read-only mode.
+// Call this before trying to access the `auth` field.
+func (c *Client) IsReadOnly() bool { return c.auth == nil }
 
 // GetRollupInfoByBlockRange function retrieves the Rollup information that are included in all this ethereum blocks
 // from block x to block y.
@@ -271,7 +272,7 @@ func (etherMan *Client) WaitTxToBeMined(ctx context.Context, tx *types.Transacti
 
 // EstimateGasSequenceBatches estimates gas for sending batches
 func (etherMan *Client) EstimateGasSequenceBatches(sequences []ethmanTypes.Sequence) (*types.Transaction, error) {
-	if etherMan.readOnly {
+	if etherMan.IsReadOnly() {
 		return nil, ErrIsReadOnlyMode
 	}
 	noSendOpts := *etherMan.auth
@@ -286,7 +287,7 @@ func (etherMan *Client) EstimateGasSequenceBatches(sequences []ethmanTypes.Seque
 
 // SequenceBatches send sequences of batches to the ethereum
 func (etherMan *Client) SequenceBatches(ctx context.Context, sequences []ethmanTypes.Sequence, gasLimit uint64, gasPrice, nonce *big.Int) (*types.Transaction, error) {
-	if etherMan.readOnly {
+	if etherMan.IsReadOnly() {
 		return nil, ErrIsReadOnlyMode
 	}
 	sendSequencesOpts := *etherMan.auth
@@ -331,7 +332,7 @@ func (etherMan *Client) sequenceBatches(opts *bind.TransactOpts, sequences []eth
 
 // EstimateGasForVerifyBatch estimates gas for verify batch smart contract call
 func (etherMan *Client) EstimateGasForVerifyBatch(batchNumber uint64, resGetProof *pb.GetProofResponse) (uint64, error) {
-	if etherMan.readOnly {
+	if etherMan.IsReadOnly() {
 		return 0, ErrIsReadOnlyMode
 	}
 	verifyBatchOpts := *etherMan.auth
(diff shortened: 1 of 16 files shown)


Psykepro:
lgtm just 1 thing: I think you need to add the new field also in `test/config/config.test.toml`

kind84:
@Psykepro nice catch! fixed here 4d2a55a84b3f49fdf9180328348e5fc043063a66
I've also added the other missing fields.

Psykepro (approved):

etherman/etherman.go
```diff
@@ -245,6 +260,9 @@ func (etherMan *Client) EstimateGasSequenceBatches(sequences []ethmanTypes.Seque
 
 // SequenceBatches send sequences of batches to the ethereum
 func (etherMan *Client) SequenceBatches(sequences []ethmanTypes.Sequence, gasLimit uint64, gasPrice, nonce *big.Int) (*types.Transaction, error) {
+	if etherMan.readOnly {
+		return nil, errors.New("etherman client in read-only mode, cannot send sequences of batches")
+	}
```
ARR552, line 265: I'm worried about the sequencer and aggregator behavior when this error is returned

log/log.go
```diff
@@ -38,6 +39,27 @@ func Init(cfg Config) {
...
+	}
+	return observedLogs, nil
+}
+
+// DeinitTestLogger deinitializes the test logger and restore the default one.
+func DeinitTestLogger() {
+	log = nil
+	getDefaultLog()
+}
+
```
ARR552, line 62: It is ok to see the logs. To avoid problems check how to initialize logs at the beginning of your test. You can take a look here.
https://github.com/0xPolygonHermez/zkevm-node/blob/717270900a03987a3b4c9e71736f00dbaf964c73/etherman/etherman_test.go#L23

log/log.go
```diff
@@ -85,6 +107,16 @@ func NewLogger(cfg Config) (*zap.SugaredLogger, *zap.AtomicLevel, error) {
...
+// NewTestLogger returns a new logger for testing purposes and the logs it observed.
+func NewTestLogger(cfg Config) (*zap.SugaredLogger, *observer.ObservedLogs, error) {
+	level, err := zapcore.ParseLevel(cfg.Level)
+	if err != nil {
+		return nil, nil, err
+	}
+	observedLogger, observedLogs := observer.New(level)
+	return zap.New(observedLogger).Sugar(), observedLogs, nil
+}
+
```
ARR552, line 119: same here. Check Init logs in other test

ARR552 (changes requested):

etherman/etherman.go
```diff
@@ -233,6 +245,9 @@ func (etherMan *Client) WaitTxToBeMined(hash common.Hash, timeout time.Duration)
 
 // EstimateGasSequenceBatches estimates gas for sending batches
 func (etherMan *Client) EstimateGasSequenceBatches(sequences []ethmanTypes.Sequence) (*types.Transaction, error) {
+	if etherMan.readOnly {
+		return nil, errors.New("etherman client in read-only mode, cannot estimate gas for sending batches")
```
KonradIT, line 249: Small suggestion, add a new error in line 48 like:

    ErrIsReadOnlyMode = errors.New("etherman client in read-only mode, cannot estimate gas for sending batches")

Then:

```suggestion
                return nil, ErrIsReadOnlyMode
````

Also for line 264.

**Review by KonradIT — APPROVED** [review]

File: etherman/etherman.go [hunk 5]
```diff
@@ -79,6 +79,10 @@ type Client struct {
 	Matic                 *matic.Matic
 	SCAddresses           []common.Address
 
+	// readOnly if `true` the client won't have an account.
+	// Check this field before trying to access the `auth` field.
+	readOnly bool
+
```
tclemos, line 85: My suggestion is to remove this configuration and if the key is not provided, the methods requiring the auth should fail, returning an error message.

it's ok to have a variable to control if it is read-only or not, but this variable should be related to the keystore config and not a new one.

in summary, if we have a keystore file, we instantiate it and provide it to etherman, so etherman knows it can perform actions that require authorization. If we don't have the keystore, we don't instantiate it and provide nil to etherman, so etherman knows it cannot perform actions that require authorization.

tclemos (changes requested):
please review the etherman.Readonly configuration

etherman/etherman.go
kind84: @KonradIT  good point, I applied it on every occurrence here ed8a793590e46a073c12307adb7b6801c883f38c

log/log.go
kind84: @ARR552 I think the comment for `InitTestLogger` was a bit misleading and I slightly modified it here a927ba45b7a6b3899c04312dd09d14e637f09dfc.

This logger is meant to be used in tests where you need to inspect the logs. You can see how I used it in this test: 
https://github.com/0xPolygonHermez/zkevm-node/blob/033877649b2fb3a088ea957b22c91fef656aaa40/ethtxmanager/ethtxmanager_test.go#L45

Anyway the latest version of that test has changed (see 85221c0d861788aef453b8e161ea722bdd590590) because, after rebasing on `develop`, the signature of the tested method has changed. Now it returns an error and so inspecting the logs is not required any more.
So my test-logger is not used any more, but I'd keep it in any case, it may become useful for other tests, WDYT?

I haven't found a way to use our default logger for this purpose (it wants either a URL or a file path for outputting the logs. I'm not sure a buffer can fit there)

I hope this clarifies a bit.

log/log.go
kind84: @ARR552 see my comment above.

etherman/etherman.go
kind84: thanks @tclemos, it makes totally sense. It should be better now 346ac5d85620d86e5991e6c3ae744f409262f51a, WDYT?

etherman/etherman.go
kind84: @ARR552 I added a test for the specific error you are referring to https://github.com/0xPolygonHermez/zkevm-node/blob/85221c0d861788aef453b8e161ea722bdd590590/ethtxmanager/ethtxmanager_test.go#L23

More broadly, in case of a read-only etherman client, the affected methods are:
- `EstimateGasSequenceBatches`: non-trivial to test. The error is handled by the caller.
- `SequenceBatches`: added UT. The caller handles the error.
- `EstimateGasForVerifyBatch`: method not used.
- `VerifyBatch`: added UT. The caller handles the error.
- `ApproveMatic`: error trivially handled in `cmd/approve.go`
- `GetPublicAddress`: not easy to test. The new error caused the method to change signature. I added the error handling here.
https://github.com/0xPolygonHermez/zkevm-node/blob/85221c0d861788aef453b8e161ea722bdd590590/aggregator/aggregator.go#L367-L370

I don't see major concerns, since the new error is handled in every case. But I clearly don't have the full picture. I trust you to see any blocking point.

log/log.go
ARR552: If it is not used I would preffer remove it just to avoid have unused code. We can always add it again when is required

kind84: Fair enough @ARR552 , removed here 7221252b96505e4cbf812ef36c61b6818cb0e66d

etherman/etherman.go
```diff
@@ -46,6 +46,8 @@ var (
 
 	// ErrNotFound is used when the object is not found
 	ErrNotFound = errors.New("Not found")
+	// ErrIsReadOnlyMode is used when the EtherMan client is in read-only mode.
+	ErrIsReadOnlyMode = errors.New("etherman client in read-only mode")
```
tclemos, line 50: I was thinking about having a better error message to allow the user to understand why this is running on read-only mode.

maybe if we complement the error message with the reason, something like `etherman client in read-only mode, because <add something here that makes sense>`, wdyt?

I was thinking something like this: "etherman client in read-only mode because there was no account configured to send transactions to L1. Please check the [Etherman] PrivateKeyPath and PrivateKeyPassword configuration."

tclemos (approved):

KonradIT (approved):

ARR552 (approved):

Issue #1262, opened 2022-10-11
Reuse logic in ethttxmanager


arnaubennassar:
The same logic for sending sequence and verify batch is duplicated, it would be nice to reuse it so we don't have to do changes twice when code needs to be changed

arnaubennassar:
Closing in favor of #1271

Issue #1245, opened 2022-10-10
Update timestamp if there aren't new transactions


arnaubennassar:
If the network is idle and the open batch has no txs, the timestamp should be periodically updated to avoid having a timestamp that is too old. This could be a reason to decide to close a batch, but this is already controlled through configuration

PR #1330, opened 2022-10-24
Revamp sequencer and aggregator integration with ethtxmanager

### LINKED ISSUE #1271: Better handling for sending L1 txs
Instead of sending txs synchronously from multiple points of the code (sequencer and aggregator), store the txs that need to be sent on a DB, and have an additional component that deals with sending those txs.

This should enable:

- Better security (only one component needs the private key)
- More resilient: components can go down, but the new component still keeps track of pending txs
- Better nonce management: on the DB we can store what nonce is used for each tx. If facing a mismatched nonce this could even be edited manually by the admins

tclemos:
Closes #1271 and #1262.

### Reviewers


ethtxmanager/ethtxmanager.go
```diff
@@ -7,157 +7,283 @@ package ethtxmanager
...
-				nonce = new(big.Int).SetUint64(tx.Nonce())
-				gasPrice = increaseGasPrice(tx.GasPrice(), c.cfg.PercentageToIncreaseGasPrice)
-				log.Infof("tx %s reached timeout, retrying with gas price = %d", tx.Hash(), gasPrice)
-				continue
+			// if the tx is already know, refresh the update date to give it more time to get mined
+			if errors.Is(err, core.ErrAlreadyKnown) {
+				err := c.state.UpdateSequenceGroupTx(ctx, sequenceGroup.TxHash, sequenceGroup.TxHash, nil)
+				if err != nil {
+					log.Errorf("give it more time to the sequence group related to the batches %v to get mined: %v", sequenceGroup.BatchNumbers, err)
+				}
```
ARR552, line 215: This sounds weird for me. If this error happens, it means that even the gasPrice of the resent tx is the same. In this case, wouldn't it be better to increase the gasPrice even more??

sequencer/sequencer.go
```diff
@@ -91,16 +90,26 @@ func (s *Sequencer) Start(ctx context.Context) {
...
 		}
 	}()
+	// send and monitor persisted sequences to L1
+	go func() {
+		for {
+			s.txManager.SyncPendingSequences()
+			waitTick(ctx, tickerSendSequence)
+		}
+	}()
+
```
ARR552, line 112: I thought the ethtxmanager should be autonomous

ethtxmanager/ethtxmanager.go
```diff
@@ -7,157 +7,283 @@ package ethtxmanager
...
+		// recalculated with the current prices of the network
+		tx, err := c.ethMan.VerifyBatch(proof.BatchNumber, proof.Proof, 0, nonce)
+		if err != nil {
+			// if the tx is already know, refresh the update date to give it more time to get mined
+			if errors.Is(err, core.ErrAlreadyKnown) {
+				err := c.state.UpdateProofTx(ctx, proof.BatchNumber, *proof.TxHash, nil)
+				if err != nil {
+					log.Errorf("give it more time to the proof related to the batch %v to get mined: %v", proof.BatchNumber, err)
+				}
+				return
```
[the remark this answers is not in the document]
tclemos: I don't know, we can add this feature, but I think this is not needed for now, we just need to make sure the tx get mined, and following the network costs, we can achieve this even if the tx takes a bit to get mined. 
Also, this error can be forced if the interval is too short.

ethtxmanager/ethtxmanager.go
tclemos: exaplained above

sequencer/sequencer.go
tclemos: I've explained this here: https://github.com/0xPolygonHermez/zkevm-node/issues/1271#issuecomment-1288968113

ToniRamirezM (approved):

sequencer/closingchecker.go
```diff
@@ -66,23 +66,15 @@ func (s *Sequencer) shouldCloseDueToNewDeposits(ctx context.Context) (bool, erro
...
-	}
-	isPreviousBatchVirtualized, err := s.state.IsBatchVirtualized(ctx, lastBatchNumber-1, nil)
-	if err != nil {
-		log.Errorf("failed to get last virtual batch num, err: %w", err)
-		return false, err
-	}
-	if time.Unix(s.sequenceInProgress.Timestamp, 0).Add(s.cfg.MaxTimeForBatchToBeOpen.Duration).Before(time.Now()) &&
-		isPreviousBatchVirtualized && len(s.sequenceInProgress.Txs) > 0 {
+func (s *Sequencer) shouldCloseTooLongSinceLastSequence(ctx context.Context) bool {
+	timeoutExpired := s.sequenceInProgress.Timestamp.Add(s.cfg.MaxTimeForBatchToBeOpen.Duration).Before(time.Now())
```
Mikelle, line 72: This check is not working as expected. The batch could be open any time long, but the time from the previous virtualized batch shouldn't be more, than a time.Now()-value from the config.
https://github.com/0xPolygonHermez/zkevm-node/issues/1245

tclemos: the sequencer logic was changed to not depend on virtualized batches, now it just depends on the sequences created in the DB.

Issue #1062, opened 2022-08-19
Revive sequencer test


Mikelle:
Revive commented unit tests for the sequencer (sequencer_internal_test.go)

PR #1332, opened 2022-10-25
Sequencer tests

Mikelle:
Closes https://github.com/0xPolygonHermez/zkevm-node/issues/1062

### Reviewers

Main reviewers:

- @ToniRamirezM


sequencer/closingchecker.go
```diff
@@ -12,12 +12,12 @@ import (
...
 	// Check if sequence is full
-	if s.isSequenceTooBig {
+	if s.sequenceInProgress.IsSequenceTooBig {
 		log.Infof("current sequence should be closed because it has reached the maximum data size")
-		s.isSequenceTooBig = false
+		s.sequenceInProgress.IsSequenceTooBig = false
 		return true
 	}
-	if len(s.sequenceInProgress.Txs) == int(maxTxsPerBatch) {
+	if len(s.sequenceInProgress.Txs) >= int(maxTxsPerBatch) {
```
ToniRamirezM, line 20: Can it be a problem if the number os transactions is > the max value?
Should this be handled in a separated way?

ToniRamirezM (approved):

sequencer/closingchecker.go
Mikelle: This problem shouldn't appear at all, bcs from the pool sequencer is always requesting the amount of txs (maxAmount - s.sequenceInProgress.Txs)

PR #1333, opened 2022-10-25
improve: adding commit hash as env var in docker and printing it along with the version on start

### LINKED ISSUE #1173: Logging improvements
- Do not log merkle tree
- Log version + commit hash on start

Psykepro:

Closes #1173.

### What does this PR do?

Getting the `COMMIT_HASH`  inside root `Makefile` and passing it to both `docker build` commands as argument `BUILD_COMMIT`. Exporting the commit hash as an environment variable inside the `Dockerfile`. Getting the variable inside `main.go` and printing the  `COMMIT_HASH` along with the `VERSION` at the beginning.

### Reviewers

Main reviewers:

- @KonradIT 
- @ARR552 
- @kind84 

Codeowner reviewers:

- @KonradIT
- @tclemos

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,6 +13,8 @@ RUN cd /src && make build
 
 # CONTAINER FOR RUNNING BINARY
 FROM alpine:3.16.0
+ARG BUILD_COMMIT
+ENV COMMIT_HASH $BUILD_COMMIT
 COPY --from=build /src/dist/zkevm-node /app/zkevm-node
 COPY --from=build /src/config/environments/local/local.node.config.toml /app/example.config.toml
 EXPOSE 8123
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,7 @@ GOBIN := $(GOBASE)/dist
 GOENVVARS := GOBIN=$(GOBIN) CGO_ENABLED=0 GOOS=linux GOARCH=amd64
 GOBINARY := zkevm-node
 GOCMD := $(GOBASE)/cmd
+$(eval BUILD_COMMIT:=$(shell git rev-parse --short HEAD))
 
 .PHONY: build
 build: ## Builds the binary locally into ./dist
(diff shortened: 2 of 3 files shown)


Dockerfile
```diff
@@ -1,6 +1,9 @@
 # CONTAINER FOR BUILDING BINARY
 FROM golang:1.17 AS build
 
+# ARGUMENTS
+ARG BUILD_COMMIT
+
```
tclemos, line 6: I think this is not needed

tclemos (approved):
just fix the linter issue and we are good

Makefile
```diff
@@ -8,18 +8,19 @@ GOBIN := $(GOBASE)/dist
 GOENVVARS := GOBIN=$(GOBIN) CGO_ENABLED=0 GOOS=linux GOARCH=amd64
 GOBINARY := zkevm-node
 GOCMD := $(GOBASE)/cmd
+$(eval BUILD_COMMIT:=$(shell git rev-parse HEAD))
```
kind84, line 11: @Psykepro do know if we need the full commit hash? Otherwise we could use `--short` flag.

Psykepro: @KonradIT what do you think do we need the short or the full?

Dockerfile
Psykepro: removed

Makefile
Psykepro: I've updated it to use the short for now.

cmd/main.go
```diff
@@ -34,7 +33,14 @@ const (
...
 )
 
+const (
+	//envCommitHash environment variable name for COMMIT_HASH
+	envCommitHash = "COMMIT_HASH"
+)
+
 var (
+	// commit represents the program based on the git commit
+	commit         = testutils.GetEnv(envCommitHash, "dev")
```
kind84, line 43: @Psykepro this `GetEnv` looks useful even outside tests. WDYT of moving it outside the `testutils` (maybe in a `utils` package)?.
This is not directly related to this PR (and maybe it can involve other utils functions) so feel free to resolve this comment and keep it for later.
CC @KonradIT

kind84 (approved):
LGTM

cmd/main.go
Psykepro: Yep, I was thinking of that because we have several functions in the `test` package which we use in other files like for example also the functions from `test/operations` outside of the test context. But as we discussed with @tclemos it is not good to have a too generic package like `utils` because it can get too big when everyone puts something there. But maybe we can have a package `environment` where to put this `GetEnv` function specifically. What do you think guys? @tclemos @kind84 @KonradIT @ARR552

KonradIT: It can stay as it is. Getting value from env is only used in testing scenarios (be it E2E tests or standalone tests), so it makes sense the code stays within the `test` folder. Same with `dbutils`, etc...

KonradIT (approved):

cmd/main.go
kind84: @Psykepro l close this thread and we can continue the discussion elsewhere as it seems to be a broader topic.

ARR552 (approved):

PR #1336, opened 2022-10-26
improve: making the timeout of the RPC server to be configurable and fixing deprecated func for creating temp file.

### LINKED ISSUE #614: Timeout for the RPC server should be configurable
Right now it's constant at one second

Psykepro:

Closes #614 .

### What does this PR do?

Making the `ReadTimeout` and `WriteTimeout` of the `RPC HTTP Server` to be configurable by expecting values in the `config file` in seconds with the following names:
`ReadTimeoutInSec`
`WriteTimeoutInSec`

And fixing deprecated function use in `config.go` for creating temp file.

### Reviewers

- @KonradIT 
- @kind84 
- @tclemos 
- @ARR552 

Main reviewers:

- @KonradIT 
- @kind84

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -2,7 +2,6 @@ package config_test
 
 import (
 	"flag"
-	"io/ioutil"
 	"math/big"
 	"os"
 	"reflect"
@@ -212,6 +211,14 @@ func Test_Defaults(t *testing.T) {
 			expectedValue: int(8123),
 		},
 		{
+			path:          "RPC.ReadTimeoutInSec",
+			expectedValue: time.Duration(1),
+		},
+		{
+			path:          "RPC.WriteTimeoutInSec",
+			expectedValue: time.Duration(1),
+		},
+		{
 			path:          "RPC.SequencerNodeURI",
 			expectedValue: "",
 		},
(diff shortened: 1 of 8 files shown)


config/config_test.go
```diff
@@ -268,7 +275,7 @@ func Test_Defaults(t *testing.T) {
 			expectedValue: 61090,
 		},
 	}
-	file, err := ioutil.TempFile("", "genesisConfig")
+	file, err := os.CreateTemp("", "genesisConfig")
```
ARR552, line 278: why do you think is better os.CreateTemp vs ioutil.TempFile??

config/default.go
```diff
@@ -47,6 +47,8 @@ PercentageToIncreaseGasLimit = 10
 [RPC]
 Host = "0.0.0.0"
 Port = 8123
+ReadTimeoutInSec = 1
+WriteTimeoutInSec = 1
```
ARR552, line 51: this values configured seems too short

config/config_test.go
Psykepro: because the ioutil.TempFile is deprecated from go 1.17:
image

:)

ToniRamirezM (approved):

kind84 (approved):
LGTM :ok_hand:

PR #1361, opened 2022-11-04
Txsender script

kind84:
### Background

For development purposes, it would be nice to have a script to send an arbitrary number of transactions.

Basically the behavior should be the same of the `TestEthTransfer` e2e test, but without dealing with creating/destroying the node local stack, which is assumed to be already running.

### What does this PR do?

This PR proposes to add a `txsender` script in `test/scripts` that accepts as argument the number of transactions to be sent (it defaults to 1 transaction).

- Can send either on layer 1, layer 2 or both, specifying the `--network` flag.
- Can wait for transactions to be validated using the `--wait flag`.

Please refer to the `README.md` file in the script folder to see some examples.

This PR also removes the `send_transfer` script as the one proposed here is a more flexible solution of that script.
In the future we may think about a full fledged CLI development application to include this and other useful script.

In order avoid duplication, the code shared between `TestEthTransfer` and the new script has been moved to `operations.ApplyL1Txs` and `operations.ApplyL2Txs`.

### Reviewers

Main reviewers:

@ARR552 
@KonradIT 
@arnaubennassar 
@ToniRamirezM

### DIFF
--- a/test/scripts/txsender/main.go
+++ b/test/scripts/txsender/main.go
@@ -76,6 +76,7 @@ Optionally it can wait for the transactions to be validated.`
 	txsender.Commands = []*cli.Command{
 		{
 			Name:    "send",
+			Before:  setLogLevel,
 			Aliases: []string{},
 			Usage:   "Sends the specified number of transactions",
 			Description: `This command sends the specified number of transactions.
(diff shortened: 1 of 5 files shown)


test/scripts/txsender/main.go
```diff
@@ -0,0 +1,210 @@
...
+		senderBalance, err := client.BalanceAt(ctx, auth.From, nil)
+		if err != nil {
+			return err
+		}
+		log.Debugf("ETH Balance for %v: %v", auth.From, senderBalance)
+
+		amount := big.NewInt(10) //nolint:gomnd
+		log.Debugf("Transfer Amount: %v", amount)
+
+		senderNonce, err := client.PendingNonceAt(ctx, auth.From)
```
ARR552, line 153: I would like to customize the starting nonce and the gasPrice. Could you add this tiny improvement? This will allow us to unblock transactions due to nonce or gasPrice problems if it happens. @ToniRamirezM what do you think?

ToniRamirezM: Yes, this could be usefull in some cases. I don't know how good idea is to put it here of leave that for another espeficic script.

ToniRamirezM (approved):

test/scripts/txsender/main.go
kind84: @ARR552 I resolve this in order to merge the PR. We can add that feature in a second PR if you agree.
I was thinking about adding a `--file` flag to parse transactions from a json file, WDYT?

PR #1372, opened 2022-11-08
Fix pointer

ToniRamirezM:
Fixes nil pointer in RPC during gas estimation.


arnaubennassar (approved):

ToniRamirezM:
Not needed after refactor of batches without l2block.

PR #1374, opened 2022-11-08
Add metrics to jsonrpc package

kind84:
What does this PR do?

Following https://github.com/0xPolygonHermez/zkevm-node/pull/892, this PR adds metrics to the `jsonrpc` package.

The goal here is to instrument the package with some the most relevant metrics, it doesn't mean all of them are covered. Please suggest any relevant metrics missing.

### Reviewers

@arnaubennassar 
@ToniRamirezM 
@KonradIT

### DIFF
--- a/jsonrpc/server_test.go
+++ b/jsonrpc/server_test.go
@@ -29,26 +29,6 @@ type mocks struct {
 	DbTx              *dbTxMock
 }
 
-func TestServerMetrics(t *testing.T) {
-	pool := newPoolMock(t)
-	state := newStateMock(t)
-	gasPriceEstimator := newGasPriceEstimatorMock(t)
-	storage := newStorageMock(t)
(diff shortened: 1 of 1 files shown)


kind84:
> LGTM! Just to make sure I've understood what it does:
> 
>     * It basically counts the requests grouped by  success / fail
> 
> 
> This is good enough for now, but I've interesting ideas to improve this innocent

@arnaubennassar yes, but as of now that metric is more for invalid/valid requests, regardless if then the valid request is successfully served or not.

This was the metrics as initially set up by @xavier-romero. 
Basically the counter has buckets for `invalid`, `single` and `batch` requests.
We didn't know if `OPTIONS` and `GET` requests were to be included and we left a `TODO` for them.

If you wish to get metrics for success/failure they need to be added separately.

PR #1401, opened 2022-11-16
Revamp logs

kind84:
### What does this PR do?

This PR proposes to revamp our `log` package. The idea is to have more information in the logs without having to add them manually on each log line.

This is accomplished by providing a way to get **child loggers** derived from the **root** one: 
- **Every function** that needs to inject new fields in the logs can do so by calling `WithFields` on the current logger. 
- **This returns** a new child logger that will print the provided fields on every log line (within the function scope) in addition to the parent logger fields.

### Example

```go
// use root logger
log.Debug("This is a debug log line from the root logger")

// derive a child logger from the root one, with additional fields
childLog := log.WithFields("txHash", "0xbanana", "chainId", 42)

// these two log lines will be decorated with the `txHash` and `chainId` fields
childLog.Debug("This is a debug log line")
childLog.Debug("This is another debug log line")

// derive a child logger from the previous child, adding another field
grandChildLog := childLog.WithFields("proverId", "prover123")

// this log line will be decorated with the `txHash`, `chainId` and `proverId` fields
grandChildLog.Debug("This is a debug log line from the granchild logger")

// derive another child logger from the root one, with another field
childLog2 := log.WithFields("batchNum": 8)

// this line will only have the `batchNum` field
childLog2.Debug("This is a debug log line from the second child logger")
```

**NOTE:** here 2 default fields have been added to **every** log line: `version` and `pid`.
They are purely demonstrative and can be removed/replaced with better ones.

Other logs have been decorated with fields inside `jsonrpc` and `metrics` package. The idea is to gradually start using the feature while developing.

**ENCODING**
Now the encoding of the logs can be specified in the config file choosing between `console` and `json`. The default value is `console`, but `json` can be used in a deployed environment for better logs ingestion.

### Bonus

The log message on startup has been polished a bit:
```sh
Version:      v0.0.1-RC1-5-g8538ace
Git revision: 8538ace
Git branch:   feature/revamp-logs
Go version:   go1.17.13
Built:        Fri, 02 Dec 2022 16:47:58 +0000
OS/Arch:      linux/amd64
```

The env var to pass the commit hash is not needed any more, instead it is sourced from the `Makefile` through a linker flag.

### Reviewers

Main reviewers:

@arnaubennassar 
@ToniRamirezM 
@tclemos 
@Psykepro 

CC: @xavier-romero

### DIFF
--- a/etherman/etherman.go
+++ b/etherman/etherman.go
@@ -104,6 +104,8 @@ type Client struct {
 	GasProviders externalGasProviders
 
 	auth *bind.TransactOpts // nil in case of read-only client
+
+	cfg Config
 }
 
 // NewClient creates a new etherman.
@@ -152,6 +154,7 @@ func NewClient(cfg Config, auth *bind.TransactOpts) (*Client, error) {
 			Providers:        gProviders,
 		},
 		auth: auth,
+		cfg:  cfg,
 	}, nil
 }
 
@@ -457,7 +460,6 @@ func (etherMan *Client) forcedBatchEvent(ctx context.Context, vLog types.Log, bl
 	}
 	msg, err := tx.AsMessage(types.NewLondonSigner(tx.ChainId()), big.NewInt(0))
 	if err != nil {
-		log.Error(err)
 		return err
 	}
 	if fb.Sequencer == msg.From() {
@@ -526,7 +528,6 @@ func (etherMan *Client) sequencedBatchesEvent(ctx context.Context, vLog types.Lo
 	}
 	msg, err := tx.AsMessage(types.NewLondonSigner(tx.ChainId()), big.NewInt(0))
 	if err != nil {
-		log.Error(err)
 		return err
 	}
 	sequences, err := decodeSequences(tx.Data(), sb.NumBatch, msg.From(), vLog.TxHash, msg.Nonce())
@@ -651,7 +652,6 @@ func (etherMan *Client) forceSequencedBatchesEvent(ctx context.Context, vLog typ
 	}
 	msg, err := tx.AsMessage(types.NewLondonSigner(tx.ChainId()), big.NewInt(0))
 	if err != nil {
-		log.Error(err)
 		return err
 	}
 	fullBlock, err := etherMan.EtherClient.BlockByHash(ctx, vLog.BlockHash)
--- a/etherman/etherman.go
+++ b/etherman/etherman.go
@@ -104,8 +104,6 @@ type Client struct {
 	GasProviders externalGasProviders
 
 	auth *bind.TransactOpts // nil in case of read-only client
-
-	cfg Config
 }
 
 // NewClient creates a new etherman.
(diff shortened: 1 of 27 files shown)


etherman/etherman.go
```diff
@@ -104,6 +104,8 @@ type Client struct {
 	GasProviders externalGasProviders
 
 	auth *bind.TransactOpts // nil in case of read-only client
+
+	cfg Config
```
tclemos, line 108: I don't understand this change.

tclemos (changes requested):
Amazing job @kind84, LGTM! 🎉 

As a suggestion, since the log encoding is a string but has restricted values, what do you think about providing them as a const for the internal packages, like the test packages having to set it to "console"?

If you want to be more specific, you can create a type for each, like this:

```go
package log

const (
    LogEncodingJSON = LogEncoding("json")
    LogEncodingConsole = LogEncoding("console")
)

type LogEncoding string

func (le *LogEncoding) String() {
    return string(le)
}
```

Also, looking into the zap repo, looks like they already have methods to help with it, they called Development and Production logs, check this out:

https://github.com/uber-go/zap/blob/a55bdc32f526699c3b4cc51a2cc97e944d02fbbf/config.go#L115-L133

https://github.com/uber-go/zap/blob/a55bdc32f526699c3b4cc51a2cc97e944d02fbbf/config.go#L155-L170

etherman/etherman.go
kind84: Oups well spotted @tclemos. I was testing to log some values from the config. Removed 309a01585b0f4e335731e599ae7dd83905ea4880

kind84:
> Amazing job @kind84, LGTM! tada
> 
> As a suggestion, since the log encoding is a string but has restricted values, what do you think about providing them as a const for the internal packages, like the test packages having to set it to "console"?
> 
> If you want to be more specific, you can create a type for each, like this:
> 
> ```go
> package log
> 
> const (
>     EncodingJSON = LogEncoding("json")
>     EncodingConsole = LogEncoding("console")
> )
> 
> type LogEncoding string
> 
> func (le *LogEncoding) String() {
>     return string(le)
> }
> ```
> 
> Also, looking into the zap repo, looks like they already have methods to help with it, they called Development and Production logs, check this out:
> 
> https://github.com/uber-go/zap/blob/a55bdc32f526699c3b4cc51a2cc97e944d02fbbf/config.go#L115-L133
> 
> https://github.com/uber-go/zap/blob/a55bdc32f526699c3b4cc51a2cc97e944d02fbbf/config.go#L155-L170

Thank you for pointing me to the `zap` functions @tclemos. They look neat and I decided to bring them in our log initialization.
I introduced an `Environment` config value (`production` or `development`) to control which of the two `zap` configuration will be used. WDYT? 879771d2970cb2efe28d12a6e8fd6a8142675d05

tclemos (approved):

PR #1419, opened 2022-11-25
Recursive Aggregator

### LINKED ISSUE #1364: Multiproof Aggregator

kind84:
## The New Recursive Aggregator

Here we introduce the new Recursive Aggregator. It brings a new communication protocol with the Provers and a new algorithm to produce **Recursive Proofs**. The goal is to minimize the number of proofs sent to L1 and therefore save Gas.

### Background

**RECURSIVE PROOFS**

We want to be able to send a proof to L1 only at scheduled intervals and with a small proof.
This can be accomplished thanks to the new Provers which are capable of producing **Recursive Proofs**. These proofs are generated by aggregating **2 subsequent proofs** into a single one.
The process can be repeated to obtain **a single Recursive Proof from a tree of proofs**. Then, when it's time to send to L1, a **final Groth16 proof** is generated from the root of this proofs tree. This is the only proof sent to L1.

recursive-proofs

**AGGREGATOR SERVICE**

This new Aggregator (as opposed to the old one) is a grpc server for the new Provers to connect to.
The communication is performed by establishing a **long-lived connection**, through which different type of messages can be exchanged between Aggregator and Provers.

https://github.com/0xPolygonHermez/zkevm-node/blob/21ae7b5496863d5d7d81dddcb1121f2350522e5f/proto/src/proto/aggregator/v1/aggregator.proto#L17-L47

### Proposal

This PR fully replaces the old `aggregator` package with the new implementation.
The main points of the new code are:

- **The `Start` method** starts listening to connecting Provers.
- **The `Channel` method** is the endpoint to which a Prover connects and starts generating proofs. Here is where all the aggregation logic happens:
    - `tryAggregateProofs`: the Aggregator tries first to feed the Prover with proofs to aggregate.
    - `tryGenerateBatchProof`: if there are no proofs to aggregate, then feed the Prover with a batch to prove.
    -  `tryBuildFinalProof`: when it's time to prepare a final proof, the Aggregator asks the Prover to build it.
- **The `sendFinalProof`** method receives generated final proofs and sends them to L1.

This implementation uses the old Aggregator as starting point. This required to introduce some locking mechanisms to coordinate the Provers. In the near future this will be refactored in favor of a non-blocking version.

Closes #1364.

### Reviewers

Main reviewers:

- @arnaubennassar 
- @ToniRamirezM


aggregator/prover/prover.go
```diff
@@ -2,81 +2,254 @@ package prover
...
 )
 
-// Prover struct
+// Prover abstraction of the grpc prover client.
 type Prover struct {
-	URI    string
-	Client proverclientpb.ZKProverServiceClient
-	Conn   *grpc.ClientConn
+	id                                         string
+	IntervalFrequencyToGetProofGenerationState types.Duration
```
ToniRamirezM, line 17: Is this needed to be exported?

aggregator/prover/prover.go
```diff
@@ -2,81 +2,254 @@ package prover
...
+	}
+	return status.Status == pb.GetStatusResponse_IDLE
 }
 
-func (p *Prover) checkConnection(ctx context.Context, ticker *time.Ticker) {
-	state := p.Conn.GetState()
-	log.Debugf("Checking connection to prover %v. State: %v", p.URI, state)
+// BatchProof instructs the prover to generate a batch proof for the provided
+// input. It returns the ID of the proof being computed.
+func (p *Prover) BatchProof(input *pb.InputProver) (string, error) {
```
ToniRamirezM, line 67: Isn't it better to resturn a *string so it can be nil?

aggregator/prover/prover.go
```diff
@@ -2,81 +2,254 @@ package prover
...
+	res, err := p.call(req)
+	if err != nil {
+		return "", err
+	}
+	if msg, ok := res.Response.(*pb.ProverMessage_GenBatchProofResponse); ok {
+		switch msg.GenBatchProofResponse.Result {
+		case pb.Result_OK:
+			return msg.GenBatchProofResponse.Id, nil
+		case pb.Result_ERROR:
+			return "", errors.New("GenBatchProofResponse.Result: ERROR")
```
ToniRamirezM, line 82: This error could be defined as constants so they could be checked properly from outside.

aggregator/prover/prover.go
```diff
@@ -2,81 +2,254 @@ package prover
...
+	return "", errors.New("GenBatchProofResponse.Result: UNKNOWN")
+}
 
-	if state != connectivity.Ready {
-		log.Infof("Connection to prover %v seems broken. Trying to reconnect...", p.URI)
-		if err := p.Conn.Close(); err != nil {
-			log.Errorf("Could not properly close gRPC connection: %v", err)
+// AggregatedProof instructs the prover to generate an aggregated proof from
+// the two inputs provided. It returns the ID of the proof being computed.
+func (p *Prover) AggregatedProof(inputProof1, inputProof2 string) (string, error) {
```
ToniRamirezM, line 97: Why do you use strings instead of byte arrays? 

Again I think it is better to return *string.

db/migrations/state/0001.sql
```diff
@@ -52,6 +52,18 @@ CREATE TABLE state.forced_batch
     block_num        BIGINT NOT NULL REFERENCES state.block (block_num) ON DELETE CASCADE
 );
 
+CREATE TABLE state.proof
```
ToniRamirezM, line 55: You should create a new migration file, not modify an existing one.

ToniRamirezM (changes requested):

db/migrations/state/0001.sql
kind84: good point @ToniRamirezM, if you agree, I set up a new migration that simply drops and recreates the `state.proof` table with the new fields. This because we probably don't care about existing data on that table if we are migrating to the new aggregator, WDYT? 285299a76d4b8ba14f59cb0eaa92ea764b344b34

aggregator/prover/prover.go
kind84: @ToniRamirezM good point, the errors were still not properly handled in this file.
I tried to handle them better with 59e28c843ca4745b69e6c51869958d30c2b87780 and I added error types as per your suggestion in 5a3fe36d925d00ec4450ed172c97c032a66af703

aggregator/prover/prover.go
kind84: not at all @ToniRamirezM :) , nice catch fbd90aa39452cf8dbe46c1a6549ad632cbdbf0d3

db/migrations/state/0001.sql
ToniRamirezM: I agree

aggregator/aggregator.go
```diff
@@ -46,395 +60,776 @@ func NewAggregator(
...
+	if err != nil {
+		log.Fatalf("failed to listen: %v", err)
+	}
 
-	return a, nil
+	a.srv = grpc.NewServer()
+	pb.RegisterAggregatorServiceServer(a.srv, a)
+
+	healthService := newHealthChecker()
+	grpc_health_v1.RegisterHealthServer(a.srv, healthService)
```
ToniRamirezM, line 105: This naming does not look fine to me. I would remove the v1 from it everywhere.

aggregator/prover/prover.go
kind84: @ToniRamirezM see my answer above https://github.com/0xPolygonHermez/zkevm-node/commit/1252a0e9284de44e8ef5540e071f72dcb9f1544f

aggregator/aggregator.go
kind84: @ToniRamirezM I agree that they choose an ugly package name :smile: 
I added an alias here dbdc312e7eb3f0a7c9a8d0e1088102cbe469fc5f, does it sound better to you?

ethtxmanager/ethtxmanager.go
```diff
@@ -143,14 +145,14 @@ func (c *Client) VerifyBatch(ctx context.Context, batchNum uint64, resGetProof *
...
-			return fmt.Errorf("tx %s failed, err: %w", tx.Hash(), err)
-		} else {
-			log.Infof("batch verification sent to L1 successfully. Tx hash: %s", tx.Hash())
-			time.Sleep(c.cfg.FrequencyForResendingFailedVerifyBatch.Duration)
-			return nil
+			return nil, fmt.Errorf("tx %s failed, err: %w", tx.Hash(), err)
 		}
+
+		log.Infof("batch verification sent to L1 successfully. Tx hash: %s", tx.Hash())
+		time.Sleep(c.cfg.FrequencyForResendingFailedVerifyBatch.Duration)
```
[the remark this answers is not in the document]
kind84: @ToniRamirezM yes you are right, but the change has been recently reverted in the target branch by @ARR552.
We were facing too many conflicts against `develop` and we decided to prioritize finishing the Aggregator.
If you agee we could merge this one and deal with `develop` changes before merging`feature/#1346_update_smc`.

ToniRamirezM (approved):

PR #1421, opened 2022-11-25
Add Grafana docker for visualization

xavier-romero:
### What does this PR do?

Adds Grafana docker to the run command so you can monitor performance and custom metrics.

image

### Reviewers

Main reviewers:

- @kind84 
- @arnaubennassar


test/Makefile
```diff
@@ -1,4 +1,4 @@
-DOCKERCOMPOSE := docker-compose -f docker-compose.yml
+DOCKERCOMPOSE := DOCKERGID=`stat -c '%g' /var/run/docker.sock` docker-compose -f docker-compose.yml
```
tclemos, line 1: is this needed for all the docker-compose commands or only for grafana?
I mean, can we keep this line clean and add this to the beginning of the line 32?

like this
```
DOCKERCOMPOSE := docker-compose -f docker-compose.yml
...
RUNGRAFANA := DOCKERGID=`stat -c '%g' /var/run/docker.sock` $(DOCKERCOMPOSE) up -d $(DOCKERCOMPOSEGRAFANA)
```

tclemos (approved):

test/Makefile
```diff
@@ -230,13 +241,14 @@ stop-approve-matic: ## Stops approve in node container
 
 .PHONY: run
 run: ## Runs a full node
+	$(RUNGRAFANA)
```
kind84, line 244: @xavier-romero in this `run` recipe we don't have the `RUNMETRICS` so we don't launch the Prometheus scraper.
The idea was, since metrics are disabled by default, to launch it separately using `run-metrics` only when they are enabled.
IMHO we could do the same for Grafana (I see you added the `run-grafana`), or maybe have a single recipe to start both Prometheus and Grafana, WDYT?

kind84 (changes requested):

test/Makefile
xavier-romero: @kind84 This setup does not really put the focus on prometheus endpoint, since there is only one available yet, but sets up the components needed to grab it when ready, and also grabs basic docker info like cpu, memory, net, disk, that can be usefull (I guess!) when running tests or metrics. From this point, you can easily add more prometheus endpoints to grab info from when available.
So, where/when does it fits best to set these containers up I don't really know, please feel free to set up where you as a node team feels it best suits. I just pushed the skeleton, I expect you to adjust as per your requeriments.

test/Makefile
xavier-romero: Done, thank you @tclemos

test/Makefile
kind84:
```suggestion
```

@xavier-romero, discussing with @arnaubennassar, we feel it's better to remove this line from `make run` and use `make run-grafana` when explicitly want to enable it.

I remind you that `make run` is the command that we normally use in development and we probably don't need grafana most of the time.

We can discuss about having another Makefile recipe for a "deployment" node including our custom metrics, Prometheus and Grafana.

kind84 (approved):
LGTM :ok_hand:

PR #1422, opened 2022-11-25
improve: making sequencer's batchbuilder to send txs one by one to the executor.

Psykepro:

A part from # (1410).

### What does this PR do?

Changing `sequencer`'s `batchbuilder` to send `txs one by one` to the `executor`.
(Will work on improvement in the next PR to update pending transactions async and to keep the last transaction hash processed in memory to skip it on the next fetch.)

### Reviewers:

- @arnaubennassar 
- @Mikelle 
- @ToniRamirezM 
- @tclemos 
- @KonradIT 

Main reviewers:
- @arnaubennassar 
- @Mikelle 

Codeowner reviewers:

- @arnaubennassar 
- @Mikelle


sequencer/batchbuilder.go
```diff
@@ -17,6 +17,10 @@ import (
 	"github.com/jackc/pgx/v4"
 )
 
+var (
+	getTxsLimit = uint64(1)
```
ToniRamirezM, line 21: Is not this var being used as a constant in the code?

Psykepro: Yep, it its. Will change it to const.

ToniRamirezM (approved):

sequencer/batchbuilder.go
```diff
@@ -73,23 +77,18 @@ func (s *Sequencer) tryToProcessTx(ctx context.Context, ticker *time.Ticker) {
 		return
 	}
 	// process batch
-	log.Infof("processing batch with %d txs. %d txs are new from this iteration", len(s.sequenceInProgress.Txs), appendedTxsAmount)
+	log.Infof("processing tx number [%d] for current batch...", len(s.sequenceInProgress.Txs))
```
kind84, line 80:
```suggestion
	log.Infof("processing [%d] transactions for current batch...", len(s.sequenceInProgress.Txs))
```
@Psykepro WDYT of changing the sentence like suggested? IMO `tx number` sounds like it's processing a single transaction with ID or index `[%d]`.

Same for the other logs.

kind84 (changes requested):

sequencer/batchbuilder.go
```diff
@@ -361,8 +325,11 @@ func (s *Sequencer) processTxs(ctx context.Context) (processTxResponse, error) {
 		log.Errorf("failed to get last batch number, err: %w", err)
 		return processTxResponse{}, err
 	}
+	lenOfTxs := len(s.sequenceInProgress.Txs)
+	isFirstTx := lenOfTxs == 1
+	lastTx := s.sequenceInProgress.Txs[len(s.sequenceInProgress.Txs)-1]
+	s.lastProcessBatchResp, err = s.state.ProcessSingleTransaction(ctx, lastBatchNumber, lastTx, isFirstTx, s.lastProcessBatchResp, dbTx)
```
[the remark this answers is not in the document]
Psykepro: Yep, actually, it is for only 1 will rename it. :)

test/config/test.sequencer.config.toml
```diff
@@ -0,0 +1,122 @@
+IsTrustedSequencer = true
```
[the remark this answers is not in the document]
Psykepro: Committed by mistake.

sequencer/batchbuilder.go
Psykepro: This is misleading because here we are processing a specific transaction for the batch otherwise it will look like we are processing 3 transactions instead of transaction number 3 when we are processing 1 by 1 for example.

kind84: @Psykepro ok then I probably misunderstood the meaning of the log. It really tells about the index of the current processed transaction, right. The index is always the latest transaction in `s.sequenceInProgress.Txs`, that's why you use `len`, isnt' it?

Psykepro: Yep, that's true! :))

PR #1447, opened 2022-12-09
test: adding sequencer benchmark test for - 'measuring the time needed to select 10000 txs from the pool'.

### LINKED ISSUE #1375: Add benchmarks for the sequencer
Add benchmarks test for the sequencer which to do:
  1.   Send all transactions through the JSON RPC
  2.   Wait for all transactions to be added
  3.   Start a timer
  4.   Start the Sequencer
  5.   Finish when all are selected

Deploy on the server using development environment and run it against both sequencer approaches (all txs sent to executor together vs one by one)

Psykepro:

Closes #1375

### What does this PR do?

Adding a `benchmark test` for the `sequencer` which measures the time needed for the sequencer to change the status of all txs (10000) to selected. Also shows the time in the `executor` and the `processing time` in the `sequencer` without the executor.

### Reviewers
@arnaubennassar
@tclemos  
@ToniRamirezM 
@cool-develope 

Main reviewers:
@arnaubennassar
@tclemos


arnaubennassar (approved):
Nice job :tada: 

Ideas for future improvements on the benchmarks:

- Have the abilityu to reuse this logic with different "vectors", so we can have different combos of txs being added into the pool, but the benchmarking logic is the same, just changing how the txs are being generated.

Psykepro:
> Nice job tada
> 
> Ideas for future improvements on the benchmarks:
> 
> * Have the abilityu to reuse this logic with different "vectors", so we can have different combos of txs being added into the pool, but the benchmarking logic is the same, just changing how the txs are being generated.

We can create different independent tests for each different type of transaction to be isolated testing. And in that case all other logic from this PR is reusable. :))

ToniRamirezM (approved):

Issue #1478, opened 2022-12-22
Implementing Finalizer from new Sequencer Implementation


Psykepro:
Implementing the Finalizer from new Sequencer specification docs:
https://github.com/0xPolygonHermez/zkevm-node/blob/feature/sequencer-refactor-spec/docs/sequecer-mango.drawio.png
https://github.com/0xPolygonHermez/zkevm-node/blob/feature/sequencer-refactor-spec/docs/simplesequencer.go#L194

PR #1481, opened 2022-12-22
wip: adding some implementation of the finalizer, defining needed int…

Psykepro:
…erfaces by the finalizer from the dbmanager, and implementing the remaining resourses Sub.

Related to issue 0xPolygonHermez/zkevm-node#1478.

### What does this PR do?

Adding some of the implementations (still WIP). Mostly to define needed interfaces by the dbManager to unblock @ToniRamirezM. I've added to the proto the TouchedAddresses and generated the proto files. 
I got carried away and implemented the `remainingResources.Sub `method but I think the implementation is very clean. :)

### Reviewers
- @ToniRamirezM 
- @arnaubennassar 
- @agnusmor 

Main reviewers:
- @ToniRamirezM 
- @arnaubennassar
- @agnusmor


ToniRamirezM (approved):

pool/transaction.go
```diff
@@ -73,6 +73,18 @@ func (zkc *ZkCounters) SumUpZkCounters(txZkCounters ZkCounters) {
 	zkc.UsedSteps += txZkCounters.UsedSteps
 }
 
+// SubZkCounters subtract zk counters with passed zk counters (not safe)
+func (zkc *ZkCounters) SubZkCounters(txZkCounters ZkCounters) {
```
arnaubennassar, line 77: I'm not very happy about this (because for how it used to be, not your changes). Things that IMO could imporve current situation:

- Move the struct to `state` package
- Ideally reuse the counters from the proto interface so we don't need to reparse (maybe by wrapping)
- `ZkCounters` => `ZKCounters` :pray: 
- Include the underflow logic already in the `SubZkCounters` method
- `SubZkCounters` => `Sub`

arnaubennassar: The only actual request here is the renamings of the functions, otherwise it's just suggestions up to you to implement if you consider them legit

sequencer/finalizer.go
```diff
@@ -13,38 +19,59 @@ type finalizer struct {
...
+	if err != nil {
+		log.Fatal("failed to get last batch")
+	}
 
-	wipBatch := wipBatch{}
+	wipBatch := wipBatch{
+		GER:                   lastBatch.GlobalExitRoot,
+		initialStateRoot:      lastBatch.StateRoot,
+		intermediaryStateRoot: lastBatch.StateRoot,
+		txs:                   make([]TxTracker, 0, f.maxTxsPerBatch),
```
arnaubennassar, line 63: Exisitng transactions in the batch should be also loaded from DB. You can do this in a later PR as here the priority is to define interfaces ASAP

sequencer/finalizer.go
```diff
@@ -13,38 +19,59 @@ type finalizer struct {
...
+func (f *finalizer) Start(ctx context.Context) {
+	lastBatch, err := f.dbManager.GetLastBatch(ctx)
+	if err != nil {
+		log.Fatal("failed to get last batch")
+	}
 
-	wipBatch := wipBatch{}
+	wipBatch := wipBatch{
+		GER:                   lastBatch.GlobalExitRoot,
+		initialStateRoot:      lastBatch.StateRoot,
```
arnaubennassar, line 61: The initial state root is actually yhe state root stored in the previous batch (the last closed batch). You can do this in a later PR as here the priority is to define interfaces ASAP

sequencer/finalizer.go
```diff
@@ -13,38 +19,59 @@ type finalizer struct {
...
+		intermediaryStateRoot: lastBatch.StateRoot,
```
arnaubennassar, line 62: This state root should be taken from the latest tx included into the batch. You can do this in a later PR as here the priority is to define interfaces ASAP

sequencer/finalizer.go
```diff
@@ -81,52 +108,105 @@ func (f *finalizer) Start() {
 		for {
 			tx := f.worker.GetBestFittingTx(wipBatch.remainingResources)
 			if tx != nil {
+				lenOfTxs := len(wipBatch.txs)
+				isFirstTx := lenOfTxs == 1
```
arnaubennassar, line 112:
```suggestion
				isFirstTx := lenOfTxs == 0
```

At this point, tx is not added yet into `wipBatch.txs`. If `wipBatch.txs == 1` it means that in previous iterations there has been a tx added into the batch, and that tx was actually the first one.

Beyond that we may consider to execute a batch with no txs at all to produce the GER update, and here use always empty hash. We can talk about this later. You can do this in a later PR as here the priority is to define interfaces ASAP

sequencer/finalizer.go
```diff
@@ -81,52 +108,105 @@ func (f *finalizer) Start() {
...
-				// // f.worker.UpdateAfterSingleSuccessfulTxExecution()
+				if result.Error != nil {
+					// // decide if we [MoveTxToNotReady, DeleteTx, UpdateTx]
+
+				} else {
+					processedTx := result.Responses[0]
+					from, err := types.Sender(types.NewEIP155Signer(processedTx.Tx.ChainId()), &processedTx.Tx)
+					if err != nil {
+						from, err = types.Sender(types.HomesteadSigner{}, &processedTx.Tx)
+					}
```
arnaubennassar, line 135: We should be able to get the `from` from the `TxTracker`. You can do this in a later PR as here the priority is to define interfaces ASAP

sequencer/finalizer.go
```diff
@@ -81,52 +108,105 @@ func (f *finalizer) Start() {
...
+						continue
+					}
+
+					// We have a successful processing if we are here
+					wipBatch.intermediaryStateRoot = result.NewStateRoot
+					processBatchRequest.StateRoot = result.NewStateRoot
+					processBatchRequest.OldAccInputHash = result.NewAccInputHash
+					// We store the processed transaction, add it to the batch and delete from the pool atomically.
+					go f.dbManager.StoreProcessedTxAndDeleteFromPool(ctx, wipBatch.batchNumber, processedTx)
+					go f.worker.UpdateAfterSingleSuccessfulTxExecution(from, result.TouchedAddresses)
```
arnaubennassar, line 154:
```suggestion
					f.worker.UpdateAfterSingleSuccessfulTxExecution(from, result.TouchedAddresses)
```

We decided to do this synchronously. The idea is to compromise a bit of performance but to gain more security (improve "hit ratio")

sequencer/finalizer.go
```diff
@@ -81,52 +108,105 @@ func (f *finalizer) Start() {
...
+							log.Fatalf("failed to decode tx")
+						}
+						ethTxs = append(ethTxs, *tx)
+					}
+					receipt := state.ProcessingReceipt{
+						BatchNumber:   wipBatch.batchNumber,
+						AccInputHash:  processBatchRequest.OldAccInputHash,
+						StateRoot:     wipBatch.intermediaryStateRoot,
+						LocalExitRoot: processBatchRequest.GlobalExitRoot,
+						Txs:           ethTxs,
```
arnaubennassar, line 173: I would like to review this to see if we can directly pass `wipBatch.txs` here. Also it would be nice if we can stop working with `types.Transaction` in the first place and use the `TxTracker` for the sequencer and "raw txs" (just the encoded tx as `[]byte`) in the state.

I'm not sure about how feasible is this, but at least is worth exploring

state/types.go
```diff
@@ -1,28 +1,45 @@
...
+// ProcessBatchRequest represents the request of a batch process.
+type ProcessBatchRequest struct {
+	BatchNumber      uint64
+	StateRoot        common.Hash
+	GlobalExitRoot   common.Hash
+	OldAccInputHash  common.Hash
+	TxData           []byte
+	SequencerAddress common.Address
+	Timestamp        uint64
+	IsFirstTx        bool
```
arnaubennassar, line 28: I think this is not needed here. This should be used for the GER empty hack, but we're already taking into consideration this in the finalizer.

With that being said, since this is something related to how the executor works, it will make a lot of sense to take care bout this using the `isFirstTx`, but then we don't need to do it in the finalizer. Also if it's the state who needs to do that, then it also needs to know if the inteeractions is meant to be done in the `tx by tx` mode or in the `full batch mode`...

Up to you, essentially you have to either:

- Remove `IsFirstTx` from the struct
- Remoe the GER = empty hash from finalizer, and add a `tx by tx` flag in the process request struct

arnaubennassar (changes requested):
- Sorry for too many comments saying `You can do this in a later PR as here the priority is to define interfaces ASAP`. It was to help identifying what needs to be done now and what can be done later... Was lazy so just copy pasted it :upside_down_face: 
- Also not to be done now, but in some parts of the code there already too many identation levels (at least I've counted 5...). As a rule of thumb when this happens it's time to start thinking about splitting into multiple functions

sequencer/finalizer.go
Psykepro: Yes, this is haven't done it there are several things I need to add just wanted to add some logic but to focus on the interfaces there are a lot of things to be finished :)) I was about to fix all of the things here it's in WIP.

sequencer/worker.go
```diff
@@ -103,9 +108,55 @@ type RemainingResources struct {
...
-func (r *RemainingResources) sub(tx TxTracker) error {
-	// Substract resources
-	// error if underflow (restore in this case)
+func (r *RemainingResources) Sub(other RemainingResources) error {
+	err := r.checkForResourcesOverflow(other)
+	if err != nil {
+		return err
+	}
+	r.remainingBytes -= other.remainingBytes
+	r.remainingGas -= other.remainingGas
```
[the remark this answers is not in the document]
Psykepro: It has it in line 112 this check everything

sequencer/worker.go
```diff
@@ -103,9 +108,55 @@ type RemainingResources struct {
...
+	r.remainingZKCounters.SubZkCounters(other.remainingZKCounters)
+
+	return nil
+}
+
+func (r *RemainingResources) checkForResourcesOverflow(other RemainingResources) error {
```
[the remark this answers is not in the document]
Psykepro: Ops, yep mistake here.

sequencer/finalizer.go
Psykepro: This I had in mind just wanted to make quick PR for the interfaces.

sequencer/finalizer.go
Psykepro: This I had in mind just wanted to make quick PR for the interfaces.

sequencer/finalizer.go
Psykepro: This I had in mind just wanted to make quick PR for the interfaces.

sequencer/finalizer.go
Psykepro: Ops, yep mistake here..  from hurrying :D

sequencer/finalizer.go
Psykepro: So this I guess to be provided by the DbManager to the struct form the database?

sequencer/finalizer.go
Psykepro: Yep, sorry mistake here.

sequencer/finalizer.go
Psykepro: Okay, will check if possible to use everywhere the raw tx.

state/types.go
Psykepro: Okay

sequencer/interfaces.go
```diff
@@ -43,11 +47,17 @@ type txManager interface {
 
 type workerInterface interface {
```
agnusmor, line 48: We need to add also "func (w *Worker) AddTx(tx TxTracker)" function that is used by the Pool Loader to load txs from the Pool to the Worker

Psykepro: This PR is just for the interfaces needed in the `dbManager` by the `Finalizer`. But after the vacations, as you will be still on vacation I can take a look at adding some of the needed in the `Worker` no probs! :))

sequencer/interfaces.go
```diff
@@ -29,25 +29,42 @@ type etherman interface {
...
+// The dbManager will need to handle the errors inside the functions which don't return error as they will be used async in the other abstractions.
+// Also if dbTx is missing this needs also to be handled ion the dbManager
 type dbManagerInterface interface {
-}
-
-type executorInterface interface {
-	ExecuteTransaction(processBatchRequest state.ProcessBatchRequest) state.ProcessBatchResponse
+	BeginStateTransaction(ctx context.Context) (pgx.Tx, error)
+	CreateFirstBatch(ctx context.Context, sequencerAddress common.Address) state.ProcessingContext
+	GetLastBatchNumber(ctx context.Context) (uint64, error)
```
ToniRamirezM, line 59:
```suggestion
	GetLastBatchNumber(ctx context.Context, dbTx pgx.Tx) (uint64, error)
```

sequencer/interfaces.go
```diff
@@ -29,25 +29,42 @@ type etherman interface {
...
+	StoreProcessedTransaction(ctx context.Context, dbTx pgx.Tx, batchNumber uint64, processedTx *state.ProcessTransactionResponse) error
+	DeleteTxFromPool(ctx context.Context, dbTx pgx.Tx, txHash common.Hash) error
+	StoreProcessedTxAndDeleteFromPool(ctx context.Context, batchNumber uint64, response *state.ProcessTransactionResponse)
```
ToniRamirezM, line 62:
```suggestion
	StoreProcessedTxAndDeleteFromPool(ctx context.Context, batchNumber uint64, response *state.ProcessTransactionResponse, dbTx pgx.Tx)
```

sequencer/interfaces.go
```diff
@@ -29,25 +29,42 @@ type etherman interface {
...
+	CloseBatch(ctx context.Context, params ClosingBatchParameters)
```
ToniRamirezM, line 63:
```suggestion
	CloseBatch(ctx context.Context, params ClosingBatchParameters, dbTx pgx.Tx)
```

sequencer/interfaces.go
```diff
@@ -29,25 +29,42 @@ type etherman interface {
...
+	GetWIPBatch(ctx context.Context) (wipBatch, error)
```
ToniRamirezM, line 64:
```suggestion
	GetWIPBatch(ctx context.Context, dbTx pgx.Tx) (wipBatch, error)
```

sequencer/interfaces.go
```diff
@@ -29,25 +29,42 @@ type etherman interface {
...
+	GetLastBatch(ctx context.Context) (state.Batch, error)
```
ToniRamirezM, line 65:
```suggestion
	GetLastBatch(ctx context.Context, dbTx pgx.Tx) (state.Batch, error)
```

sequencer/interfaces.go
```diff
@@ -29,25 +29,42 @@ type etherman interface {
...
+	GetLastNBatches(ctx context.Context, numBatches uint) ([]*state.Batch, error)
```
ToniRamirezM, line 66:
```suggestion
	GetLastNBatches(ctx context.Context, numBatches uint, dbTx pgx.Tx) ([]*state.Batch, error)
```

sequencer/interfaces.go
```diff
@@ -29,25 +29,42 @@ type etherman interface {
...
+	GetLastClosedBatch(ctx context.Context) (state.Batch, error)
```
ToniRamirezM, line 67:
```suggestion
	GetLastClosedBatch(ctx context.Context, dbTx pgx.Tx) (state.Batch, error)
```

sequencer/interfaces.go
```diff
@@ -29,25 +29,42 @@ type etherman interface {
...
+	IsBatchClosed(ctx context.Context, batchNum uint64) (bool, error)
```
ToniRamirezM, line 68:
```suggestion
	IsBatchClosed(ctx context.Context, batchNum uint64, dbTx pgx.Tx) (bool, error)
```

sequencer/interfaces.go
```diff
@@ -29,25 +29,42 @@ type etherman interface {
...
+	MarkReorgedTxsAsPending(ctx context.Context) error
```
ToniRamirezM, line 69:
```suggestion
	MarkReorgedTxsAsPending(ctx context.Context, dbTx pgx.Tx) error
```

sequencer/interfaces.go
```diff
@@ -29,25 +29,42 @@ type etherman interface {
...
 
 type workerInterface interface {
-	GetBestFittingTx(resources RemainingResources) *TxTracker
+	GetBestFittingTx(resources BatchResources) *TxTracker
+	UpdateAfterSingleSuccessfulTxExecution(from common.Address, touchedAddresses map[common.Address]*state.TouchedAddress)
+	UpdateTx(txHash common.Hash, from common.Address, ZKCounters state.ZKCounters)
 }
 
+// The dbManager will need to handle the errors inside the functions which don't return error as they will be used async in the other abstractions.
+// Also if dbTx is missing this needs also to be handled ion the dbManager
```
ToniRamirezM, line 55:
```suggestion
// Also if dbTx is missing this needs also to be handled in the dbManager
```

What do you mean? I think all method should have dbTx in their signature

ToniRamirezM (changes requested):

sequencer/interfaces.go
```diff
@@ -29,25 +29,42 @@ type etherman interface {
...
 type dbManagerInterface interface {
-}
-
-type executorInterface interface {
-	ExecuteTransaction(processBatchRequest state.ProcessBatchRequest) state.ProcessBatchResponse
+	BeginStateTransaction(ctx context.Context) (pgx.Tx, error)
+	CreateFirstBatch(ctx context.Context, sequencerAddress common.Address) state.ProcessingContext
```
ToniRamirezM, line 58:
```suggestion
	CreateFirstBatch(ctx context.Context, sequencerAddress common.Address, dbTx pgx.Tx) state.ProcessingContext
```

ToniRamirezM (changes requested):

sequencer/dbmanager.go
```diff
@@ -15,6 +26,126 @@ func (d *dbManager) Start() {
...
+}
+
+func (d *dbManager) GetLastNBatches(ctx context.Context, numBatches uint) ([]*state.Batch, error) {
+	// TODO: Returns last N batches
+	return []*state.Batch{}, errors.New("")
+
+}
+
+// ClosingBatchParameters contains the necessary parameters to close a batch
+type ClosingBatchParameters struct {
```
ToniRamirezM, line 136: I am moving this into the state package

ToniRamirezM (approved):

PR #1484, opened 2022-12-22
new eth tx manager

### LINKED ISSUE #1464: eth tx manager refactoring
review and do a refactoring to make sure the current eth tx manager implementation is fault tolerant with L1 tx and make sure they get mined without duplicating data.

The current implementation seems to be sending repeated sequences due to timeout reached when waiting for the tx to get mined.

tclemos:
Closes #1464.

### What does this PR do?

make eth tx manager generic for any tx

### Reviewers

Main reviewers:

@ToniRamirezM 
@arnaubennassar 
@ARR552


test/scripts/batchsender/main.go
```diff
@@ -172,7 +172,7 @@ func sendBatches(cliCtx *cli.Context) error {
 		}
 
 		// send to L1
-		tx, err := ethMan.SequenceBatches(ctx, seqs, 0, nil, nil)
+		tx, err := ethMan.SequenceBatches(ctx, seqs, 0, nil, nil, false)
```
tclemos, line 175: I need to review this

Dockerfile
```diff
@@ -14,6 +14,6 @@ RUN cd /src && make build
 # CONTAINER FOR RUNNING BINARY
 FROM alpine:3.16.0
 COPY --from=build /src/dist/zkevm-node /app/zkevm-node
-COPY --from=build /src/config/environments/local/local.node.config.toml /app/example.config.toml
+COPY --from=build /src/config/environments/public/public.node.config.toml /app/example.config.toml
```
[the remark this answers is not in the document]
tclemos: What problem?

cmd/approve.go
```diff
@@ -11,12 +11,16 @@ import (
...
+	amountArg := ctx.String(config.FlagAmount)
+	amount, _ := new(big.Float).SetString(amountArg)
 	if amount == nil {
 		fmt.Println("Please, introduce a valid amount. Use '.' instead of ',' if it is a decimal number")
 		return nil
 	}
+
+	addrKeyStorePath := ctx.String(config.FlagKeyStorePath)
+	addrPassword := ctx.String(config.FlagPassword)
+
```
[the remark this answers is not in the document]
tclemos: The credentials were moved from [Etherman] to other components and there is no [Approve] section in the config file, that's why they are parameters.

cmd/run.go
```diff
@@ -79,18 +75,27 @@ func start(cliCtx *cli.Context) error {
...
-	ethTxManager := ethtxmanager.New(c.EthTxManager, etherman, st)
+	ethTxManagerStorage, err := ethtxmanager.NewPostgresStorage(c.StateDB)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	etmStorage, err := ethtxmanager.NewPostgresStorage(c.StateDB)
+	if err != nil {
+		log.Fatal(err)
+	}
```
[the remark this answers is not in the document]
tclemos: oops, copy/paste problems, fixing

etherman/etherman.go
```diff
@@ -292,44 +299,54 @@ func (etherMan *Client) updateGlobalExitRootEvent(ctx context.Context, vLog type
...
+func (etherMan *Client) BuildSequenceBatchesTxData(sender common.Address, sequences []ethmanTypes.Sequence) (to *common.Address, value *big.Int, data []byte, err error) {
+	opts, err := etherMan.getAuthByAddress(sender)
+	if err == ErrNotFound {
+		return nil, nil, nil, fmt.Errorf("failed to build sequence batches, err: %w", ErrPrivateKeyNotFound)
 	}
-	if nonce != nil {
-		sendSequencesOpts.Nonce = nonce
+	opts.NoSend = true
+	// force nonce, gas limit and gas price to avoid querying it from the chain
+	opts.Nonce = big.NewInt(1)
```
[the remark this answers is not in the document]
tclemos: This method doesn't send any tx to the chain or process this in the EVM, we just use the methods in order to build the proper Txs fields taking advantage of the code generated from the SC. 

Since we set the `NoSend` field we are safe to not get the tx sent to the chain, but we must set the fields `Nonce`, `GasLimit`, and `GasPrice` to a fixed value in order to avoid accessing the blockchain to try to estimate the gas or waste time and resources.

etherman/etherman.go
```diff
@@ -861,6 +849,160 @@ func (etherMan *Client) getGasPrice(ctx context.Context) *big.Int {
...
+	if !found {
+		return bind.TransactOpts{}, ErrNotFound
+	}
+	return auth, nil
+}
+
+// generateRandomAuth generates an authorization instance from a
+// randomly generated private key to be used to estimate gas for PoE
+// operations NOT restricted to the Trusted Sequencer
+func (etherMan *Client) generateRandomAuth() (bind.TransactOpts, error) {
```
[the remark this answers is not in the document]
tclemos: Well, in this new version of the `EthTxManager`, the `Sequencer` and `Aggregator` shouldn't know about the account secrets and only about the public address, the `EthTxManager` is the holder of the secrets and is the only one able to sign the Txs with it. Also, all the `Transaction` logic happens inside of the `EthTxManager` and they are not triggered by the `Sequencer` or `Aggregator` using `Etherman`.

With that said, `Etherman` should be used by the `Sequencer` and `Aggregator` in read-only mode and shouldn't require an authorization object, BUT the generated code to interact with the SC doesn't help much in terms of creating a Tx without signing it, because the methods are meant to send the TX to the blockchain, so in order to use these methods we must provide an authorization object.

That's why we have this method.

etherman/etherman.go
```diff
@@ -292,44 +299,54 @@ func (etherMan *Client) updateGlobalExitRootEvent(ctx context.Context, vLog type
...
-		return nil, ErrIsReadOnlyMode
-	}
-	sendSequencesOpts := *etherMan.auth
-	sendSequencesOpts.GasLimit = gasLimit
-	if gasPrice != nil {
-		sendSequencesOpts.GasPrice = gasPrice
-	} else if etherMan.GasProviders.MultiGasProvider {
-		sendSequencesOpts.GasPrice = etherMan.getGasPrice(ctx)
+// BuildSequenceBatchesTxData builds a []bytes to be sent to the PoE SC method SequenceBatches.
+func (etherMan *Client) BuildSequenceBatchesTxData(sender common.Address, sequences []ethmanTypes.Sequence) (to *common.Address, value *big.Int, data []byte, err error) {
```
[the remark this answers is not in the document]
ARR552: As we are not going to need this value I think it is better to not have it

Dockerfile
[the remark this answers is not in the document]
ARR552: Never mind. I think it is ok as this is just an example config file.

aggregator/aggregator.go
```diff
@@ -496,12 +507,12 @@ func (a *Aggregator) getAndLockProofsToAggregate(ctx context.Context, prover pro
 
 	if err != nil {
 		dbTx.Rollback(ctx) //nolint:errcheck
```
ToniRamirezM, line 509: Maybe is good to log the rollback error. I would denote a DB access error.

aggregator/aggregator.go
```diff
@@ -460,12 +471,12 @@ func (a *Aggregator) unlockProofsToAggregate(ctx context.Context, proof1 *state.
...
```
ToniRamirezM, line 473: Maybe is good to log the rollback error. I would denote a DB access error.

db/migrations/state/0005.sql
```diff
@@ -0,0 +1,22 @@
...
+    to_addr    VARCHAR,
+    nonce      DECIMAL(78, 0) NOT NULL,
+    value      DECIMAL(78, 0),
+    data       VARCHAR,
+    gas        DECIMAL(78, 0) NOT NULL,
+    gas_price  DECIMAL(78, 0) NOT NULL,
+    status     VARCHAR NOT NULL,
+    history    VARCHAR[],
+    block_num  BIGINT,
+    created_at TIMESTAMP NOT NULL,
```
ToniRamirezM, line 19: In other tables we use TIMESTAMP WITH TIME ZONE. Maybe we should follow the same approach everywhere.

ethtxmanager/ethtxmanager.go
```diff
@@ -11,156 +11,612 @@ import (
...
+	mTxs, err := c.storage.GetByStatus(ctx, nil, statusesFilter, nil)
+	if err != nil {
+		return fmt.Errorf("failed to get created monitored txs: %v", err)
+	}
+
+	log.Infof("found %v monitored tx to process", len(mTxs))
+
+	for _, mTx := range mTxs {
+		mTx := mTx // force variable shadowing to avoid pointer conflicts
+		mTxLog := log.WithFields("monitored tx", mTx.id)
```
kind84, line 249: @tclemos since logs can now be formatted in JSON, I would suggest camel-case for log fields.

ethtxmanager/ethtxmanager.go
```diff
@@ -11,156 +11,612 @@ import (
...
+// for the confirmed and failed ones, the resultHandler will be triggered
+func (c *Client) ProcessPendingMonitoredTxs(ctx context.Context, owner string, resultHandler ResultHandler, dbTx pgx.Tx) {
+	statusesFilter := []MonitoredTxStatus{
+		MonitoredTxStatusCreated,
+		MonitoredTxStatusSent,
+		MonitoredTxStatusFailed,
+		MonitoredTxStatusConfirmed,
+		MonitoredTxStatusReorged,
+	}
+	// keep running until there ware pending monitored txs
```
kind84, line 568:
```suggestion
	// keep running until there are pending monitored txs
```

kind84 (approved):
LGTM, impressive job @tclemos :clap:

ARR552 (approved):

db/migrations/state/0005.sql
```diff
@@ -0,0 +1,22 @@
+-- +migrate Down
+DROP TABLE IF EXISTS state.monitored_txs;
```
ToniRamirezM, line 2: Maybe eth tx manager should have its own schema / database

ToniRamirezM (approved):

db/migrations/state/0005.sql
tclemos: we can't because we depend on reorgs and the database tx needs to be handled together to avoid out-of-sync data.

PR #1507, opened 2023-01-04
[Aggregator] Switch to new ethtxmanager interface

### LINKED ISSUE #1464: eth tx manager refactoring

kind84:
Closes #1464 (needs #1484).

### What does this PR do?

This PR adapts the Aggregator to use the new `ethtxmanager` interface.
A UT for the `sendFinalProof` method has been put in place to assert the correct behavior.

### Reviewers

Main reviewers:

@arnaubennassar 
@tclemos 
@agnusmor


aggregator/aggregator.go
```diff
@@ -220,18 +224,64 @@ func (a *Aggregator) sendFinalProof() {
...
-				err := a.State.UpdateGeneratedProof(ctx, proof, nil)
+			err = a.EthTxManager.Add(ctx, txManagerOwner, txID, pubAddr, tx.To(), tx.Value(), tx.Data(), nil)
+			if err != nil {
+				log.Errorf("Failed to add final proof transaction to the tx manager, %v", err)
+			}
+
+			go a.EthTxManager.ManageTxs()
+
+			// wait for the tx to be confirmed
+		txResultLoop:
```
agnusmor, line 245: I think is better to put this "txResultLoop" code in a separate function

PR #1521, opened 2023-01-09
Fix/rc4

arnaubennassar:


aggregator/aggregator.go
```diff
@@ -214,7 +214,6 @@ func (a *Aggregator) sendFinalProof() {
 			finalBatch, err := a.State.GetBatchByNumber(ctx, proof.BatchNumberFinal, nil)
 			if err != nil {
 				log.Errorf("Failed to retrieve batch with number [%d]", proof.BatchNumberFinal)
-				a.enableProofVerification()
```
kind84, line 217: @arnaubennassar if we want to revert all the changes introduced in `v0.0.1-rc5` we need to apply all the changes as per this commit (which reverts `rc5` changes in `release/v0.0.1` branch) https://github.com/0xPolygonHermez/zkevm-node/commit/434951a1e0c5dc85e06a612d0559e87fea98672b
Here most of them are missing.

ARR552 (approved):

aggregator/aggregator.go
arnaubennassar: maybe the release branch was not up to date when we did the fix branch

PR #1523, opened 2023-01-10
Feature/#1506 forced batch

### LINKED ISSUE #1506: Review forceBatch

ARR552:
Closes #1506 

### What does this PR do?

Adds unit test in the synchronicer and applies some fixes

### Reviewers

- @cool-develope 
- @tclemos 
- @arnaubennassar


state/pgstatestorage.go
```diff
@@ -112,21 +109,34 @@ func (p *PostgresStorage) getExecQuerier(dbTx pgx.Tx) execQuerier {
 // Reset resets the state to a block for the given DB tx
 func (p *PostgresStorage) Reset(ctx context.Context, blockNumber uint64, dbTx pgx.Tx) error {
 	e := p.getExecQuerier(dbTx)
+	const resetForcedBatchSQL = `
+		UPDATE state.forced_batch 
+		SET batch_num = null
+		FROM state.virtual_batch 
+		INNER JOIN state.block ON state.block.block_num = state.virtual_batch.block_num
```
cool-develope, line 116: why we need the `inner join` with block table? `virtual_batch` has already `block_num` ???

state/pgstatestorage.go
```diff
@@ -112,21 +109,34 @@ func (p *PostgresStorage) getExecQuerier(dbTx pgx.Tx) execQuerier {
 // Reset resets the state to a block for the given DB tx
 func (p *PostgresStorage) Reset(ctx context.Context, blockNumber uint64, dbTx pgx.Tx) error {
```
cool-develope, line 110: what is the plan for the test of the `Reset` function?

sequencer/broadcast/server.go
```diff
@@ -119,7 +115,7 @@ func (s *Server) genericGetBatch(ctx context.Context, batch *state.Batch) (*pb.G
 		RollupExitRoot:    rollupExitRoot,
 		Timestamp:         uint64(batch.Timestamp.Unix()),
 		Transactions:      transactions,
-		ForcedBatchNumber: forcedBatchNum,
+		ForcedBatchNumber: fb,
```
arnaubennassar, line 118: maybe this should b a pointer, as I guess that the first `ForcedBatchNumber` will be 0

arnaubennassar (changes requested):
If I understood the code well, if there is an inconsistency on the DB related to the forced batches, the synchronizer will return error, and potentially enter and endless loop of errors.

I think this is good enough for now, but eventually the behaviour should be more like the trusted state: if the content of the DB is unexpected, we log an error, but we put whatever it's on the virtual state into the node's DB

sequencer/broadcast/server.go
ARR552: No, In this case the first forcedBatch will be one

ARR552:
> If I understood the code well, if there is an inconsistency on the DB related to the forced batches, the synchronizer will return error, and potentially enter and endless loop of errors.
> 
> I think this is good enough for now, but eventually the behaviour should be more like the trusted state: if the content of the DB is unexpected, we log an error, but we put whatever it's on the virtual state into the node's DB

Yes, I think that for initial steps of this feature testing it is good to detect the problem as soon as possible. We can modify it anytime.

ToniRamirezM (approved):

arnaubennassar (approved):

PR #1524, opened 2023-01-10
Feature/dbmanager

ToniRamirezM:
DBManager WIP

Main reviewers:

- @arnaubennassar 
- @Psykepro 
- @agnusmor


sequencer/dbmanager.go
```diff
@@ -178,18 +195,50 @@ func (d *dbManager) GetWIPBatch(ctx context.Context) (*WipBatch, error) {
 	}
 
 	wipBatch := &WipBatch{
-		batchNumber:  lastBatch.BatchNumber,
-		coinbase:     lastBatch.Coinbase,
-		accInputHash: lastBatch.AccInputHash,
-		// initialStateRoot: lastBatch.StateRoot,
+		batchNumber:    lastBatch.BatchNumber,
+		coinbase:       lastBatch.Coinbase,
+		accInputHash:   lastBatch.AccInputHash,
```
Psykepro, line 200: We need to get also `initialAccInputHash` by checking if there was a previous closed batch to get I t from it otherwise we use the `lastBatch.AccInputHash`

sequencer/dbmanager.go
```diff
@@ -178,18 +195,50 @@ func (d *dbManager) GetWIPBatch(ctx context.Context) (*WipBatch, error) {
...
 		stateRoot:      lastBatch.StateRoot,
```
Psykepro, line 201: We need to get also `initialStateRoot` by checking if there was a previously closed batch to get I t from it otherwise we use the `lastBatch.StateRoot`

sequencer/dbmanager.go
```diff
@@ -178,18 +195,50 @@ func (d *dbManager) GetWIPBatch(ctx context.Context) (*WipBatch, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
+	txTrackerArray := make([]TxTracker, len(txs))
 
-		// TODO: txs
-		// TODO: remainingResources
+	// TODO: Init counters and totals to MAX values
+	var totalBytes uint64
```
Psykepro, line 215: You need to initialize the max allowed values for `totalBytes` and `batchZkCounters` I've created such function in the finalizer [getMaxRemainingResources](https://github.com/0xPolygonHermez/zkevm-node/blob/feature/newSequencer/sequencer/finalizer.go#L479) maybe will be better to move it to the sequencer as shared function we can reuse here as well.

sequencer/dbmanager.go
Psykepro, line 200: You need to init here also the `initialAccInputHash` and the `InitialStateRoot` if there is a previously closed batch. (turned out we need those fields for L2Reorgs.

sequencer/interfaces.go
```diff
@@ -96,4 +99,6 @@ type dbManagerStateInterface interface {
 	GetLastL2Block(ctx context.Context, dbTx pgx.Tx) (*types.Block, error)
 	MarkReorgedTxsAsPending(ctx context.Context) error
 	GetLatestGer(ctx context.Context) (state.GlobalExitRoot, time.Time, error)
+	GetLatestGlobalExitRoot(ctx context.Context, maxBlockNumber uint64, dbTx pgx.Tx) (state.GlobalExitRoot, time.Time, error)
+	GetSender(tx types.Transaction) (common.Address, error)
```
Psykepro, line 103: this `GetSender` function is not on the state struct but is just defined as a standalone function: https://github.com/0xPolygonHermez/zkevm-node/blob/develop/state/transaction.go#L9

sequencer/interfaces.go
```diff
@@ -96,4 +99,6 @@ type dbManagerStateInterface interface {
 	GetLastL2Block(ctx context.Context, dbTx pgx.Tx) (*types.Block, error)
 	MarkReorgedTxsAsPending(ctx context.Context) error
 	GetLatestGer(ctx context.Context) (state.GlobalExitRoot, time.Time, error)
```
Psykepro, line 101: The state interface don't have `GetLatestGer` method

Psykepro (changes requested):

PR #1548, opened 2023-01-17
Feature/#1486 l2gasprice

### LINKED ISSUE #1486: Set L2 gas price sugestion dynamically

ARR552:
Closes #1486 

### What does this PR do?

This PR creates the gaspricer as a component that runs independent. Also adds the follower method to calculate l2 gasprice. This methods calculates the l2 gas price based on the l1 gas price applying a correction factor.
Some Unit tests added.

### Reviewers

- @arnaubennassar 
- @ToniRamirezM 
- @Psykepro


arnaubennassar (changes requested):
Not fully reviewed, overall, please review naming to not create a confusion with estimate gas for L1 / L2, gas price fro L1, ... I'd try to use something like "L2 gas price suggestor"

tclemos (commented):
The code in general is ok to me, but I don't feel comfortable with the names used, `Suggestor` has a typo, the correct one is `Suggester`, but, I still think we shouldn't use `Suggester` in the names, because they are setting the price directly to the DB, not suggesting it to be set.

Default, LastNBatches and Follower doesn't let it clear they are the pricers, a follower is a follower... following that?

I know we use the word `interface` for some interfaces, but I strongly recomend avoiding it when possible.

Here are my suggestions:
- get rid of the `Suggestor` suffix
- get rid of the L1 and L2 prefixes
  - Etherman is meant to work with L1
  - GasPricers are meant to work with L2
- rename Default by DefaultGasPricer
- rename Follower by BasedOnL1GasPricer
- rename LastNBatches by LastNBatchesGasPricer
- rename L2GasPriceSuggestorInterface by L2GasPricer
- change the New methods to return L2GasPricer
- remove GasPriceSuggestor struct
- merge NewL2GasPriceSuggestor with Start method

@ARR552  let me know if you want to talk about it or not, they are just suggestions, so it's up to you, great job 🎉

gasprice/gaspricesuggestor.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/0xPolygonHermez/zkevm-node/state"
+)
+
+// L2GasPricer interface for gas price suggestor.
+type L2GasPricer interface {
+	UpdateGasPriceAvg()
+}
+
+// NewL2GasPriceSuggestor init.
+func NewL2GasPriceSuggestor(ctx context.Context, cfg Config, pool pool, ethMan *etherman.Client, state *state.State) {
```
tclemos, line 18: ~~Suggestor~~ -> **Suggester**

everywhere this word is found it should be fixed.

gasprice/lastnbatches.go
```diff
@@ -2,49 +2,52 @@ package gasprice
...
 	if err != nil {
-		return nil, fmt.Errorf("failed to get last l2 block number, err: %v", err)
+		log.Errorf("failed to get last l2 block number, err: %v", err)
 	}
 	g.cacheLock.RLock()
 	lastL2BlockNumber, lastPrice := g.lastL2BlockNumber, g.lastPrice
 	g.cacheLock.RUnlock()
 	if l2BlockNumber == lastL2BlockNumber {
-		return lastPrice, nil
+		log.Debug("Nothing to update")
```
tclemos, line 49: please, add more context to the log.

"Block is still the same, no need to update the gas price at the moment"

tclemos (changes requested):

gasprice/gaspricesuggestor.go
ARR552: From what I see on the dictionary both words are the same.
image

tclemos: Source?

you can check by yourself on Google: https://translate.google.com/?sl=en&tl=es&text=Suggestor%0A&op=translate

image

ARR552: Ok, I'll change it

tclemos (approved):

arnaubennassar (approved):

PR #1561, opened 2023-01-18
Merge release/v0.0.1 into develop

### LINKED ISSUE #1558: Merge release/v0.0.1 into develop

kind84:
Closes #1558 .

### What does this PR do?

Merge `release/v0.0.1` changes into `develop`. The original `release/v0.0.1` branch has been left intact to be able to perform hotfixes in case.

### Reviewers

@arnaubennassar 
@tclemos 
@ToniRamirezM 
@ARR552


config/environments/local/local.node.config.toml
```diff
@@ -72,6 +72,7 @@ LastBatchVirtualizationTimeMaxWaitPeriod = "300s"
 WaitBlocksToUpdateGER = 10
 WaitBlocksToConsiderGerFinal = 10
 ElapsedTimeToCloseBatchWithoutTxsDueToNewGER = "60s"
+MinTimeToCloseBatch = "60s"
```
ARR552, line 75: I think we also need to include this field into public.node.config.toml

ToniRamirezM (approved):

ARR552 (changes requested):

config/environments/local/local.node.config.toml
kind84: @ARR552 I didn't add it because there is no `[Sequencer]` section in the `public.node.config.toml` file. I think it's because at this moment the public node is meant to rely on other nodes for sequencing batches. Does it make sense to you?

ToniRamirezM:
LGTM

ToniRamirezM (approved):

agnusmor (commented):
@kind84 In line 267 of the aggregator.go file I think we are missing a.endProofVerification() after the a.resetVerifyProofTime():
266:			a.resetVerifyProofTime()
267:			a.endProofVerification()

if we not do the a.endProofVerification() the a.verifyingProof is not set to false and it will not process new verifications anymore.

ARR552 (approved):

kind84:
> @kind84 In line 267 of the aggregator.go file I think we are missing a.endProofVerification() after the a.resetVerifyProofTime(): 266: a.resetVerifyProofTime() 267: a.endProofVerification()
> 
> if we not do the a.endProofVerification() the a.verifyingProof is not set to false and it will not process new verifications anymore.

nice catch @agnusmor!
I got misled by the new verification mechanism.
Please take a look at my changes here: 5faad485627ca94acc8953c46aface600b83fd2b, I also tried to make sure the behavior is the same of the previous version (the sequence of operation in `sendFinalProof` after the verification is not exactly the same, due to the new `handleMonitoredTxResult` method, but the behavior should not change).

agnusmor (approved):

PR #1602, opened 2023-01-27
Merge develop into new sequencer branch

ToniRamirezM:
Merges develop into new sequencer branch


test/benchmarks/sequencer/pool_processing_test.go
```diff
@@ -0,0 +1,251 @@
+package sequencer
```
Psykepro, line 1: We don't need this file. It is not used anymore.

Psykepro (approved):

PR #1603, opened 2023-01-27
[Aggregator] Add recovery mechanism for stuck proofs

kind84:
Closes #1516 .

### Background

There may be chances in which a proof under generation by a prover end up being lost (network failure, db failure, ...).
These proofs stays forever in our `proof` table because they looked to be reserved by a prover, thus the aggregator does not send them to other provers.
For this reason we needed a mechanism to clear those stuck proofs from the table.

### Proposal

This PR  introduces a new method in the `Aggregator` called `cleanupLockedProofs` that continuously checks for locked proofs in the table and if any it clears them out.
The frequency of this check and the threshold from which a proof is considered stuck are configurable with `TimeCleanupLockedProofs` and `GeneratingProofCleanupThreshold` respectively.

The `Generating` (`bool`) field has been refactored to `GeneratingSince` (`*time.Time`) to allow the time comparison against the `GeneratingProofCleanupThreshold` config value. If the field is `nil` it means that the proof is not currently being generating by a prover, otherwise we have the time in which the generation started.

Additional two fields have been added to the `Proof` struct: `CreatedAt` and `UpdatedAt`. For now they don't have a specific usecase, but the may come useful in debugging and troubleshooting.

Other stuff included:

- added UT in `state/pgstatestorage_test.go` for `CleanupLockedProofs`
- fixed `batchsender` script
- added Aggregator section in `config/config_test.go`
- Makefile `sail` and `resail` commands rebranded `ship` and `reship`
- cleaned up `go.sum` by running `go mod tidy -compat=1.17` (:warning: if we need to keep compatibility this can be reverted and run `go mod tidy -go=1.16 && go mod tidy -go=1.17` instead)

### Reviewers

Main reviewers:

@agnusmor 
@ToniRamirezM

PR #1604, opened 2023-01-27
[Aggregator] Add new prover_name field

### LINKED ISSUE #1601: Update Aggregator to use the new Prover name field
Provers are now exposing a new `prover_name` field for a unique identifier across reboots. The `Aggregator` needs to integrate this new field and show it in the logs.

kind84:
Closes #1601.

### What does this PR do?

This PR imports the new `prover_name` field from the updated proto definition in `aggregator.proto`.
This field is the unique identifier of the Prover that does not change across Prover restarts. 
Here it has been included in all the Prover-related logs and in the `state.Proof` struct.

### Reviewers

Main reviewers:

@agnusmor 
@ToniRamirezM


ToniRamirezM (approved):

db/migrations/state/0006.sql
```diff
@@ -1,15 +1,28 @@
...
 
 -- +migrate Up
 ALTER TABLE state.proof
+ADD COLUMN prover_id VARCHAR;
+UPDATE state.proof
+SET prover_id = prover;
+UPDATE state.proof
+SET prover = NULL;
+ALTER TABLE state.proof
 DROP COLUMN IF EXISTS generating;
```
agnusmor, line 26: To be consistent with the "migrate down" steps we can add an UPDATE query to update the field generating_since to NOW() when generating = TRUE

kind84: @agnusmor, good point. I updated the migration file to apply your suggestion. I did this in #1603 because the code is related to that PR: ec4a096562dff1f26ef4a43ed71be7db54efe3f3.
I also addressed the `Migrate down` step to take care of setting the `generating` field to `true` when the `generating_since` was not null.

agnusmor (approved):

PR #1734, opened 2023-03-03
Change %w for %v in formated logs

### LINKED ISSUE #1518: Review logs to use `%v` and `%w` correctly
When we log errors using `%w` but the error is actually not wrapped to contain the trace, the error message is actually not logged. A pattern that we could use is to:

- Always log the error with `%v` at the deepest point in the stack (when it's generated), and wrap it here
- Log again in the highest point in the stack using `%w`

arnaubennassar:
Closes #1518

### DIFF
--- a/aggregator/aggregator.go
+++ b/aggregator/aggregator.go
@@ -118,7 +118,7 @@ func (a *Aggregator) Start(ctx context.Context) error {
 	// Delete ungenerated recursive proofs
 	err := a.State.DeleteUngeneratedProofs(ctx, nil)
 	if err != nil {
-		return fmt.Errorf("failed to initialize proofs cache %w", err)
+		return fmt.Errorf("failed to initialize proofs cache %v", err)
 	}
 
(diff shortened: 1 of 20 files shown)


C001-developer:
could you explain why change the format %w to %v?

tclemos (changes requested):
As I've mentioned in the issue: 
- we must always return errors with `%w`
- we must never log errors with `%w`

Please, review the `returned errors` to keep the `%w` instead of removing it, and please keep the changes to all the `logs`

tclemos:
> could you explain why change the format %w to %v?

The explanation is in the issue when an error is logged with a `%w` instead of the `%v`, the error message is not logged properly.

C001-developer:
> > could you explain why change the format %w to %v?
> 
> The explanation is in this issue #1518, when an error is logged with a `%w` instead of the `%v`, the error message is not logged properly.

Can you share the detail example of it? 
Because , AFAIK, %w will return the wrapped one if error implement Wrap, in the other case %w is same with %v.

For me, the reason of log not working properly is due to the wrong implementation of Wrap function

PR #1747, opened 2023-03-05
fix: fix not append logs when tx was reverted

### LINKED ISSUE #1746: Show event logs in reverted tx hash
Event logs are shown when using eth_getLogs on a failed transaction hash:

https://explorer.public.zkevm-test.net/tx/0xca5ec141dd945c72d8db7fb6be12add20bf10ba5d91c61b236d8f9a528d39d77/logs

```typescript
import { ethers } from 'ethers'

const provider = new ethers.providers.JsonRpcProvider(
  'https://rpc.public.zkevm-test.net',
)

const main = async () => {
  const logs = await provider.send('eth_getLogs', [
    {
      fromBlock: '4426',
      toBlock: '4426',
    },
  ])
  console.log(logs)
}

main()

```

graykode:
Closes #1746.

### What does this PR do?

Skip event logs on failed tx.

### Reviewers

Main reviewers:

- @John
- @Doe

Codeowner reviewers:

- @Alice
- @Bob

### DIFF
--- a/jsonrpc/endpoints_eth.go
+++ b/jsonrpc/endpoints_eth.go
@@ -347,7 +347,14 @@ func (e *EthEndpoints) internalGetLogs(ctx context.Context, dbTx pgx.Tx, filter 
 
 	result := make([]rpcLog, 0, len(logs))
 	for _, l := range logs {
-		result = append(result, logToRPCLog(*l))
+		log := logToRPCLog(*l)
+		receipt, err := e.state.GetTransactionReceipt(ctx, log.TxHash, dbTx)
+		if err != nil {
+			return rpcErrorResponse(defaultErrorCode, "failed to get logs from state", err)
+		}
+		if receipt.Status == types.ReceiptStatusSuccessful {
+			result = append(result, log)
+		}
 	}
 
 	return result, nil


tclemos:
Hey @graykode thanks for your feedback on this.

I'll take a look into that to make sure we don't return the logs from reverted transactions when calling `eth_getLogs`.

The solution you propose fixes the problem, but solving it in the `RPC` layer will add unnecessary overhead to the `State` layer. 

I'll be performing an investigation around this topic to make sure the behavior works as is in Ethereum and also want to try a different solution to reduce the overhead in the `State` layer.

graykode:
@tclemos Cool I'll check it out too in Ethereum

graykode:
@tclemos I have a question, https://github.com/0xPolygonHermez/zkevm-node/blob/develop/jsonrpc/endpoints_eth.go#L343 already uses context related to `State`. is this okay?

tclemos:
> @tclemos I have a question, https://github.com/0xPolygonHermez/zkevm-node/blob/develop/jsonrpc/endpoints_eth.go#L343 already uses context related to `State`. is this okay?

yep, the `State` is where everything is stored and the context is shared from the top layer until the bottom, so we can control the whole request.

graykode:
@tclemos Then, isn't this also the overhead caused by using `State`?

graykode:
Also I find bug that failed txs(reverted) are not mined. It was denined by rpc node.

tclemos:
> @tclemos Then, isn't this also the overhead caused by using `State`?

No, the overhead is calling the state for every log, we can filter it directly when reading the logs from the db for example, but we could also avoid storing these logs if they are not needed, I still need to investigate if they should affect the state somehow.

arnaubennassar:
@cla-bot check

graykode:
@tclemos 

I checked the code in go-ethereum and found that event logs are not recorded in a block when transactions fail. This causes issues when trying to debug failed transactions using tools like etherscan, which execute `traceTransaction` when displaying event logs.

Here are the reasons I found:

1. The `makeLog` function (in [instructions.go](https://github.com/ethereum/go-ethereum/tree/master/core/vm/instructions.go#L837)) creates a log for each event. It records the log using `Addlog` (in [instructions.go](https://github.com/ethereum/go-ethereum/tree/master/core/vm/instructions.go#L851)) if the transaction was not failed.
2. An opcode is run by the evm.interpreter.Run function (in [evm.go](https://github.com/ethereum/go-ethereum/tree/master/core/vm/evm.go#L235)).
3. However, the stateDB is rolled back to the snapshot before execution (in [evm.go](https://github.com/ethereum/go-ethereum/tree/master/core/vm/evm.go#L243)). This means that the logs created by `makeLog` are not recorded in the block if the transaction fails.

Therefore, I reverted before commit and added a filter where `reciept.Status = 1` (txSucceeded) in pgstatestorage: https://github.com/0xPolygonHermez/zkevm-node/pull/1747/commits/c060e9bc9851dd30f4db0abef065b95a0e2d8076

tclemos:
Yes, nice investigation.

The approach to filter logs in the query DB is correct when trying to avoid the overhead by calling the DB multiple times, but the change to the DB query is wrong.

Inner join is meant to join tables, so when adding the `receipt` table to the query, we should be able to join it by matching the `tx hash`, as we do when we join the table `state.transaction` to the `state.log` table, and then, we must add the receipt status condition to the `where` clause.

So, instead of having this:
```go
INNER JOIN state.receipt r ON r.status = 1
```

we should have something like this:
```go 
INNER JOIN state.receipt r ON r.tx_hash = t.hash
...
WHERE ...
AND r.status = 1
```

arnaubennassar:
@graykode This will be fixed in the [executor](https://github.com/0xPolygonHermez/zkevm-prover) (the actual EVM implementation), so this logs are not created in the first place.

We expect this changes to reach testnet in ~24 hours. Thanks for the feedback and the effort 💜

graykode:
@tclemos Then, can I open new pr about fixing postgresql query.

tclemos:
@graykode It's not needed, the problem is going to be solved in a step before, the current query will work and the data will be filtered when the tx is executed, not generating logs for reverted TXs.

PR #1755, opened 2023-03-06
Fix/jsonrpc call incompatible

icydark:
### What does this PR do?

This PR is about fixing the RPC discrepancy of eth_call. 

commit 80faa38 
fixes "eth_call" does not support requests with block hash instead of block number

commit 9367571
fixes "eth_call" returns an error when the sender addr is nil, or when the gas limit is higher than header.GasLimit

### Reviewers

Main reviewers:

- @ToniRamirezM 
- @ARR552

### DIFF
--- a/jsonrpc/codec.go
+++ b/jsonrpc/codec.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 
 	"github.com/0xPolygonHermez/zkevm-node/encoding"
+	"github.com/ethereum/go-ethereum/rpc"
 	"github.com/jackc/pgx/v4"
 )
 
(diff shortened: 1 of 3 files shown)


tclemos:
Hey @icydark, thanks for your feedback on this. 

I'll take a look into this issue and review your solution to figure out what's it the best way to add support to that into the current code.

Also, I want to add some tests to make sure the problem doesn't return in the future.

arnaubennassar:
@cla-bot check

arnaubennassar:
@icydark have you signed the CLA?

icydark:
> @icydark have you signed the CLA?

Not yet. How can I sign the CLA? Should I open a PR in the cla-bot repository?

PR #1779, opened 2023-03-09
improve: implementing todos

### LINKED ISSUE #1752: Reviewing all TODOs in sequencer
Reviewing all TODOs if they still make sense and implementing missing things.

Psykepro:
Closes #1752

### What does this PR do?

Fixing todos.

### Reviewers

Main reviewers:
- @ToniRamirezM 
- @ARR552 

Codeowner reviewers:
- @ToniRamirezM


config/environments/local/local.node.config.toml
```diff
@@ -93,8 +93,15 @@ MaxTxLifetime = "3h"
...
 		PrivateKeys = [{Path = "/pk/sequencer.keystore", Password = "testonly"}]
+<<<<<<< HEAD
 	[Sequencer.DBManager]
 		PoolRetrievalInterval = "500ms"
+=======
+	[Sequencer.Worker]
+		ResourceCostMultiplier = 1000
+	[Sequencer.DbManager]
+		SleepDurationInSec = "5s"
+>>>>>>> e5e4fe5e (wip)
```
tclemos, line 104: Please, fix conflict before merging

sequencer/worker.go
```diff
@@ -55,20 +56,17 @@ func (w *Worker) AddTxTracker(ctx context.Context, tx *TxTracker) {
 
 		root, err := w.state.GetLastStateRoot(ctx, nil)
 		if err != nil {
-			// TODO: How to manage this
 			log.Errorf("AddTx GetLastStateRoot error: %v", err)
 			return
```
ARR552, line 60:
```suggestion
			return false
```

sequencer/worker.go
```diff
@@ -55,20 +56,17 @@ func (w *Worker) AddTxTracker(ctx context.Context, tx *TxTracker) {
...
 		}
 		nonce, err := w.state.GetNonceByStateRoot(ctx, tx.From, root)
 		if err != nil {
 			log.Errorf("AddTx GetNonceByStateRoot error: %v", err)
-			// TODO: How to manage this
 			return
```
ARR552, line 65:
```suggestion
			return false
```

sequencer/worker.go
```diff
@@ -55,20 +56,17 @@ func (w *Worker) AddTxTracker(ctx context.Context, tx *TxTracker) {
...
 		}
 		balance, err := w.state.GetBalanceByStateRoot(ctx, tx.From, root)
 		if err != nil {
 			log.Errorf("AddTx GetBalanceByStateRoot error: %v", err)
-			// TODO: How to manage this
 			return
```
ARR552, line 70:
```suggestion
			return false
```

sequencer/worker.go
Psykepro: OMG, why does the IDE even let me compile like that? Good catch!

sequencer/dbmanager.go
```diff
@@ -58,8 +54,7 @@ func (d *dbManager) Start() {
 	go d.loadFromPool()
 	go func() {
 		for {
-			// TODO: Move this to a config parameter
-			time.Sleep(wait * time.Second)
+			time.Sleep(d.cfg.PoolRetrievalInterval.Duration)
```
agnusmor, line 57: PoolRetrieveInterval is used to polling for pending txs in the pool. In this go func we are detecting a L2Reorg, then we need to create an additional parameter to define this polling interval or let it hard-coded to the previous value (wait = 5 seconds)

sequencer/txtracker.go
```diff
@@ -13,46 +13,84 @@ import (
...
 	}
 
-	txTracker.IsClaim = isClaim
-	txTracker.BatchResources.zKCounters = counters
-	txTracker.BatchResources.bytes = uint64(tx.Size())
-	txTracker.HashStr = txTracker.Hash.String()
-	txTracker.FromStr = txTracker.From.String()
 	txTracker.Benefit = new(big.Int).Mul(new(big.Int).SetUint64(txTracker.Gas), txTracker.GasPrice)
-	txTracker.calculateEfficiency(constraints, weights)
+	txTracker.constraints = convertBatchConstraintsToFloat64(constraints)
```
agnusmor, line 89: We are calling convertBatchConstraintsToFloat64, calculating the txTracker.totalWeight and calculateWeightMultipliers each time we are calling newTxTracker, when the result of this functions/calculation is always the same (same values for all the Tx). I think it will more optimal to calculate this in the worker side (and calculate only one time) and then pass the structs/values when calling newTxTracker

sequencer/worker.go
```diff
@@ -55,21 +56,18 @@ func (w *Worker) AddTxTracker(ctx context.Context, tx *TxTracker) {
 
 		root, err := w.state.GetLastStateRoot(ctx, nil)
 		if err != nil {
-			// TODO: How to manage this
 			log.Errorf("AddTx GetLastStateRoot error: %v", err)
-			return
+			return false
```
agnusmor, line 60: we are not managing correctly this scenario, if we got an error when getting the nonce/balance and we return from the function, we are not adding the tx to the worker but I understand we are putting the tx as wip in the database. If this is the case, this tx we will not be processed as it doesn't exists for the worker but we are updating in the pool table as wip, therefore this can cause some confusion about the state of this tx. We need to review how to manage this scenario

sequencer/dbmanager.go
ToniRamirezM: Yes, this MUST be a different configuration parameter.

sequencer/worker.go
Psykepro: For this one, it would be best to return it to pending status and non-wip. but I'll need to introduce a 2nc return value is_wip.

PR #1786, opened 2023-03-09
fix claim pre execution

### LINKED ISSUE #1782: Pre execute tx doesn't get the correct GER
When testing with pre RC14 we realised that the tx for a claim in a empty network was always reverted in the pre-execution (so tx was not added in to the pool).

We check and the GER sent to the executor was always 0x0...0, but we could see that in the DB there were (empty) batches being closed and the GER was updated correctly. Probably the bug comes from not having a single tx in the stateDB

### LINKED ISSUE #1783: jRPC returns "nonce intrinsic error" when sending tx
ir should return standard eth string

### LINKED ISSUE #1784: Pre execution returns error for intrinsic in the wrong case
After running a script that sends many txs one after the other controling the nonce locally, one of them always fail with intrinsic error (nonce).

Looks like the pre-execute has a problem when a nonce is being changed concurrently fast

ToniRamirezM:
Closes #1782
Closes #1783
Closes #1784

### What does this PR do?

Fixes pre execution of claims.

### Reviewers

Main reviewers:

- @arnaubennassar 
- @ARR552 
- @agnusmor


pool/pool.go
```diff
@@ -103,10 +85,9 @@ func (p *Pool) ReorgTx(ctx context.Context, tx types.Transaction) error {
 	// Execute transaction to calculate its zkCounters
 	zkCounters, err := p.PreExecuteTx(ctx, tx)
 	if err != nil {
-		log.Warnf("ignoring error while preexecuting tx: %+v", poolTx)
+		poolTx.ZKCounters = zkCounters
```
tclemos, line 88: Is it ok to continue if the `err != nil`?

arnaubennassar (approved):

pool/pool.go
ToniRamirezM: Yes! We try our best to get the proper zkCounters but the state may not be ready yet to execute this Tx. Imagine a sc deployment followed by a call to that SC. If the pre-execution of the call happens before the SC is stored in the state, it will fail in pre-execution, but not on real execution. We won't have quality zk counters, but we must process that tx.

PR #1792, opened 2023-03-09
add support to jrpc to accept short hashes and addresses

### LINKED ISSUE #1791: Support 0x00 as a valid hash in JSON RPC
Example:

for the endpoint `eth_getStorageAt`, this should be a valid input:

```json
{
	"jsonrpc":"2.0", 
	"method": "eth_getStorageAt", 
	"params": [
		"0xA766B25F0577EB4802e659dF0C1bc4b3C8A0fDeD", 
		"0x00",
		"latest"
	], 
	"id": 1
}
```

Note that now this call would return `Invalid Params`, but if the 0x00 is changed to `0x0000000000000000000000000000000000000000000000000000000000000000`, then the request is valid. This could affect any other endpoint that has a input param that is a hash

tclemos:
Closes #1791.

### What does this PR do?

add support to jrpc to accept short hashes and addresses strings like 0x00

### Reviewers

Main reviewers:

@arnaubennassar 
@ToniRamirezM

### DIFF
--- a/jsonrpc/types.go
+++ b/jsonrpc/types.go
@@ -2,6 +2,7 @@ package jsonrpc
 
 import (
 	"context"
+	"fmt"
 	"math/big"
 	"strconv"
 	"strings"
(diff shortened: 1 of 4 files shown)


jsonrpc/types.go
```diff
@@ -116,6 +117,52 @@ func encodeToHex(b []byte) []byte {
...
 }
 
+// argHash represents a common.Hash that accepts strings
+// shorter than 64 bytes, like 0x00
+type argHash common.Hash
+
+// UnmarshalText unmarshals from text
+func (arg *argHash) UnmarshalText(input []byte) error {
+	if !strings.HasPrefix(string(input), "0x") {
+		return fmt.Errorf("invalid address, it needs to be a hexadecimal value starting with 0x")
```
ToniRamirezM, line 127: Should this be "invalid hash" ?

ToniRamirezM (approved):

PR #1803, opened 2023-03-13
ForkID logic for an old batch and fix error handling in execute batch

### LINKED ISSUE #1780: ForkId update
We can have 2 situations for updating the forID:
1- It is a scheduled update and we wait until all the virtual batches are verified.
2- We face problems in a batch and we need to update the forkID.
For the first situation we don't need to add extra logic.
Regarding the second situation:
When a forkID is detected, the synchronizer has to force a trustedReorg. This will force the sequencer to start building batches again using the correct forkID.
If the event has the batchNumber 15, the synchronizer will delete batches after batch 14.
we would also need to delete the proof table for proofs after proof 14

### LINKED ISSUE #1830: Propery handle EXECUTOR_ERROR_UNSUPPORTED_FORK_ID
It has been detected that if this error happens during synchronization, no error is returned and so state root = 0x000... is used.

ARR552:
Closes #1780 
Closes #1830 

### What does this PR do?

ForkID logic for an old batch and fix error handling in execute batch

Following cases are tested:
- Update forkID for a past batch
- Update forkId for a future batch
- Update several times the forkId in the same batch

### Reviewers

- @ToniRamirezM 
- @agnusmor 
- @cool-develope 
- @arnaubennassar


etherman/types.go
```diff
@@ -66,3 +67,10 @@ type SequencedForceBatch struct {
 	Nonce       uint64
 	polygonzkevm.PolygonZkEVMForcedBatchData
 }
+
+// ForkID is a sturct to track the ForkID event.
+type ForkID struct {
```
ToniRamirezM, line 72: What do you think about merging this with state/forkid.go ? Maybe moving everything to the state or the etherman. (IMO this is more related to state)

state/pgstatestorage.go
```diff
@@ -2239,6 +2267,19 @@ func (p *PostgresStorage) CountReorgs(ctx context.Context, dbTx pgx.Tx) (uint64,
 	return count, nil
 }
 
+// GetForkIDTrustedReorg returns the forkID
+func (p *PostgresStorage) GetForkIDTrustedReorg(ctx context.Context, forkID uint64, version string, dbTx pgx.Tx) (uint64, error) {
```
ToniRamirezM, line 2271: I suggest adding Count to the name of the function

ToniRamirezM (changes requested):

etherman/types.go
ARR552: I see your point, but the problem is that the state.ForkID struct doesn't fit very well with the needs of the event. The synchronizer translates the info received from L1 to the infor required by the state. We are doing the same thing with other structs such as `etherman.SequencedBatch` and `state.Batch and state.VirtualBatch`

ToniRamirezM (approved):

PR #1814, opened 2023-03-14
fix: moving print for storing tx in finalizer.

### LINKED ISSUE #1812: New wipBatch shouldn't store txs
In `v0.0.3-RC17`:

```log
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x11d4032]

goroutine 145 [running]:
github.com/0xPolygonHermez/zkevm-node/sequencer.(*finalizer).handleSuccessfulTxProcessResp(0xc00055cf00, {0x1918478, 0xc0000c4000}, 0x0, 0xc017efaa90)
	/src/sequencer/finalizer.go:389 +0x152
github.com/0xPolygonHermez/zkevm-node/sequencer.(*finalizer).processTransaction(0xc00055cf00, {0x1918478, 0xc0000c4000}, 0x0)
	/src/sequencer/finalizer.go:363 +0x7a5
github.com/0xPolygonHermez/zkevm-node/sequencer.(*finalizer).newWIPBatch(0xc00055cf00, {0x1918478?, 0xc0000c4000})
	/src/sequencer/finalizer.go:260 +0xf5
github.com/0xPolygonHermez/zkevm-node/sequencer.(*finalizer).finalizeBatch(0xc00055cf00, {0x1918478, 0xc0000c4000})
	/src/sequencer/finalizer.go:244 +0xb0
github.com/0xPolygonHermez/zkevm-node/sequencer.(*finalizer).finalizeBatches(0xc00055cf00, {0x1918478, 0xc0000c4000})
	/src/sequencer/finalizer.go:218 +0x33c
github.com/0xPolygonHermez/zkevm-node/sequencer.(*finalizer).Start(0xc00055cf00, {0x1918478?, 0xc0000c4000}, 0x56f3aa?, 0x1?)
	/src/sequencer/finalizer.go:131 +0x209
created by github.com/0xPolygonHermez/zkevm-node/sequencer.(*Sequencer).Start
	/src/sequencer/sequencer.go:157 +0xd0c
```

The panic itself is already fixed in develop, but not the underlying problem:

- `newWIPBatch` is called, current batch has no txs, so it's re-executed to ensure GER update is reflected in the new state root.
- `processTransaction` returns successful execution, however there are no txs being processed
- `handleSuccessfulTxProcessResp` is called, but the tx that is passed is `nil`! This function shouldn't be called in this context

So here is were we have the crazy situation: our protocol can alter the state without executing transactions 🤯 : An empty batch with a new GER creates a state update. We need to:

1. Handle this special case by updating the state root of the batch that we're closing, without trying to store transactions and L2 blocks
2. Review that we correctly get the state root for next batches, RPC calls, reboots, ... in this situations

Psykepro:
Closes #1812

### What does this PR do?

Moving tx print for storing tx from `handleSuccessfulTxProcessResp` to `storeProcessedTx` where we already check if the tx is `nil` to return.

### Reviewers

- @ToniRamirezM
- @arnaubennassar 
- @agnusmor 

Main reviewers:

- @ToniRamirezM
- @arnaubennassar 

Codeowner reviewers:

- @ToniRamirezM

### DIFF
--- a/sequencer/finalizer.go
+++ b/sequencer/finalizer.go
@@ -254,16 +254,6 @@ func (f *finalizer) newWIPBatch(ctx context.Context) (*WipBatch, error) {
 	defer f.sharedResourcesMux.Unlock()
 
 	var err error
-	// Passing the batch without txs to the executor in order to update the State
-	if f.batch.countOfTxs == 0 {
-		// backup current sequence
-		err = f.processTransaction(ctx, nil)
-		for err != nil {
-			log.Errorf("failed to process tx, err: %v", err)
-			err = f.processTransaction(ctx, nil)
-		}
-	}
-
 	if f.batch.stateRoot.String() == "" || f.batch.localExitRoot.String() == "" {
 		return nil, errors.New("state root and local exit root must have value to close batch")
 	}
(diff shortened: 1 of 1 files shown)


ToniRamirezM (approved):

arnaubennassar:
I still don't understand why we need to call a method named `handleSuccessfulTxProcessResp` that in return will call `storeProcessedTx` if we don't have a tx to store... Wouldn't it be better to just update the batch, without dealing with the concept of tx?

If you don't feel confortable doing this, it's ok, but I think that this is missleading and can introduce problems. Feel free to merge if you have already considered what I'm saying

Psykepro:
> I still don't understand why we need to call a method named `handleSuccessfulTxProcessResp` that in return will call `storeProcessedTx` if we don't have a tx to store... Wouldn't it be better to just update the batch, without dealing with the concept of tx?
> 
> If you don't feel confortable doing this, it's ok, but I think that this is missleading and can introduce problems. Feel free to merge if you have already considered what I'm saying

Okay will move the `storeProcessedTx` outside of the `handleSuccessfulTxProcessResp`.

agnusmor (approved):

sequencer/finalizer.go
```diff
@@ -360,22 +360,29 @@ func (f *finalizer) processTransaction(ctx context.Context, tx *TxTracker) error
...
+	if len(result.Responses) > 0 && tx != nil {
+		err = f.handleTxProcessResp(ctx, tx, result, oldStateRoot)
+		if err != nil {
+			return err
+		}
 	}
 
+	// Update in-memory batch and processRequest
+	f.processRequest.OldStateRoot = result.NewStateRoot
+	f.batch.stateRoot = result.NewStateRoot
```
arnaubennassar, line 373: Is this stored on the DB as well?

arnaubennassar (changes requested):

arnaubennassar (approved):

ToniRamirezM (approved):

PR #1817, opened 2023-03-14
Add IP to Pool txs

### LINKED ISSUE #1759: Add origin IP of a TX in the pool

### LINKED ISSUE #1761: Log OOC txs in the pool including the source IP

ToniRamirezM:
Closes #1759
Closes #1761

### What does this PR do?

Stores sender IP in the Pool and logs OOC events during pre-execution in the pool.

### Reviewers

Main reviewers:

- @ARR552 
- @agnusmor 
- @tclemos


state/state_test.go
```diff
@@ -2872,3 +2872,26 @@ func TestExecuteWithoutUpdatingMT(t *testing.T) {
 	assert.Equal(t, executorclientpb.RomError(1), processBatchResponse.Responses[1].Error)
 	assert.Equal(t, "0000000000000000000000000000000000000000000000000000000000000001", hex.EncodeToString(processBatchResponse.Responses[1].ReturnValue))
 }
+
+func TestStoreEvent(t *testing.T) {
```
ARR552, line 2876: It is not a big deal but, shouldn't this test be in `pgstatestorage_test.go` file??

db/migrations/state/0004.sql
```diff
@@ -0,0 +1,12 @@
+-- +migrate Up
+CREATE TABLE state.event
```
ARR552, line 2: Not sure about storing this table in the state schema. What about create a diferent schema named "motinorization" for instance and store this table there. We can use the same db (stateDB) but without mixing the monitor info with the node state info. WDYT?

ToniRamirezM: Yes, I agree. We also have the "degug" table. I think we should open a new issue and to this in the (near) future.

state/state_test.go
ToniRamirezM: Yes

jsonrpc/endpoints_eth.go
```diff
@@ -614,11 +614,11 @@ func (e *EthEndpoints) newPendingTransactionFilter(wsConn *websocket.Conn) (inte
 // SendRawTransaction has two different ways to handle new transactions:
 // - for Sequencer nodes it tries to add the tx to the pool
 // - for Non-Sequencer nodes it relays the Tx to the Sequencer node
-func (e *EthEndpoints) SendRawTransaction(input string) (interface{}, rpcError) {
+func (e *EthEndpoints) SendRawTransaction(input, ip string) (interface{}, rpcError) {
```
tclemos, line 617: this breaks the eth_sendRawTransaction interface and should be reverted

tclemos (changes requested):
Actions:
- Changes to the `endpoints_eth.go` must be reverted.
- `handler.go` must not have any hardcoded method, it's a dynamic and agnostic component.
- IP tracking must be controlled by the HTTP layer ONLY, which is the `server.go`
- `Request` structure is a `JSON RPC` request not an `HTTP Request` and a `JSON RPC` request doesn't have an IP property, this change is affecting ALL the `JSON RPC` endpoints.

Suggestions:
- Add the needed information to the `context`
- Drill down the context from the `HTTP layer`, to the `handler` and then to the `RPC endpoints`
- Redesign all the `RPC endpoints` signatures to have the first parameter as the `context`
- Get the IP or any other information from the context and use as needed by any `RPC endpoint`

synchronizer/synchronizer.go
```diff
@@ -1143,7 +1143,7 @@ func (s *ClientSynchronizer) reorgPool(dbTx pgx.Tx) error {
 
 	// Add txs to the pool
 	for _, tx := range txs {
-		err = s.pool.StoreTx(s.ctx, *tx)
+		err = s.pool.StoreTx(s.ctx, *tx, "reorg")
```
tclemos, line 1146: "reorg" as an IP? why it is needed to change the IP at this moment?

ToniRamirezM:
> Actions:
> 
>     * Changes to the `endpoints_eth.go` must be reverted.
> 
>     * `handler.go` must not have any hardcoded method, it's a dynamic and agnostic component.
> 
>     * IP tracking must be controlled by the HTTP layer ONLY, which is the `server.go`
> 
>     * `Request` structure is a `JSON RPC` request not an `HTTP Request` and a `JSON RPC` request doesn't have an IP property, this change is affecting ALL the `JSON RPC` endpoints.
> 
> 
> Suggestions:
> 
>     * Drill down the `HTTP Request` from the `HTTP layer`, to the `handler` and then to the `RPC endpoints`
> 
>     * Redesign all the `RPC endpoints` signatures to have the first parameter as the `HTTP Request`
> 
>     * Get the IP or any other information from the `HTTP Request` and use it as needed by any `RPC endpoint`

Ok! I will try this approach.

synchronizer/synchronizer.go
tclemos: yes, but I understand this is more related to the status than the IP. 

We just need this because we are re-adding to the pool a tx that was already added by a user before and was affected by the reorg, my suggestion is that either we save the IP to the transaction table(not recommended) or save the IP of the synchronizer machine(we can still use this information as a debug if we use a local IP instead of a public one). It's a bit weird to me to have a column that should contain an IP and find a word.

tclemos:
> > Actions:
> > ```
> > * Changes to the `endpoints_eth.go` must be reverted.
> > 
> > * `handler.go` must not have any hardcoded method, it's a dynamic and agnostic component.
> > 
> > * IP tracking must be controlled by the HTTP layer ONLY, which is the `server.go`
> > 
> > * `Request` structure is a `JSON RPC` request not an `HTTP Request` and a `JSON RPC` request doesn't have an IP property, this change is affecting ALL the `JSON RPC` endpoints.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Suggestions:
> > ```
> > * Drill down the `HTTP Request` from the `HTTP layer`, to the `handler` and then to the `RPC endpoints`
> > 
> > * Redesign all the `RPC endpoints` signatures to have the first parameter as the `HTTP Request`
> > 
> > * Get the IP or any other information from the `HTTP Request` and use it as needed by any `RPC endpoint`
> > ```
> 
> Ok! I will try this approach.

let me know if you need any help! 👍

synchronizer/synchronizer.go
ToniRamirezM: The goal of this information is to create scripts that will ban IPs based on behaviour. I don't want the IP of the syncrhonizer to get banned by error.

db/migrations/state/0004.sql
ToniRamirezM: It is easier for devops to read a table thant the logs.

db/migrations/pool/0003.sql
```diff
@@ -0,0 +1,13 @@
+-- +migrate Up
```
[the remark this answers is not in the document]
ToniRamirezM: Issue created

tclemos (approved):

PR #1834, opened 2023-03-15
Feature/#1833 execute batch

### LINKED ISSUE #1833: Avoid execute twice each batch when syncing from scratch

ARR552:
Closes #1833 

### What does this PR do?

avoids to execute each batch twice when syncing from scratch

### Reviewers

- @C001-developer 
- @ToniRamirezM 
- @arnaubennassar


state/state.go
```diff
@@ -823,19 +823,19 @@ func (s *State) ProcessAndStoreClosedBatch(
...
 		err = s.StoreTransactions(ctx, processingCtx.BatchNumber, processedBatch.Responses, dbTx)
 		if err != nil {
-			return err
+			return nil, err
 		}
 	}
 
 	// Close batch
-	return s.closeBatch(ctx, ProcessingReceipt{
+	return processed, s.closeBatch(ctx, ProcessingReceipt{
```
ToniRamirezM, line 838: Consider returning just the new state root if that is enough.

ToniRamirezM (approved):

C001-developer (approved):

docker-compose.yml
tclemos, line 1: I think we can't remove these lines from the docker-compose because this was added by @arnaubennassar in order to be compatible with the instructions to set up the public environment.

@arnaubennassar can you confirm if this can be removed or not?

PR #1835, opened 2023-03-15
Fix/unsigned tx state root

### LINKED ISSUE #1824: eth_call fails when batch has more than 1 tx for the same acc
The way the node process unsigned transactions is by getting the l2 block, then identifying the batch related to this block in order to set the state root we must use when processing the transaction. The code always uses the state root of the latest l2 block related to the batch, so depending on the order of the transactions we might have an false `ProviderError: nonce intrinsic error`

Here is how it works:

Imagine a scenario like this:

batch 1
block 1 acc 1 tx 1 nonce 1
block 2 acc 2 tx 1 nonce 1
block 3 acc 1 tx 2 nonce 2

if I do a call for block 2, the nonce of the acc 1 should be 1... but since we have the block 3 attached to batch 1, we use the state root of block 3 instead of the root of block 2... and this causes the problem

tclemos:
Closes #1824.

### What does this PR do?

Fix unsigned tx to use the specified l2block state root instead of the last l2block state root related to the batch.

### Reviewers

Main reviewers:

@ToniRamirezM
@ARR552

### DIFF
--- a/jsonrpc/endpoints_eth.go
+++ b/jsonrpc/endpoints_eth.go
@@ -73,12 +73,7 @@ func (e *EthEndpoints) Call(arg *txnArgs, number *BlockNumber) (interface{}, rpc
 			return rpcErrorResponse(defaultErrorCode, "failed to convert arguments into an unsigned transaction", err)
 		}
 
-		var blockNumberToProcessTx *uint64
-		if number != nil && *number != LatestBlockNumber && *number != PendingBlockNumber {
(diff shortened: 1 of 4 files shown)


ToniRamirezM (approved):

ToniRamirezM:
If I understood properly, now we are forcing the nonce not only on pre-execution, but for all the unsigned transactions.

ARR552 (approved):

PR #1899, opened 2023-03-21
allow pool to be blocked

tclemos:
### What does this PR do?

Allows node to block pool

### Reviewers

Main reviewers:

@ToniRamirezM 
@agnusmor 
@ARR552

### DIFF
--- a/pool/pgpoolstorage/pgpoolstorage.go
+++ b/pool/pgpoolstorage/pgpoolstorage.go
@@ -649,3 +649,19 @@ func (p *PostgresPoolStorage) GetAllAddressesBlocked(ctx context.Context) ([]com
 
 	return addrs, nil
 }
+
+// IsAddressBlocked check if an address is added to the blocked table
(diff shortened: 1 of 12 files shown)


pool/pgpoolstorage/pgpoolstorage.go
```diff
@@ -622,3 +622,19 @@ func (p *PostgresPoolStorage) UpdateTxWIPStatus(ctx context.Context, hash common
 	}
 	return nil
 }
+
+// IsAddressBlocked check if an address is added to the blocked table
+func (p *PostgresPoolStorage) IsAddressBlocked(ctx context.Context, address common.Address) (bool, error) {
```
ToniRamirezM, line 627: The idea is to load the full list during startup. It won't be a long list and can be kept in memory.

There must be a go func reloading this is memory each 5 / 10 minutes.

ToniRamirezM (approved):

ToniRamirezM (approved):

arnaubennassar (approved):

PR #1907, opened 2023-03-22
feat: adding check gas price of tx to be greater that min gas price and min suggested gas price before adding to pool.

### LINKED ISSUE #1809: Filter transactions based on gas price
Prevent txs from being addded into the pool

- Reject txs where the gas price < X (config param)
- Reject txs where gas price < MIN gas price suggested by the RPC in the last X mins (config param). This data can be obtained from ppol DB

Psykepro:
Closes #1809.

### What does this PR do?

Adding check gas price of tx to be greater than `MinGasPrice = 1000000000` and min suggested gas price for last `MinSuggestedGasPriceInterval = "5m"` before adding to the pool.

### Reviewers

Main reviewers:
- @arnaubennassar 
- @tclemos 
- @ARR552

Codeowner reviewers:
- @tclemos
- @ARR552

### DIFF
--- a/config/config_test.go
+++ b/config/config_test.go
@@ -258,6 +258,14 @@ func Test_Defaults(t *testing.T) {
 			expectedValue: 30000,
 		},
 		{
+			path:          "Pool.MinGasPrice",
+			expectedValue: uint64(1000000000),
+		},
+		{
+			path:          "Pool.MinSuggestedGasPriceInterval",
+			expectedValue: types.NewDuration(5 * time.Minute),
+		},
+		{
 			path:          "Pool.DB.User",
 			expectedValue: "pool_user",
 		},
--- a/pool/errors.go
+++ b/pool/errors.go
@@ -42,6 +42,6 @@ var (
 	// ErrGasUintOverflow is returned when calculating gas usage.
 	ErrGasUintOverflow = errors.New("gas uint64 overflow")
 
-	// ErrIntrinsicGasPrice is returned if the transaction has specified lower gas price than the minimum allowed.
-	ErrIntrinsicGasPrice = errors.New("intrinsic gas price too low")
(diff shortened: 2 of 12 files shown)


tclemos (approved):

pool/errors.go
```diff
@@ -41,4 +41,7 @@ var (
 
 	// ErrGasUintOverflow is returned when calculating gas usage.
 	ErrGasUintOverflow = errors.New("gas uint64 overflow")
+
+	// ErrIntrinsicGasPrice is returned if the transaction has specified lower gas price than the minimum allowed.
+	ErrIntrinsicGasPrice = errors.New("intrinsic gas price too low")
```
arnaubennassar, line 46:
```suggestion
	ErrGasPrice = errors.New("gas price too low")
```

We use intrinsic for errors from the rom that prevent the tx from executing (nonce, balance, signature, ...)

arnaubennassar (approved):

config/config_test.go
```diff
@@ -257,6 +257,14 @@ func Test_Defaults(t *testing.T) {
 			path:          "Pool.MaxTxDataBytesSize",
 			expectedValue: 30000,
 		},
+		{
+			path:          "Pool.MinGasPrice",
+			expectedValue: uint64(1000000000),
```
agnusmor, line 262: if I am not wrong the L2GasPriceSuggester has the param "DefaultGasPriceWei" that is the min GasPrice suggested. It is stored in the pool.gas_price table if the gas price calculated by the GasPricerSuggester is minor than this value. In this case it's not needed to have this MinGasPrice param as this min value is manager by the GasPricerSuggester. If we keep this MinGasPrice value and it is not equal to the GasPricerSuggester we can have issues or misunderstanding in the behaviour of the min gas price management

ARR552: you are right @agnusmor It is redundant

PR #1914, opened 2023-03-23
fix: fixing comments.

Psykepro:
Related #1809

### What does this PR do?

Fixing comments from previous closed PR

### Reviewers

Main reviewers:

- @agnusmor 
- @ARR552 
- @tclemos  
- @arnaubennassar   

Codeowner reviewers:

- @tclemos  
- @ARR552


pool/pool.go
```diff
@@ -271,6 +287,20 @@ func (p *Pool) validateTx(ctx context.Context, tx types.Transaction) error {
...
 
+func (p *Pool) pollMinSuggestedGasPrice(ctx context.Context) {
+	fromTimestamp := time.Now().UTC().Add(-p.cfg.MinSuggestedGasPriceInterval.Duration)
+	gasPrice, err := p.storage.MinGasPriceSince(ctx, fromTimestamp)
+	if err == state.ErrNotFound {
+		log.Warnf("No suggested min gas price since: %v", fromTimestamp)
+	} else if err != nil {
+		log.Errorf("Error getting min gas price since: %v", fromTimestamp)
+	} else {
+		p.minSuggestedGasPriceMux.Lock()
```
agnusmor, line 298: Could be it will useful to log (in debug mode) the new suggested gas price. I know it will be retrieved from the gasprice table, but maybe is some situations it will be useful to have it in the logs

pool/pool.go
```diff
@@ -33,25 +35,54 @@ var (
...
+		minSuggestedGasPriceMux: new(sync.RWMutex),
 	}
 }
 
+// StartPollingMinSuggestedGasPrice starts polling the minimum suggested gas price
+func (p *Pool) StartPollingMinSuggestedGasPrice(ctx context.Context) {
+	err := p.pollMinSuggestedGasPrice(ctx)
+	if err != nil && err != state.ErrNotFound {
+		log.Fatalf("Error polling min suggested gas price: %v", err)
+	}
```
tclemos, line 63: this code is not needed

Psykepro: It is because we can't start if there is problem with the DB and we don't have value here. Because we will have later on nil pointer reference panic.

pool/pool.go
```diff
@@ -33,25 +35,54 @@ var (
...
+	// Retrying until we have a valid minSuggestedGasPrice
+	for err == state.ErrNotFound {
+		err = p.pollMinSuggestedGasPrice(ctx)
+		log.Infof("Retrying to poll min suggested gas price ...")
+		time.Sleep(retryInterval)
+	}
```
tclemos, line 69:
```go
       for {
                if err = p.pollMinSuggestedGasPrice(ctx); err == nil {
                    break
                }
		log.Infof("Retrying to poll min suggested gas price ...")
		time.Sleep(retryInterval)
	}
```

pool/pool.go
Psykepro: we can't because then we can let tx with smaller than minimum gas price allowed

tclemos: The gas is a suggestion.
The suggestion provider can not suggest it or be out of service at the moment.

RPC is not going to start if we get into this.
RPC is going to misbehave if the l2 gas pricer has stopped working during its execution.

This validation should consider the l2 gas pricer stopping working and having a fallback action, either blocking all transactions or allowing the ones with smaller gas than the minimum.

This locked execution shouldn't stop the RPC to start, the service has only a single endpoint that depends on this feature and many others that don't.

config/config_test.go
```diff
@@ -257,14 +257,19 @@ func Test_Defaults(t *testing.T) {
 			path:          "Pool.MaxTxDataBytesSize",
 			expectedValue: 30000,
 		},
+
 		{
-			path:          "Pool.MinGasPrice",
+			path:          "Pool.DefaultMinGasPriceToSuggest",
```
ARR552, line 262:
```suggestion
			path:          "Pool.DefaultMinGasPriceAllowed",
```

config/default.go
```diff
@@ -22,8 +22,9 @@ MaxConns = 200
 FreeClaimGasLimit = 150000
 MaxTxBytesSize=30132
 MaxTxDataBytesSize=30000
-MinGasPrice = 1000000000
+DefaultMinGasPriceToSuggest = 1000000000
```
ARR552, line 25: same here

pool/pool.go
```diff
@@ -271,6 +292,31 @@ func (p *Pool) validateTx(ctx context.Context, tx types.Transaction) error {
...
 }
 
+func (p *Pool) pollMinSuggestedGasPrice(ctx context.Context) error {
+	fromTimestamp := time.Now().UTC().Add(-p.cfg.MinSuggestedGasPriceInterval.Duration)
+	gasPrice, err := p.storage.MinGasPriceSince(ctx, fromTimestamp)
+	if err == nil {
+		if err == state.ErrNotFound {
+			log.Warnf("No suggested min gas price since: %v", fromTimestamp)
+			return err
+		} else if err != nil {
```
tclemos, line 302: how can the `error be nil` and then be `NotFound` or `not nil`?

pool/pool.go
```diff
@@ -271,6 +292,31 @@ func (p *Pool) validateTx(ctx context.Context, tx types.Transaction) error {
...
+			log.Errorf("Error getting min gas price since: %v", fromTimestamp)
+			return err
+		}
+		// Ensuring we always have suggested minimum gas price
+		if p.minSuggestedGasPrice == nil {
+			p.minSuggestedGasPrice = big.NewInt(0).SetUint64(p.cfg.DefaultMinGasPriceToSuggest)
```
tclemos, line 308: missing lock to updated the field

pool/pool.go
```diff
@@ -271,6 +292,31 @@ func (p *Pool) validateTx(ctx context.Context, tx types.Transaction) error {
 	return nil
 }
 
+func (p *Pool) pollMinSuggestedGasPrice(ctx context.Context) error {
+	fromTimestamp := time.Now().UTC().Add(-p.cfg.MinSuggestedGasPriceInterval.Duration)
+	gasPrice, err := p.storage.MinGasPriceSince(ctx, fromTimestamp)
+	if err == nil {
+		if err == state.ErrNotFound {
+			log.Warnf("No suggested min gas price since: %v", fromTimestamp)
+			return err
```
tclemos, line 301: shouldn't return this error to be able to updated to default later

ARR552 (changes requested):

pool/pool.go
Psykepro: omg yep I tried to switch them to see something and forgot to fix it (facepalm)

tclemos (approved):

agnusmor (approved):

ARR552 (approved):

PR #1921, opened 2023-03-23
Reprocess empty batch to update state root before sanity check (full batch)

agnusmor:


sequencer/finalizer.go
```diff
@@ -281,6 +281,15 @@ func (f *finalizer) newWIPBatch(ctx context.Context) (*WipBatch, error) {
 		return nil, errors.New("state root and local exit root must have value to close batch")
 	}
 
+	// We need to process the batch to update the state root before closing the batch
+	if f.batch.initialStateRoot == f.batch.stateRoot {
```
ToniRamirezM, line 285: We only need to do this if the batch has no transacgions

ToniRamirezM (changes requested):

sequencer/finalizer.go
agnusmor: but if initialStateRoot is equal to the stateRoot this means that the batch is empty (no txs executed), isn't it?

ToniRamirezM (approved):

PR #1945, opened 2023-03-28
trusted reorg if it is not a trusted sequencer

ARR552:
### What does this PR do?

Allow trusted reorg in non trusted sequencers

### Reviewers

- @arnaubennassar 
- @ToniRamirezM


arnaubennassar (approved):

synchronizer/synchronizer.go
```diff
@@ -529,7 +529,9 @@ func (s *ClientSynchronizer) checkTrustedState(batch state.Batch, tBatch *state.
 	if reorgReasons.Len() > 0 {
 		reason := reorgReasons.String()
 		log.Warnf("Trusted Reorg detected for Batch Number: %d. Reasons: %s", tBatch.BatchNumber, reason)
-		log.Fatal("TRUSTED REORG DETECTED! Batch: ", batch.BatchNumber)
+		if s.isTrustedSequencer {
```
ToniRamirezM, line 532: I think we should halt the sequencer here. It will keep restarting if we do this.

ToniRamirezM (approved):

PR #1975, opened 2023-03-31
mainnet and rollup exit roots

### LINKED ISSUE #1974: Add mainnetExitRoot and RollupExitRoot custom endpoint to get batch info

ARR552:
Closes #1974 
### What does this PR do?
adds mainnetexitroot and rollupexitroot to the custom endpoint
### Reviewers

Main reviewers:

- @tclemos 
- @cool-develope


ARR552:
> could you share more detail context of this change?

Sure. As you can see, the node was only returning the globalexitroot but, previously In  the broadcast the node was returning the mainnetexitroot and the rollupexitroot in the same format we receive the info from l1 to allow the bridge service to fill the exit_root table. This change adds these fields to keep this info in the bridge

C001-developer (approved):

PR #1989, opened 2023-04-04
event log

### LINKED ISSUE #1605: Improve debug table behaviour
- Add a label to understand what component is logging the error (sequencer, sync, ...)
- Encode byte arrays in hexa instead of base 64 (default conversion done by Go)
- capture the executor top level error, not only the request
- capture requests also when something unexpected happens even if there is not a top level error from the executor (for instance an error on the RLP or tx hash decoding)

### LINKED ISSUE #1832: Move debug and event tables out to the state schema

ToniRamirezM:
Closes #1605
Closes #1832 

### What does this PR do?

Creates an event log component.

### Reviewers

Main reviewers:

- @ARR552 
- @agnusmor 
- @tclemos 
- @xavier-romero

### DIFF
--- a/cmd/run.go
+++ b/cmd/run.go
@@ -135,18 +135,31 @@ func start(cliCtx *cli.Context) error {
 
 	etm := ethtxmanager.New(c.EthTxManager, etherman, ethTxManagerStorage, st)
 
+	ev := &event.Event{
+		ReceivedAt: time.Now(),
+		Source:     event.Source_Node,
+		Level:      event.Level_Info,
+		EventID:    event.EventID_NodeComponentStarted,
+	}
+
 	for _, component := range components {
 		switch component {
 		case AGGREGATOR:
-			log.Info("Running aggregator")
+			ev.Component = event.Component_Aggregator
+			ev.Description = "Running aggregator"
+			eventLog.LogEvent(ctx, ev)
 			go runAggregator(ctx, c.Aggregator, etherman, etm, st)
 		case SEQUENCER:
-			log.Info("Running sequencer")
+			ev.Component = event.Component_Sequencer
+			ev.Description = "Running sequencer"
+			eventLog.LogEvent(ctx, ev)
 			poolInstance := createPool(c.Pool, c.NetworkConfig.L2BridgeAddr, l2ChainID, st, eventLog)
 			seq := createSequencer(*c, poolInstance, ethTxManagerStorage, st, eventLog)
 			go seq.Start(ctx)
 		case RPC:
-			log.Info("Running JSON-RPC server")
+			ev.Component = event.Component_RPC
+			ev.Description = "Running JSON-RPC server"
+			eventLog.LogEvent(ctx, ev)
 			poolInstance := createPool(c.Pool, c.NetworkConfig.L2BridgeAddr, l2ChainID, st, eventLog)
 			if c.RPC.EnableL2SuggestedGasPricePolling {
 				// Needed for rejecting transactions with too low gas price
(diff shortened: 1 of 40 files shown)


cmd/run.go
```diff
@@ -72,6 +75,24 @@ func start(cliCtx *cli.Context) error {
 	}
 	checkStateMigrations(c.StateDB)
 
+	// Prepare event log
+	var eventLog *event.EventLog
+	var eventStorage event.Storage
+
+	if c.EventLog.DB.Name != "disabled" {
```
agnusmor, line 82: What happens if the EventLog.DB entry doesn't exists in the config file? if it doesn't exists we need to manage it in the same way if it have the "disabled" value. For non-trusted nodes (that will not use the event table) it will be more easy to avoid the creation of the [EventLog] [EventLog.DB] sections in the config file, it will be more transparent.

event/pgeventstorage/pgeventstorage.go
```diff
@@ -0,0 +1,61 @@
...
+func (p *PostgresEventStorage) Close() error {
+	p.db.Close()
+	return nil
+}
+
+// LogEvent logs an event to the database
+func (p *PostgresEventStorage) LogEvent(ctx context.Context, ev *event.Event) error {
+	const insertEventSQL = "INSERT INTO event (received_at, ip_address, source, component, level, event_id, description, data, json) VALUES ($1, $2, $3, $4, $5, $6, $7, $8, $9)"
+
+	switch ev.Level {
```
agnusmor, line 40: We have this same code in the LogEvent function to the NilEventStorage package, there is no way to have a common function for this or reuse the NilEventStorage.LogEvent function? this is to avoid duplicate code as maybe in the future we can miss update on of the 2 codes when doing a change related to this

cmd/run.go
ToniRamirezM: This disabled literal will be read from the default.go so it does not need to be in the config toml files

db/migrations/state/0005.sql
```diff
@@ -0,0 +1,20 @@
+-- +migrate Up
+DROP table state.event;
+DROP table state.debug;
+
+-- +migrate Down
+CREATE TABLE state.event
```
ARR552, line 6: I'd prefer to use `IF NOT EXIST` to avoid problems. I know this situation shouldn't happen in theory

event/nileventstorage/nileventstorage.go
```diff
@@ -0,0 +1,46 @@
...
+		log.Errorf("Event: %+v", ev)
+	case event.Level_Warning:
+		log.Warnf("Event: %+v", ev)
+	case event.Level_Notice:
+		log.Infof("Event: %+v", ev)
+	case event.Level_Info:
+		log.Infof("Event: %+v", ev)
+	case event.Level_Debug:
+		log.Debugf("Event: %+v", ev)
+	}
```
ARR552, line 45:
```suggestion
	switch ev.Level {
	case event.Level_Emergency, event.Level_Alert, event.Level_Critical, event.Level_Error:
		log.Errorf("Event: %+v", ev)
	case event.Level_Warning:
		log.Warnf("Event: %+v", ev)
	case event.Level_Notice, event.Level_Info::
		log.Infof("Event: %+v", ev)
	case event.Level_Debug:
		log.Debugf("Event: %+v", ev)
	}
```

agnusmor (approved):

ARR552 (approved):

agnusmor (approved):

PR #1992, opened 2023-04-05
improve: storing failed rx reason in db.

### LINKED ISSUE #1754: Add failed reason to pool

Psykepro:
Closes #1754

### What does this PR do?

This pull request addresses the issue of providing more context for failed and invalid transactions in the `pool.transaction` table. The changes introduced in this PR are:

A new migration file, `0007.sql`, has been created in the `db/pool` directory. This migration adds a `varchar` column called `failed_reason` to the `pool.transaction` table.

The `TxTracker` structure has been updated to include a `FailedReason` attribute. This enables attaching the cause of failure once identified and retrieving the reason before saving the entry in the database.

The `UpdateTxStatus` method of `DBManager` has been modified to include a new parameter, `failedReason`, of type `*string` (a pointer, since it's only expected for `failed` and `invalid` status). The updated method signature is: `UpdateTxStatus(ctx context.Context, hash common.Hash, newStatus pool.TxStatus, isWIP bool, failedReason *string) error.` The new `failedReason` value is also saved in the `failed_reason` column in the database.

The `FailedReason` attribute is now passed on the `TxTracker` field when known.

And updated unit test when storing transactions with `invalid` status.

### Reviewers

Main reviewers:
- @ToniRamirezM 
- @ARR552 
- @tclemos


sequencer/finalizer.go
```diff
@@ -456,19 +456,23 @@ func (f *finalizer) storeProcessedTx(ctx context.Context, previousL2BlockStateRo
 }
 
 // handleTransactionError handles the error of a transaction
-func (f *finalizer) handleTransactionError(ctx context.Context, result *state.ProcessBatchResponse, tx *TxTracker) {
+func (f *finalizer) handleTransactionError(ctx context.Context, result *state.ProcessBatchResponse, tx *TxTracker) *sync.WaitGroup {
```
agnusmor, line 459: We are returning the WaitGroup but I don't see any play where we are using the WaitGroup, is it ok?

Psykepro: It is for the unit tests in `sequencer/finalizer_test.go` line 1147, to be testable. Otherwise, the next test fails if we don't wait for the queries when we do assert expectations.

db/migrations/pool/0008.sql
```diff
@@ -0,0 +1,5 @@
+-- +migrate Up
+ALTER TABLE pool.transaction ADD COLUMN failed_reason varchar;
```
ToniRamirezM, line 2:
```suggestion
ALTER TABLE pool.transaction ADD COLUMN failed_reason VARCHAR;
```

db/migrations/pool/0008.sql
```diff
@@ -0,0 +1,5 @@
...
+
+-- +migrate Down
+ALTER TABLE pool.transaction DROP COLUMN failed_reason;
```
ToniRamirezM, line 5:
```suggestion
ALTER TABLE pool.transaction DROP COLUMN failed_reason;

```

ARR552 (approved):

ToniRamirezM (approved):

PR #2007, opened 2023-04-12
Improve timestamp resolution

### LINKED ISSUE #2004: Limit timestamp resolution
# Rationale

Some dapps base its logic in the timestamp and we should update it at a max interval to be defined.

# Implementation

When a new TX arrives if the batch timestamp is > than the max allowed, we must close the batch, create a new one and add this new TX to the new batch.

If it is the first tx in the batch, update the batch's timestamp to time.now()

Use a new config parameter: timestampResolution

Remove current parameter to close a batch due to a timeout.
Take into account this timestamp for gas estimation and eth_call: use time.Now() if block is not forced.

ToniRamirezM:
Closes #2004.

### What does this PR do?

Improves timestamp resolution

### Reviewers

Main reviewers:

- @arnaubennassar 
- @agnusmor 
- @tclemos

### DIFF
--- a/state/transaction.go
+++ b/state/transaction.go
@@ -609,13 +609,15 @@ func (s *State) internalProcessUnsignedTransaction(ctx context.Context, tx *type
 
 	timestamp := uint64(lastBatch.Timestamp.Unix())
 
-	latestL2BlockNumber, err := s.PostgresStorage.GetLastL2BlockNumber(ctx, dbTx)
-	if err != nil {
-		return nil, err
-	}
+	if l2BlockNumber != nil {
+		latestL2BlockNumber, err := s.PostgresStorage.GetLastL2BlockNumber(ctx, dbTx)
+		if err != nil {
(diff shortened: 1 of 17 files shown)


state/transaction.go
```diff
@@ -607,6 +607,17 @@ func (s *State) internalProcessUnsignedTransaction(ctx context.Context, tx *type
 		previousBatch = lastBatches[1]
 	}
 
+	timestamp := uint64(lastBatch.Timestamp.Unix())
+
+	latestL2BlockNumber, err := s.PostgresStorage.GetLastL2BlockNumber(ctx, dbTx)
```
fsegoviam, line 612: This must be called only if l2BlockNumer != nil, the we can put this function call and the "if" of the line 617 inside an "if l2BlockNumber != nil" block (and remove the "l2BlockNumer != nil" from line 617)

fsegoviam (approved):

PR #2028, opened 2023-04-19
fix wrong state root when prevlastbatch & lastbatch are empty + restart

ARR552:
### What does this PR do?

This PR fixes a bug that consist in getting the wrong state root when the last two batches are empty (no txs = no l2blocks) and happens a reset of the sequencer. In this situation the sequencer was using the third to last batch stateroot.
image
The sequencer was getting the stateRoot `0xc4c.....` for the batch 1236 as oldStateRoot instead of the `0x35.....` when it was executing tx by tx. However when it executed the whole batch it was using the correct one (`0x35...`).

### Reviewers

- @agnusmor 
- @ToniRamirezM

### DIFF
--- a/sequencer/dbmanager.go
+++ b/sequencer/dbmanager.go
@@ -271,8 +271,8 @@ func (d *dbManager) GetWIPBatch(ctx context.Context) (*WipBatch, error) {
 	}
 
 	var lastStateRoot common.Hash
-	// If the last two batches has no txs, the stateRoot can not be retrieve from the l2block because there is no tx.
-	// In this case, the stateRoot must be get from the previousLastBatch
+	// If the last two batches have no txs, the stateRoot can not be retrieved from the l2block because there is no tx.
+	// In this case, the stateRoot must be gotten from the previousLastBatch
 	if len(lastBatchTxs) == 0 && previousLastBatch != nil && len(prevLastBatchTxs) == 0 {
 		lastStateRoot = previousLastBatch.StateRoot
 	} else {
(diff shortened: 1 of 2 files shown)


agnusmor (approved):

sequencer/dbmanager.go
```diff
@@ -260,12 +260,26 @@ func (d *dbManager) GetWIPBatch(ctx context.Context) (*WipBatch, error) {
...
+	var prevLastBatchTxs []types.Transaction
+	if previousLastBatch != nil {
+		prevLastBatchTxs, _, err = state.DecodeTxs(previousLastBatch.BatchL2Data)
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	var lastStateRoot common.Hash
+	// If the last two batches has no txs, the stateRoot can not be retrieve from the l2block because there is no tx.
```
ToniRamirezM, line 274:
```suggestion
	// If the last two batches have no txs, the stateRoot cannot be retrieved from the l2block because there are no txs.
```

sequencer/dbmanager.go
```diff
@@ -260,12 +260,26 @@ func (d *dbManager) GetWIPBatch(ctx context.Context) (*WipBatch, error) {
...
+	// In this case, the stateRoot must be get from the previousLastBatch
```
ToniRamirezM, line 275:
```suggestion
	// In this case, the stateRoot must be gotten from the previousLastBatch
```

ToniRamirezM (approved):

PR #2031, opened 2023-04-20
Improve GHA

arnaubennassar:
- Remove prover mock
- Increase deadline for E2E tests


test/operations/wait.go
```diff
@@ -30,7 +30,7 @@ const (
 	// DefaultInterval is a time interval
 	DefaultInterval = 2 * time.Millisecond
 	// DefaultDeadline is a time interval
-	DefaultDeadline = 30 * time.Second
```
ToniRamirezM, line 33: IMHO the problems with the tests are not related to a timeout, but to some components (usually the executor do not start) or take too much time. As far as I know this deadline does not affect that.

arnaubennassar: I'm not sure if this is the thing that makes the timeout happen, but could be. In just 30s a lot of things need to happen:
- Pull the fake L1 docker image and start it (unreliable amount of time)
- Send and verify a tx with the L1
- Start another ~5 containers 
- Wait for the jsonRPC to be responsive

30s seems a bit tight to me. Also, after this change all actions completed at firs try. It was a long time since this happened to me, but of course it could have been coinicidence

ToniRamirezM (approved):

PR #2051, opened 2023-04-26
New sequence sender component

### LINKED ISSUE #2046: Separate sequence sender from the sequencer as a new component

ToniRamirezM:
Closes #2046

### Reviewers

Main reviewers:

- @tclemos 
- @Psykepro 
- @agnusmor


ToniRamirezM:
@cla-bot check

Psykepro (approved):

cmd/run.go
```diff
@@ -146,6 +147,8 @@ func start(cliCtx *cli.Context) error {
 		EventID:    event.EventID_NodeComponentStarted,
 	}
 
+	var poolInstance *pool.Pool
+
```
tclemos, line 151: This change would force all the components running at the same binary to concur the db access. Multiple components will run different go routines but will consume a single pool Instance. 

it's ok to initialize multiple pool instances for different components and the time it takes do instantiate shouldn't be a problem, I don't see a reason for this optimization and I'm afraid of the side effects it can cause when talking about concurrent and parallel executions to the pool.

What is the rational behind this change? Have you faced an issue during the development?

config/default.go
```diff
@@ -102,15 +100,21 @@ MaxTxSizeForL1 = 131072
 		ClosingSignalsManagerWaitForCheckingGER = "10s"
 		ClosingSignalsManagerWaitForCheckingForcedBatches = "10s"
 		ForcedBatchesFinalityNumberOfBlocks = 64
-		TimestampResolution = "15s"
-		SenderAddress = "0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266"
+		TimestampResolution = "15s"	
 		PrivateKeys = [{Path = "/pk/sequencer.keystore", Password = "testonly"}]
```
tclemos, line 104: Shouldn't this line be removed? 
Even more, shouldn't we remove from the default config, values like addresses? We can keep them in the environment files.

tclemos (approved):
I'm approving this PR because I don't see any blocker and I we want to add this feature in the next release.
The shared pool instance can create some side effects, please make sure we want to have this change.
The change to the default config makes sense to me, not sure if it makes sense to fix it now, if not, I recommend opening an issue to review this later.

cmd/run.go
ToniRamirezM: I just thought it was better to just have one pool per binary. I can undo this is you don't feel confortable, or we are not sure if it may cause problems.

tclemos: Single pool, maybe better performance, maybe side cases.
Multiple pools, work as always, and already tests multiple times.

It's up to you, my suggestion is that we shouldn't touch on a team that's winning unless you want more emotion in your life 🤣

arnaubennassar (changes requested):
txpool as a component is unused, you're using the error types but not the pool instance itself, note that:

```go
type SequenceSender struct {
	cfg          Config
	pool         txPool
// ...
```

However in the code itself the `pool` (component) is not used, instead the package is used (`txpool`):

```go
		if err == nil && tx.Size() > s.cfg.MaxTxSizeForL1 {
			metrics.SequencesOvesizedDataError()
			log.Infof("oversized Data on TX oldHash %s (txSize %d > 128KB)", tx.Hash(), tx.Size())
			err = txpool.ErrOversizedData
		}
```

and

```go
func isDataForEthTxTooBig(err error) bool {
	return errors.Is(err, ethman.ErrGasRequiredExceedsAllowance) ||
		errors.Is(err, txpool.ErrOversizedData) ||
		errors.Is(err, ethman.ErrContentLengthTooLarge)
}
```
Apart from that, the need for the `txpool` itself is artificial, you're using a error type from that package that could be defined in this package, removing the dependency: `err = txpool.ErrOversizedData`

PR #2052, opened 2023-04-26
fix default trace values for custom tracers

tclemos:
### What does this PR do?

fix default trace values for custom tracers, because they need the full trace in order to generate the custom tracer response.

### Reviewers

Main reviewers:

@ToniRamirezM


tclemos:
closed due to #2066

PR #2058, opened 2023-04-28
Make PostgreSQL version persistent to 14

bimlas:
For new installations, the PostgreSQL version would be >14 and it breaks running migrations: the `init_prover_db.sql` runs successfully but when PostgreSQL restarts after this, it shows the following errors:

```
zkevm-state-db: ERROR:  permission denied for schema public
zkevm-state-db: STATEMENT:  create or replace function get_tree (root_hash bytea, remaining_key bytea)
```

Making PostgreSQL version fixed to 14 solves the problem.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -57,7 +57,7 @@ services:
   zkevm-state-db:
     container_name: zkevm-state-db
     restart: unless-stopped
-    image: postgres
+    image: postgres:14
     healthcheck:
       test: [ "CMD-SHELL", "pg_isready" ]
       interval: 10s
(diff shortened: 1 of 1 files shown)


ToniRamirezM:
@bimlas we tried to fix this issue with the latest version in the develop branch. It will soon arrive to a new release.

I don't know if you could confirm our version fixes this issue you describe.

ToniRamirezM:
As I think we already fixed this, I will wait a few days for @bimlas to answer before closing this PR.

ToniRamirezM:
As there are no updates. I am closing this PR without merging it.

PR #2091, opened 2023-05-10
Fix decode txs

### LINKED ISSUE #2008: Reported bugs in DecodeTxs
Reported by Niv Yehezkel and Dean Pierce:

[...] The bug is in DecodeTxs: https://github.com/0xPolygonHermez/zkevm-node/blob/c55a7984f7d5fb1e7bb2db6604f5e01aa857a9c6/state/helper.go#L158 where the byte array indices can go out of bounds and then crash the process

When the sequencer reads from L1 “encoded” transactions (where theoretically anyone could submit those transactions through force batch) and it will attempt to decode them (as they are supposed to be RLP encoded)
Now, DecodeTxs maintains a len variable that denotes the length of the txs data and it checks that it is lesser than the tx data: https://github.com/0xPolygonHermez/zkevm-node/blob/c55a7984f7d5fb1e7bb2db6604f5e01aa857a9c6/state/helper.go#L154

Now, a few lines afterwards it uses len and adds to constant numbers when accessing the tx data: https://github.com/0xPolygonHermez/zkevm-node/blob/c55a7984f7d5fb1e7bb2db6604f5e01aa857a9c6/state/helper.go#L158
The idea is that you can get a len which is len(txsData)-1 so that line 154 passes successfully and then it will crash on line 158 

(it is also possible to crash the program also due to the integer overflow in here https://github.com/0xPolygonHermez/zkevm-node/blob/c55a7984f7d5fb1e7bb2db6604f5e01aa857a9c6/state/helper.go#L152)

ToniRamirezM:
Closes #2008

### What does this PR do?

Fixes reported bugs in decode txs function.

### Reviewers

Main reviewers:

- @tclemos 
- @ARR552 
- @Psykepro 
- @pierce403 
- @agnusmor

### DIFF
--- a/state/helper.go
+++ b/state/helper.go
@@ -114,54 +114,67 @@ func EncodeUnsignedTransaction(tx types.Transaction, chainID uint64, forcedNonce
 // DecodeTxs extracts Transactions for its encoded form
 func DecodeTxs(txsData []byte) ([]types.Transaction, []byte, error) {
 	// Process coded txs
-	var pos int64
+	var pos uint64
 	var txs []types.Transaction
 	const (
-		headerByteLength = 1
-		sLength          = 32
-		rLength          = 32
-		vLength          = 1
-		c0               = 192 // 192 is c0. This value is defined by the rlp protocol
-		ff               = 255 // max value of rlp header
-		shortRlp         = 55  // length of the short rlp codification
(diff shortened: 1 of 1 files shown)


Psykepro (approved):

ARR552 (approved):

pierce403 (approved):
looks good to me 👍

PR #2100, opened 2023-05-15
fix: fixing forced batches error handling and normal batches transaction level error handling.

### LINKED ISSUE #2099: Fix error handling in OOC for forced batches
finalize.go line 660

```
if err != nil || !response.IsBatchProcessed {
  f.halt(ctx, fmt.Errorf("failed to process forced batch, Executor err: %w", err))
  return lastBatchNumberInState, stateRoot
}
```

In case of OOC err is nill and it is unclear in the logs

Psykepro:
Closes #2099.

### What does this PR do?

Fixing:
- The full `error handling` for `Forced Batches`.
- The `transaction level error handling` for `Normal Batches`.
- The case for parsing`ErrIntrinsicInvalidBatchGasLimit` in `RomErrorCode` function of the state

### Reviewers

Main reviewers:
- @ToniRamirezM 
- @krlosMata 
- @ARR552 

Codeowner reviewers:
- @ToniRamirezM

### DIFF
--- a/sequencer/finalizer.go
+++ b/sequencer/finalizer.go
@@ -669,11 +669,8 @@ func (f *finalizer) processForcedBatch(ctx context.Context, lastBatchNumberInSta
 	f.nextGERMux.Lock()
 	f.lastGERHash = forcedBatch.GlobalExitRoot
 	f.nextGERMux.Unlock()
-	if len(response.Responses) > 0 {
-		romErr := executor.RomErrorCode(response.Responses[len(response.Responses)-1].RomError)
-		if !executor.IsROMOutOfCountersError(romErr) {
-			f.handleForcedTxsProcessResp(request, response, stateRoot)
-		}
+	if len(response.Responses) > 0 && !response.RomOOC {
+		f.handleForcedTxsProcessResp(request, response, stateRoot)
 	}
 
 	return lastBatchNumberInState, stateRoot
--- a/sequencer/finalizer.go
+++ b/sequencer/finalizer.go
@@ -294,7 +294,7 @@ func (f *finalizer) newWIPBatch(ctx context.Context) (*WipBatch, error) {
 
 	// Reprocess full batch as sanity check
 	processBatchResponse, err := f.reprocessFullBatch(ctx, f.batch.batchNumber, f.batch.stateRoot)
-	if err != nil || !processBatchResponse.IsBatchProcessed {
+	if err != nil || processBatchResponse.RomOOC {
 		log.Info("halting the finalizer because of a reprocessing error")
 		if err != nil {
 			f.halt(ctx, fmt.Errorf("failed to reprocess batch, err: %v", err))
@@ -400,7 +400,7 @@ func (f *finalizer) handleTxProcessResp(ctx context.Context, tx *TxTracker, resu
 	// Handle Transaction Error
 
 	errorCode := executor.RomErrorCode(result.Responses[0].RomError)
-	if !result.IsBatchProcessed || executor.IsIntrinsicError(errorCode) {
+	if result.RomOOC || executor.IsIntrinsicError(errorCode) {
 		// If intrinsic error or OOC error, we skip adding the transaction to the batch
 		f.handleTransactionError(ctx, result, tx)
 		return result.Responses[0].RomError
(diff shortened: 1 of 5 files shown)


ToniRamirezM (approved):

sequencer/finalizer.go
```diff
@@ -399,7 +398,8 @@ func (f *finalizer) processTransaction(ctx context.Context, tx *TxTracker) error
 // handleTxProcessResp handles the response of transaction processing.
 func (f *finalizer) handleTxProcessResp(ctx context.Context, tx *TxTracker, result *state.ProcessBatchResponse, oldStateRoot common.Hash) error {
 	// Handle Transaction Error
-	if result.Responses[0].RomError != nil && !errors.Is(result.Responses[0].RomError, runtime.ErrExecutionReverted) {
+	if !result.IsBatchProcessed {
+		// If intrinsic error or OOC error, we skip adding the transaction to the batch
 		f.handleTransactionError(ctx, result, tx)
 		return result.Responses[0].RomError
```
ARR552, line 404: Why returning thefirst response?

sequencer/finalizer.go
```diff
@@ -667,11 +676,7 @@ func (f *finalizer) processForcedBatch(ctx context.Context, lastBatchNumberInSta
 	f.nextGERMux.Lock()
 	f.lastGERHash = forcedBatch.GlobalExitRoot
 	f.nextGERMux.Unlock()
-
-	err = f.handleForcedBatchProcessResp(request, response, stateRoot)
-	if err != nil {
-		return lastBatchNumberInState, stateRoot
-	}
+	_ = f.handleForcedTxsProcessResp(request, response, stateRoot)
```
ARR552, line 679: why do we ignore the error?

sequencer/finalizer.go
```diff
@@ -418,22 +418,30 @@ func (f *finalizer) handleTxProcessResp(ctx context.Context, tx *TxTracker, resu
...
+func (f *finalizer) handleForcedTxsProcessResp(request state.ProcessRequest, result *state.ProcessBatchResponse, oldStateRoot common.Hash) error {
+	log.Infof("handleForcedTxsProcessResp: batchNumber: %d, oldStateRoot: %s, newStateRoot: %s", request.BatchNumber, oldStateRoot.String(), result.NewStateRoot.String())
 	for _, txResp := range result.Responses {
 		// Handle Transaction Error
-		if txResp.RomError != nil && !errors.Is(txResp.RomError, runtime.ErrExecutionReverted) {
-			errorCode := executor.RomErrorCode(txResp.RomError)
-			if executor.IsROMOutOfCountersError(errorCode) {
-				log.Warnf("handleForcedBatchProcessResp: ROM out of counters error: %s", txResp.RomError)
+		romErr := executor.RomErrorCode(txResp.RomError)
+		if txResp.RomError != nil {
```
ARR552, line 427:
```suggestion
		if txResp.RomError != nil {
		      romErr := executor.RomErrorCode(txResp.RomError)

```

sequencer/finalizer.go
```diff
@@ -418,22 +418,30 @@ func (f *finalizer) handleTxProcessResp(ctx context.Context, tx *TxTracker, resu
...
+			if executor.IsROMOutOfCountersError(romErr) {
+				// If we are out of counters, we should stop processing the batch
+				log.Errorf("handleForcedTxsProcessResp: ROM error: %s", txResp.RomError)
 				return txResp.RomError
+			} else if executor.IsIntrinsicError(romErr) {
+				// If we have an intrinsic error, we should continue processing the batch, but skip the transaction
+				if txResp.RomError != nil {
```
ARR552, line 434: I think this is redundant

ARR552 (changes requested):
What about the list of errors that increase the nonce??

sequencer/finalizer.go
Psykepro: it is not as if we skip adding the tx to the db if we have intrinsic error

sequencer/finalizer.go
Psykepro: because nothing happens even if we have error the I can just remove returning it

Psykepro: because nothing happens even if we have error I can just remove returning it

sequencer/finalizer.go
Psykepro: Here we process 1 by 1 we always can have only 1 response

ARR552 (approved):

krlosMata (approved):
:smile_cat:

PR #2137, opened 2023-05-29
Improve tx confirmation time for permissionless nodes

### LINKED ISSUE #2136: improve tx confirmation time for permissionless nodes

tclemos:
Closes #2136.

### What does this PR do?

Improve tx confirmation time for permissionless nodes

### Reviewers

Main reviewers:

@agnusmor 
@ToniRamirezM

### DIFF
--- a/jsonrpc/endpoints_eth.go
+++ b/jsonrpc/endpoints_eth.go
@@ -589,6 +589,24 @@ func (e *EthEndpoints) getTransactionByHashFromSequencerNode(hash common.Hash) (
 	return tx, nil
 }
 
+func (e *EthEndpoints) getTransactionReceiptFromSequencerNode(hash common.Hash) (interface{}, types.Error) {
+	res, err := client.JSONRPCCall(e.cfg.SequencerNodeURI, "eth_getTransactionReceipt", hash.String())
+	if err != nil {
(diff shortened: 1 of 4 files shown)


ToniRamirezM (approved):

arnaubennassar:
Talking with @tclemos about this PR I realised that introduces a problem, which IMO is too big to merge this PR:

- If a receipt is requested to a 3rd party node, and this node gets it from the sequencer node, the 3rd party node will have an inconsistent state in the eyes of the user

Let me explain with an example:

- Alice performs a token swap 2000 MATIC for 0.8 ~ 1.2 ETH, signs the tx and sends the tx to the node where she is connected, which happens to be a 3rd party node (3N for short)
- Later on Alice asks for the receipt to the 3N. 3N doesn't have the receipt so it asks for it to the sequencer node (SN). SN already has the receipt so: receipt => 3N => Alice
- After getting the receipt Alice performs a `eth_getBalance` call to 3N. Since the 3N has not synced yet the tx, it will return a value that doesn't have into consideration the state changes made by the tx that is supposed to be included in the state, since the receipt exists

Note that this example applies to many different endpoints including:
- call
- getCode
- block related endpoints
- ...

I'd say that it's quite a normal practice to do subsequent requests to the RPC once you receive a receipt. As a matter of fact we use this pattern in some tests in this repo

tclemos:
discarded

PR #2154, opened 2023-06-01
Include raw txs data into zkevm_getBatchByNumber

### LINKED ISSUE #2151: Include `raw_txs_data` on `zkevm_getBatchByNumber`
# Rationale

Batches have this piece of information. Some times the raw data can be reconstructed using transactions data, but this is not always the case. Keep in mind that the raw data can be anything, including random bytes that do not represent Ethereum transactions

What are the use-cases?

1. Simplify the implementation: by just returning data from a single table. The client can do the work of unpacking the raw data, saving time to access the DB on the server, and reducing the amount of data sent through the open internet
2. Be more accurate (as explained in the previous point)

# Implementation

Yes, just return all the data from the batch table

Yes

arnaubennassar:
Closes #2151

### DIFF
--- a/jsonrpc/endpoints_zkevm_test.go
+++ b/jsonrpc/endpoints_zkevm_test.go
@@ -661,48 +661,6 @@ func TestGetBatchByNumber(t *testing.T) {
 					Return(m.DbTx, nil).
 					Once()
 
-				batch := &state.Batch{
-					BatchNumber:    1,
-					Coinbase:       common.HexToAddress("0x1"),
-					StateRoot:      common.HexToHash("0x2"),
-					AccInputHash:   common.HexToHash("0x3"),
-					GlobalExitRoot: common.HexToHash("0x4"),
-					Timestamp:      time.Unix(1, 0),
-				}
-
-				m.State.
-					On("GetBatchByNumber", context.Background(), hex.DecodeBig(tc.Number).Uint64(), m.DbTx).
-					Return(batch, nil).
-					Once()
-
-				virtualBatch := &state.VirtualBatch{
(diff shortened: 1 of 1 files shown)


jsonrpc/types/types.go
```diff
@@ -337,10 +337,27 @@ type Batch struct {
...
+) (*Batch, error) {
+	batchL2Data := batch.BatchL2Data
+	if batchL2Data == nil {
+		batchL2dataFromTxs, err := state.EncodeTransactions(batch.Transactions)
+		if err != nil {
+			return nil, fmt.Errorf("error encoding txs into raw data: %w", err)
+		} else {
+			batchL2Data = batchL2dataFromTxs
+		}
+	}
```
ARR552, line 360: I think this situation is not possible. I mean if the batchL2Data is empty, it's impossible to have decoded transactions. And I'm not sure about this but wouldn't we get an error if the we try to decode an empty array of txs?

arnaubennassar: `batchL2Data` is only stored once the batch is closed. Therefore, while the batch is WIP there are txs but not batchL2Data.

> And I'm not sure about this but wouldn't we get an error if the we try to decode an empty array of txs?
Will try it

ARR552: Even when the batch is wip the sequencer stores the batchL2Data. You can check it in the state.Batch table
image

arnaubennassar: You're right, I had a misconception about this, glad you've clarified this to me. Fixed

ARR552 (approved):

tclemos (approved):

PR #2166, opened 2023-06-07
Debug trace fixes

ToniRamirezM:


ToniRamirezM:
Discarded as the fix will be done in the ROM

PR #2171, opened 2023-06-08
Feature/#2102 snap restore

### LINKED ISSUE #2102: Create snapshot and restore it
Create two cli commands. One that creates a snapshot of the nod and another one that restores the info

### LINKED ISSUE #2231: The synchronization from scratch is very slow. A mainnet snapshot image is needed.
# Rationale

Synchronizing from zero to the latest height takes a considerable amount of time. If there was a daily updated snapshot of the mainnet, it would reduce synchronization time and accelerate the deployment of local nodes.

# Implementation

Perhaps an AWS S3 would be sufficient.

ARR552:
Closes #2102 
Closes #2231

### What does this PR do?

Allow the ability to create archive snapshots and restore them for the node and the executor

#### Config changes:
Add this config:
```
[HashDB]
User = "prover_user"
Password = "prover_pass"
Name = "prover_db"
Host = "zkevm-state-db"
Port = "5432"
EnableLog = false
MaxConns = 200
```
### Reviewers

- @ToniRamirezM 
- @agnusmor


cmd/snapshot.go
```diff
@@ -0,0 +1,82 @@
...
+		DB:       c.StateDB.Name,
+		Username: c.StateDB.User,
+		Password: c.StateDB.Password,
+	})
+	if err != nil {
+		log.Error("error: ", err)
+		return err
+	}
+	log.Info("StateDB snapshot is being created...")
+	dump.SetFileName(fmt.Sprintf(`%v_%v_%v_%v.sql.tar.gz`, dump.DB, time.Now().Unix(), zkevm.Version, zkevm.GitRev))
```
joanestebanr, line 45: the result of executing a `pg_dump`is not a compressed tar file is:
 [Script dumps are plain-text files containing the SQL commands ](https://www.postgresql.org/docs/current/app-pgdump.html)

ARR552: As I can read this in the doc: 
`The most flexible output file formats are the “custom” format (-Fc) and the “directory” format (-Fd). They allow for selection and reordering of all archived items, support parallel restoration, and are compressed by default.`
By default the library uses custom format which should compress the dump. 
`pgDumpDefaultFormat = "c"`

ARR552: As far as I see it is compressed but not at the highest level.
`For the custom and directory archive formats, this specifies compression of individual table-data segments, and the default is to compress at a moderate level.`
I'm gonna try to force maximum compression

ARR552: @joanestebanr Now, it applies maximum compression (level 9). You can check it using command: `pg_restore -l <file>`

joanestebanr: Sorry, my fault, the format 'custom' have some plain text at begining! I checked your solution and with `-Z 9 `  reduce  80% size from the same DB with no compresion! (353Kb vs 81Kb)

ToniRamirezM (approved):

joanestebanr (approved):

PR #2176, opened 2023-06-09
Retry on resource exhausted 

### LINKED ISSUE #2156: Properly handle executor out of resources
Retry with timeout.

If timeout reaches return:

errcodeTimeout = -32002
errMsgTimeout = "request timed out"

ToniRamirezM:
Closes #2156

### What does this PR do?

Retries transaction execution in case of resource exhausted error from the executor. If the problem persists it is reported a as time out.

This PR creates 2 new config params:

[Executor]
MaxResourceExhaustedAttempts = 3
WaitOnResourceExhaustion = "1s"

### Reviewers

Main reviewers:

- @tclemos 
- @agnusmor 
- @ARR552

### DIFF
--- a/config/default.go
+++ b/config/default.go
@@ -139,7 +139,7 @@ URI = "zkevm-prover:50061"
 
 [Executor]
 URI = "zkevm-prover:50071"
-MaxResourceExhaustedAttempts = 3
+MaxResourceExhaustedAttempts = 1
 WaitOnResourceExhaustion = "1s"
 
 [Metrics]
--- a/config/default.go
+++ b/config/default.go
@@ -139,7 +139,7 @@ URI = "zkevm-prover:50061"
 
 [Executor]
 URI = "zkevm-prover:50071"
-MaxResourceExhaustedAttempts = 1
+MaxResourceExhaustedAttempts = 3
 WaitOnResourceExhaustion = "1s"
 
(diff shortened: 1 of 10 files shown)


ARR552 (approved):

config/default.go
```diff
@@ -139,6 +139,8 @@ URI = "zkevm-prover:50061"
 
 [Executor]
 URI = "zkevm-prover:50071"
+MaxResourceExhaustedAttempts = 1
```
agnusmor, line 142: The default value must be the value that we normally will use on testnet/mainnet. I see in the toml files that we are setting 3 attempts, therefore I understand that the default value must be also 3. Take in mind that one of the objectives is that the node, when no specifying a toml config file, must run with the default values for testnet/mainnet

ToniRamirezM: I am not sure about this case. I thing the default behaviour should be not retrying and so the value must be 1 by default, but I understand your point.

agnusmor: In any case I think the default values (default.go) must match the config values used in the default public testnet/mainnet toml files that we provide. But we can set the value to 1 in both places or 3. I think to do the retry (3) is fine as default behaviour

ToniRamirezM: Updated default to 3.

PR #2217, opened 2023-06-22
adds support to debug_traceBatchByNumber

### LINKED ISSUE #2078: debug_traceBatch... ByNumber and ByHash
# Rationale

since the network at the moment relies on a single transaction per block, `debug_traceTransaction` pretty much work as the same as `debug_traceBlockByNumber` or `trace_traceBlockByHash` 

but, we have the relation between batches and transactions, the idea is to provide a custom method to debug and trace a batch.

- debug_traceBatchByNumber

# Implementation

The new endpoint can be added to the jRPC, but the main change needs to be done on the `Executor` side, because the idea is to save time when `debugging` multiple transactions in a single request.

Since this process is heavy and can return a lot of data, we need to check the limits and discuss with the prover team which would be the best approach to achieve this.

tclemos:
Closes #2078

### What does this PR do?

adds support to debug_traceBatchByNumber

### Reviewers

Main reviewers:

@ToniRamirezM 
@agnusmor

### DIFF
--- a/jsonrpc/endpoints_debug.go
+++ b/jsonrpc/endpoints_debug.go
@@ -6,8 +6,13 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"net/http"
+	"net/url"
 	"strings"
+	"sync"
+	"time"
 
+	"github.com/0xPolygonHermez/zkevm-node/jsonrpc/client"
 	"github.com/0xPolygonHermez/zkevm-node/jsonrpc/types"
 	"github.com/0xPolygonHermez/zkevm-node/log"
 	"github.com/0xPolygonHermez/zkevm-node/state"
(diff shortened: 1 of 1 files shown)


jsonrpc/endpoints_debug.go
```diff
@@ -119,6 +129,131 @@ func (d *DebugEndpoints) TraceBlockByHash(hash types.ArgHash, cfg *traceConfig)
...
+	// the size of the buffer defines
+	// how many txs it will process in parallel.
+	const bufferSize = 10
+
+	// builds the url of the jRPC server from the data found in the httpRequest
+	scheme := "http"
+	if httpRequest.URL.Scheme != "" {
+		scheme = httpRequest.URL.Scheme
+	}
+	u := url.URL{
```
ToniRamirezM, line 163: We must ensure the URL we get here is not from behind load balancer.

ToniRamirezM (approved):

jsonrpc/endpoints_debug.go
tclemos: we have validated and it is the external URL

tclemos:
closed due to #2446

PR #2234, opened 2023-06-27
effective gas price comparation

ToniRamirezM:


sequencer/finalizer.go
```diff
@@ -510,41 +519,48 @@ func (f *finalizer) processTransaction(ctx context.Context, tx *TxTracker) (errW
 	if tx != nil {
 		f.processRequest.Transactions = tx.RawTx
 		hashStr = tx.HashStr
-		// Check if the guaranteed period for the break even gas price has passed
-		guaranteedPeriodDeadline := tx.ReceivedAt.Add(f.effectiveGasPriceCfg.BreakEvenGasPriceGuaranteedPeriod.Duration)
-		hasGuaranteedPeriodPassed := guaranteedPeriodDeadline.After(now())
 		breakEvenGasPrice := tx.BreakEvenGasPrice
-		if hasGuaranteedPeriodPassed || breakEvenGasPrice == nil || breakEvenGasPrice.Uint64() == 0 {
+		if breakEvenGasPrice.Uint64() == 0 {
```
agnusmor, line 523: it's not need to have the breakEvenGasPrice var we can use "if tx.BreakEvenGasPrice.Uint64()"

pool/pool.go
```diff
@@ -501,23 +525,21 @@ func (p *Pool) UpdateTxWIPStatus(ctx context.Context, hash common.Hash, isWIP bo
...
+	gasPrices := p.gasPrices
+	p.gasPricesMux.RUnlock()
+
+	// We enforce l1GasPrice to make it consistent during the lifespan of the transaction
+	gasPrices.L1GasPrice = l1GasPrice
+
 	if gasPrices.L1GasPrice == 0 {
 		log.Warn("Received L1 gas price 0. Skipping estimation...")
-		return nil, ErrReceivedZeroL1GasPrice
+		return big.NewInt(0), nil
```
agnusmor, line 538: I think we must return ErrReceivedZeroL1GasPrice instead of nil

agnusmor (approved):

Issue #2236, opened 2023-06-28
Support fork id upgrades without down time for the trusted state


agnusmor:
# Rationale

When doing a forkid upgrade, now it is needed to stop the RPC service to avoid generate new trusted batches and wait until the pending batches are virtualized and verified. When this is finished we do the forkid upgrade at SC level and then we resume the operation (new batches are generated using the new fork id). This implies to have the network down during a long time. The idea is to have a mechanism that stops to generate new virtual and verified batches while we wait that the pending batches are virtualized/verified, but without the need to stop the RPC service and continuing with the generation of trusted batches and confirming txs (thus the user don't have the "perception" that the network is stopped or it is being upgraded)

# Implementation

The idea is to add a configuration parameter to define a future batch (ForkUpgradeBatchNumber) that will be the batch that will use the new forkid. The sequencesender will continue working as usual until it reaches this batch, at this moment the sequence including this batch need to be closed and sent to L1 and the sequencesender will stop sequencing (virtualize) new batches. Also the sequencer/rpc will work with the current forkid until it reaches the ForkUpgradeBatchNumber, at this point rpc/sequencer will start to use the new forkid with the executor. When all the batches until ForkUpgradeBatchNumber are virtualized/verified, we will proceed to update the SC to set the new forkid (from this ForkUpgradeBatchNumber) and after this we will set ForkUpgradeBatchNumber=0 (to disable it) and restart the node to start to work as usual with the new SC forkid and also we will start Provers using new forkid to verfiy the new batches from ForkUpgradeBatchNumber

PR #2237, opened 2023-06-28
Support fork id upgrades without down time for the trusted state

agnusmor:
Closes [#2236](https://github.com/0xPolygonHermez/zkevm-node/issues/2236).

### What does this PR do?

Support fork id upgrades without down time for the trusted state.

Steps to perform a forkid upgrade using this feature (e.g upgrading from forkid 4 to 5):
1. Update node/executor to new forkid 5 versions. Keep provers running on previous version (previous forkid 4)
2. Set ForkUpgradeBatchNumber config parameter to the batch number that will be the last batch using the previous forkid 4 and set ForkUpgradeNewForkId to the new forkid value (5). 
3. Start/restart node components to get the new config values 
4. Wait until ForkUpgradeBatchNumber batch has been virtualized and verfied. While waiting for this the system we will continue processing txs and batches and adding it to the trusted state (but virtualization and verification will be on hold from ForkUpgradeNewForkId+1 batch)
5. Update SC to the new forkid 5
6. Stop provers using previous forkid 4
7. Set ForkUpgradeNewForkId = 0 and ForkUpgradeNewForkId = 0 to disable this feature
8. Restart node components (RPC, Synchornizer, Sequencer, SequenceSender, Aggregator) to get the config changes
9. Update provers to new forkid 5 versions
10. Check the service resumes correctly and new batches are virtualized and verified using new forkid 5

New config parameters in the node (at root level of the toml file):
ForkUpgradeBatchNumber = 0
ForkUpgradeNewForkId = 0

### Reviewers

Main reviewers:
@ToniRamirezM 
@ARR552

### DIFF
--- a/aggregator/config.go
+++ b/aggregator/config.go
@@ -60,7 +60,7 @@ type Config struct {
 	ChainID uint64
 
 	// ForkID is the L2 ForkID provided by the Network Config
-	ForkId uint64 `mapstructure:"ForkId"`
+	ForkId uint64
 
 	// SenderAddress defines which private key the eth tx manager needs to use
 	// to sign the L1 txs
--- a/test/config/debug.node.config.toml
+++ b/test/config/debug.node.config.toml
@@ -105,7 +105,7 @@ PrivateKeys = [{Path = "./test/sequencer.keystore", Password = "testonly"}]
 
 [Aggregator]
 Host = "0.0.0.0"
-Port = 500
+Port = 50081
 RetryTime = "5s"
 VerifyProofInterval = "30s"
 TxProfitabilityCheckerType = "acceptall"
--- a/test/config/debug.node.config.toml
+++ b/test/config/debug.node.config.toml
@@ -105,8 +105,7 @@ PrivateKeys = [{Path = "./test/sequencer.keystore", Password = "testonly"}]
 
 [Aggregator]
 Host = "0.0.0.0"
-Port = 50081
-ForkId = 2
+Port = 500
 RetryTime = "5s"
 VerifyProofInterval = "30s"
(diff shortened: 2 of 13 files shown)


test/config/debug.node.config.toml
```diff
@@ -105,8 +105,7 @@ PrivateKeys = [{Path = "./test/sequencer.keystore", Password = "testonly"}]
 
 [Aggregator]
 Host = "0.0.0.0"
-Port = 50081
-ForkId = 2
+Port = 500
```
ARR552, line 108: Why this change? I think it is a typo error

aggregator/config.go
```diff
@@ -60,7 +60,7 @@ type Config struct {
 	ChainID uint64
 
 	// ForkID is the L2 ForkID provided by the Network Config
-	ForkId uint64 `mapstructure:"ForkId"`
+	ForkId uint64
```
ARR552, line 63: How do you assign this variable? If it is not needed you can delete it

test/config/debug.node.config.toml
agnusmor: Yes, is a typo, only Forkid must be deleted

aggregator/config.go
agnusmor: It is assigned in run.go, after read the forkid intervals from L1 (ChanID is also assigned in this way):
https://github.com/0xPolygonHermez/zkevm-node/blob/b801f527572adb919ebf3d99be8bc603ddc24a70/cmd/run.go#L132

ToniRamirezM (approved):

ARR552 (approved):

PR #2270, opened 2023-07-10
Fix pending tx when duplicate nonce

agnusmor:
### What does this PR do?

Fixes a problem when adding a tx to an AddrQueue and a tx with the same nonce already exists for this address. In this case the tx with best gasPrice is kept in memory (AddrQueue) and the other is "discarded". The problem is that this discarded tx is not updated in the db as "failed". This causes to have tx in the pool with "pending" state forever and never are processed

### Reviewers

Main reviewers:
- @Psykepro 
- @ARR552

### DIFF
--- a/sequencer/addrqueue.go
+++ b/sequencer/addrqueue.go
@@ -32,31 +32,48 @@ func newAddrQueue(addr common.Address, nonce uint64, balance *big.Int) *addrQueu
 	}
 }
 
-// addTx adds a tx to the addrQueue and updates the ready a notReady Txs
-func (a *addrQueue) addTx(tx *TxTracker) (newReadyTx, prevReadyTx *TxTracker, dropReason error) {
+// addTx adds a tx to the addrQueue and updates the ready a notReady Txs. Also if this tx matches
+// with an existing tx with the same nonce, we return in the replacedTx the one with lower gasPrice.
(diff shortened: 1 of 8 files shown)


Psykepro (approved):

sequencer/addrqueue.go
```diff
@@ -32,31 +32,48 @@ func newAddrQueue(addr common.Address, nonce uint64, balance *big.Int) *addrQueu
...
+				return tx, oldReadyTx, disTx, nil
 			} else { // If there is not enough balance we set the new tx as notReadyTxs
 				a.readyTx = nil
 				a.notReadyTxs[tx.Nonce] = tx
-				return nil, oldReadyTx, nil
+				return nil, oldReadyTx, disTx, nil
 			}
+		} else { // We have an already readytx with the same nonce and better gas price, we discard the new tx
+			disTx = tx
+			return nil, nil, disTx, nil
```
ARR552, line 59: Shouldn't we return an error here?? I mean, a reason for the discarded tx

agnusmor: No, we specify the reason in the parent function (check dbManager.addTxToWorker) that is calling this function, as he knows that is a "discarded" tx and it store as reason "duplicated nonce"

PR #2285, opened 2023-07-14
avoid double unlock in sequencer AddTxTracker

icydark:
### What does this PR do?

Fix double unlock in worker.go AddTxTracker

Line 49 -- lock 1st
Line 56 -- unlock 1st
Line 62/68/74 return --> defer unlock 2nd

Result: leads to a fater error

Fix: do not use defer, manually unlock before each return

### Reviewers

Main reviewers:

- @ToniRamirezM 
- @ARR552


Psykepro (approved):

ToniRamirezM (approved):

Psykepro (dismissed):
Fix conflicts and wait for the GitHub actions. After that will approve.

PR #2298, opened 2023-07-18
Add IP and ZKCounter Validation, Refactor Batch Configs

Psykepro:
Closes #2284.

### What does this PR do?

This pull request introduces some crucial enhancements aimed at optimizing our system's resource management and ensuring a more consistent operation.

#### Changes

1. **Transaction Pool IP Verification:**
    - We've incorporated an IP verification process in the `tryToAddTxToPool` function. This new feature checks the validity of IP addresses obtained from the `X-Forwarded-For` header.

2. **Worker's IP Verification:**
    - To further solidify the validity of IP addresses, we've added IP verification to the `Worker.AddTxTracker`. This helps ensure that all transactions are associated with valid IP addresses before being added to the efficiency list.

3. **BatchConfig Integration and Verification:**
    - The `BatchConstraints` and `BatchResourceWeights` configuration values have been streamlined into a unified `BatchConfig` structure. The `JSON-RPC` now utilizes this to check if a transaction stays within the predefined resource constraints. Transactions surpassing these constraints will be prevented from entering the transaction pool. Similarly, the `Sequencer` will sidestep such transactions, ensuring they are not added to the Worker's efficiency list.

    - The `State`, `StateDB` and `Batch` Config Parameters now will look like that:
    ```toml
    [State]
	    [State.DB]
	    User = "state_user"
	    Password = "state_password"
	    Name = "state_db"
	    Host = "zkevm-state-db"
	    Port = "5432"
	    EnableLog = false	
	    MaxConns = 200
	    [State.Batch]
		    [State.Batch.Constraints]
		    MaxTxsPerBatch = 300
		    MaxBatchBytesSize = 120000
		    MaxCumulativeGasUsed = 30000000
		    MaxKeccakHashes = 2145
		    MaxPoseidonHashes = 252357
		    MaxPoseidonPaddings = 135191
		    MaxMemAligns = 236585
		    MaxArithmetics = 236585
		    MaxBinaries = 473170
		    MaxSteps = 7570538
		    [State.Batch.ResourceWeights]
		    WeightBatchBytesSize = 1
		    WeightCumulativeGasUsed = 1
		    WeightKeccakHashes = 1
		    WeightPoseidonHashes = 1
		    WeightPoseidonPaddings = 1
		    WeightMemAligns = 1
		    WeightArithmetics = 1
		    WeightBinaries = 1
		    WeightSteps = 1
    ```
### Reviewers
- @ToniRamirezM 
- @agnusmor 
- @ARR552 
- @tclemos 

[...]

PR #2304, opened 2023-07-19
Add WebSocket Read Limit

### LINKED ISSUE #2293: Restrict Incoming Data Length on WebSocket Connection
# Summary:

There's an opportunity to improve the handling of incoming data size in our WebSocket server. The current setup does not put constraints on the volume of data being received.

## Detailed Description:

The configuration in our WebSocket server, specifically within `jsonrpc/server.go`, lacks controls on the amount of incoming data. This situation might allow external sources to maintain a continuous stream of data that could potentially put a strain on our server resources.

Our current implementation of the gorilla WebSocket library does not inherently impose a read limit, adding to this situation.

## Steps to implement:

1. **Establish a Read Limit on the WebSocket Server:** Make use of the `SetReadLimit()` function available within the gorilla WebSocket library. This can help us limit the volume of incoming data.

## Objective:

Implementing these changes would enable us to manage the incoming data more effectively, enhancing the efficiency of our system.

Psykepro:
Closes #2293.

### What does this PR accomplish?

This PR enhances our WebSocket server's robustness by introducing a control mechanism for the size of incoming data. It ensures that our server continues to operate optimally even under high data traffic.

Implementing a read limit on incoming WebSocket communications will offer an extra layer of protection, ensuring smooth server operations. This PR consists of:

1. Invoke the `SetReadLimit()` method right after an HTTP connection has been upgraded to a WebSocket connection.
2. Incorporating a new `ReadLimit` parameter into the WebSocket server configuration.

Here's a quick overview of the changes:

### Server Configuration Update

A new parameter `ReadLimit` has been introduced under the `[RPC.WebSockets]` configuration:

```toml
[RPC.WebSockets]
    Enabled = true
    Port = 8133
    ReadLimit = 104857600
```
This `100 MiB` value is based on Logs metrics for the maximum `body_bytes_size` sent to the `JSON-RPC`, which is `75.24 MiB`.

### Reviewers
- @ToniRamirezM 
- @agnusmor 

Codeowner reviewers:
- @tclemos


test/config/debug.node.config.toml
```diff
@@ -54,6 +54,7 @@ EnableL2SuggestedGasPricePolling = true
 	[RPC.WebSockets]
 		Enabled = true
 		Port = 8133
+		ReadLimit = 1048576
```
tclemos: this change is not needed, since it is the same value as the default.

test/config/test.node.config.toml
```diff
@@ -53,6 +53,7 @@ EnableL2SuggestedGasPricePolling = true
 	[RPC.WebSockets]
 		Enabled = true
 		Port = 8133
+		ReadLimit = 1048576
```
tclemos: this change is not needed, since it is the same value as the default.

config/environments/local/local.node.config.toml
```diff
@@ -52,6 +52,7 @@ EnableL2SuggestedGasPricePolling = true
 	[RPC.WebSockets]
 		Enabled = true
 		Port = 8546
+		ReadLimit = 1048576
```
tclemos: this change is not needed, since it is the same value as the default.

config/default.go
```diff
@@ -67,6 +67,7 @@ TraceBatchUseHTTPS = true
 		Enabled = true
 		Host = "0.0.0.0"
 		Port = 8546
+		ReadLimit = 1048576
```
tclemos: - a test to the default config should be added.

tclemos (changes requested):
- We must add tests to make sure we react properly when the read limit is exceeded.
- We must add a new test case for the new default config value.
- Is there an explanation for this value to be `1048576`? We have endpoints that allows the user to send a legit message bigger than this, like `eth_estimateGas` and `eth_call`, I'm not sure if this is a too-tight value, and since the point is to avoid it to be infinite, maybe we could have a bigger value to make sure it is safe but also to avoid unnecessary friction.
- Config example should contain only the new parameter and all the config levels needed to set it, like this:
```toml
[RPC]
	[RPC.WebSockets]
		ReadLimit = 1048576 
```
- also it's important to mention what happens if the parameter is set to `0`, and test it.

test/config/debug.node.config.toml
Psykepro: Actually, there is no default Read limit. If we don't set it everyone can DDOS the node. But we definitely must set the most reasonable limit.

Psykepro:
> * We must add tests to make sure we react properly when the read limit is exceeded.
> * We must add a new test case for the new default config value.
> * Is there an explanation for this value to be `1048576`? We have endpoints that allows the user to send a legit message bigger than this, like `eth_estimateGas` and `eth_call`, I'm not sure if this is a too-tight value, and since the point is to avoid it to be infinite, maybe we could have a bigger value to make sure it is safe but also to avoid unnecessary friction.
> * Config example should contain only the new parameter and all the config levels needed to set it, like this:
> 
> ```toml
> [RPC]
> 	[RPC.WebSockets]
> 		ReadLimit = 1048576 
> ```
> 
> * also it's important to mention what happens if the parameter is set to `0`, and test it.

I tested it locally and it reacts properly. But will create Unit test for that no probs. Apart from the for the value I set (1 MB) It's maybe not the best. But why would we need more than 1MB when we have this constraint for the whole batch:
```toml
MaxBatchBytesSize = 120000
````
If someone tries to `eth_estimateGas` for more bytes than 1 MB (which fits almost 10 batches) just for 1 tx this transaction is definitely invalid. Right? Maybe also we can change to use multiplied by 10 the `MaxBatchBytesSize` config value which will be shared config struct from this PR https://github.com/0xPolygonHermez/zkevm-node/pull/2298

config/default.go
tclemos: I've added the test to the PR

tclemos:
> ```toml
> MaxBatchBytesSize = 120000
> ```
> 
> If someone tries to `eth_estimateGas` for more bytes than 1 MB (which fits almost 10 batches) just for 1 tx this transaction is definitely invalid. Right? Maybe also we can change to use multiplied by 10 the `MaxBatchBytesSize` config value which will be shared config struct from this PR #2298

this makes sense for `eth_estimateGas`, but `eth_call` allow unsigned txs to be executed to get value via read only operations, these operations can have a huge payload.

let's talk with @xavier-romero to see if we can get some metrics from `Mainnet` and `Testnet` to get a better value to avoid causing unnecessary friction.

Probably we can get the average and max size for requests sent to `eth_call` so far to both networks and see if we can define a better number.

Also, the problem at the moment is that we don't have a limit, so the user can open the socket and start sending bytes, this makes the server accumulate these bytes in the memory for this connection until the user defines the message is ready to be read by the server, but with the read limitation, as soon as the message reaches the read limit, the connection is closed and the memory will be free as soon as the GC collects it, theoretically we can support a bigger number safely.

tclemos (approved):

ToniRamirezM (approved):

Psykepro:
@tclemos 
> > ```toml
> > MaxBatchBytesSize = 120000
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > If someone tries to `eth_estimateGas` for more bytes than 1 MB (which fits almost 10 batches) just for 1 tx this transaction is definitely invalid. Right? Maybe also we can change to use multiplied by 10 the `MaxBatchBytesSize` config value which will be shared config struct from this PR #2298
> 
> this makes sense for `eth_estimateGas`, but `eth_call` allow unsigned txs to be executed to get value via read only operations, these operations can have a huge payload.
> 
> let's talk with @xavier-romero to see if we can get some metrics from `Mainnet` and `Testnet` to get a better value to avoid causing unnecessary friction.
> 
> Probably we can get the average and max size for requests sent to `eth_call` so far to both networks and see if we can define a better number.
> 
> Also, the problem at the moment is that we don't have a limit, so the user can open the socket and start sending bytes, this makes the server accumulate these bytes in the memory for this connection until the user defines the message is ready to be read by the server, but with the read limitation, as soon as the message reaches the read limit, the connection is closed and the memory will be free as soon as the GC collects it, theoretically we can support a bigger number safely.

I've updated the value to `100 MiB` based on the Logs metrics for the maximum `body_bytes_size` sent to the `JSON-RPC`, which is `75.24 MiB`

PR #2306, opened 2023-07-19
Feature/2300 synchronizer detect if executor restart

### LINKED ISSUE #2300: make synchronizer robust against executor restart
# Rationale

`Synchronizer` must detect that `executor` have restarted because the flush-id is in memory and start again from 0

#### Actual behaviour
`Synchronizer` doesn't known that and is using a wrong flush-id 

# Implementation

The idea is try to recover the situation, if possible or restart also the `synchronizer`

joanestebanr:
Closes #2300
### What does this PR do?

It checks proverId (running Id for executor) to detect if executor have been restarted, if it happens it'll  stop synchonizer and send an event

### Reviewers

Main reviewers:

- @ToniRamirezM


synchronizer/synchronizer.go
```diff
@@ -47,6 +49,11 @@ type ClientSynchronizer struct {
 		lastTrustedBatches []*state.Batch
 		lastStateRoot      *common.Hash
 	}
+	// Id of the 'process' of the executor. Each time that starts this change this value
```
ToniRamirezM:
```suggestion
	// Id of the 'process' of the executor. Each time that it starts this value changes
```

synchronizer/synchronizer.go
```diff
@@ -47,6 +49,11 @@ type ClientSynchronizer struct {
...
+	// This value is obtained from the call state.GetStoredFlushID
+	// It start with a empty string and filled in the first call later
```
ToniRamirezM:
```suggestion
	// It starts as an empty string and it is filled in the first call
```

synchronizer/synchronizer.go
```diff
@@ -47,6 +49,11 @@ type ClientSynchronizer struct {
...
+	// is checked that is the same (in function checkFlushID)
```
ToniRamirezM:
```suggestion
	// later the value is checked to be the same (in function checkFlushID)
```

synchronizer/synchronizer.go
```diff
@@ -71,8 +79,10 @@ func NewSynchronizer(
 		cancelCtx:          cancel,
 		ethTxManager:       ethTxManager,
 		zkEVMClient:        zkEVMClient,
+		eventLog:           eventLog,
```
ToniRamirezM, line 82: Do you init the envetLog variable? It looks like it is going to be nil at this point.

synchronizer/synchronizer.go
```diff
@@ -1466,12 +1476,41 @@ func (s *ClientSynchronizer) getCurrentBatches(batches []*state.Batch, trustedBa
...
+			EventID:     event.EventID_SynchonizerRestart,
+			Description: fmt.Sprintf("proverID changed from %s to %s, restarting Synchonizer ", s.proverID, proverID),
+		}
+
+		err := s.eventLog.LogEvent(context.Background(), event)
+		if err != nil {
+			log.Errorf("error storing event payload: %v", err)
+		}
+
+		log.Fatal("restarting sequencer because  executor have restarted (old=%s, new=%s)", s.proverID, proverID)
```
ToniRamirezM:
```suggestion
		log.Fatal("restarting synchronizer because  executor have restarted (old=%s, new=%s)", s.proverID, proverID)
```

synchronizer/synchronizer.go
```diff
@@ -1466,12 +1476,41 @@ func (s *ClientSynchronizer) getCurrentBatches(batches []*state.Batch, trustedBa
...
+func (s *ClientSynchronizer) updateAndCheckProverID(proverID string) {
+	if s.proverID == "" {
+		s.proverID = proverID
+		return
+	}
+	if s.proverID != proverID {
+		event := &event.Event{
+			ReceivedAt:  time.Now(),
+			Source:      event.Source_Node,
+			Component:   event.Component_Sequencer,
```
ToniRamirezM:
```suggestion
			Component:   event.Component_Synchronizer,
```

synchronizer/synchronizer_test.go
```diff
@@ -28,14 +28,18 @@ type mocks struct {
 	EthTxManager *ethTxManagerMock
 	DbTx         *dbTxMock
 	ZKEVMClient  *zkEVMClientMock
+	//EventLog     *eventLogMock
 }
 
+//func Test_Given_StartingSynchronizer_When_CallFirstTimeExecutor_Then_StoreProverID(t *testing.T) {
```
ToniRamirezM, line 38: We should remove commented code.

ToniRamirezM (changes requested):

synchronizer/synchronizer_test.go
joanestebanr: I left this because is missing the unittest and I'm going to do it after the merge of the solution

synchronizer/synchronizer.go
joanestebanr: The function `NewSynchronizer` set this field as the incomming parameter `eventLog *event.EventLog`...  unless the caller pass a nil the field is !=nil, if I missing something let me known

ToniRamirezM (approved):

PR #2343, opened 2023-07-31
`Worker`'s `AddTxTracker` Bug Fix

### LINKED ISSUE #2342: Resolve `AddTxTracker` Function Bug in Worker Module
## Overview:

A bug has been identified in the `Worker`'s `AddTxTracker` function where a new `AddrQueue` is created for an address, just after its transactions have expired, but before the pending transactions for the same address are stored. As a result, the new `AddrQueue` is incorrectly initiated using `oldStateRoot`, old `Nonce`, and `Balance`. This bug disrupts the sequence of transactions and leads to potential data inconsistency issues and makes the sequencer to get stuck.

## Recommended Changes:

1. **Modify `AddTxTracker` Function:**
    - Adjust the `AddTxTracker` function in the `Worker` module to wait for the pending transactions to be completely stored before initiating a new `AddrQueue` for an address. This can be achieved by checking the status of pending transactions and delaying the `AddrQueue` creation until all transactions are confirmed to be stored.

## Objective:

The proposed changes aim to address the identified bug in the `Worker`'s `AddTxTracker` function. By modifying the function to wait for pending transactions to be stored before initiating a new `AddrQueue`, we ensure data consistency and prevent potential errors caused by this bug.

Psykepro:
Closes #2342. Related #2342 

### What does this PR do?

This PR resolves a bug in the `Worker`'s `AddTxTracker` function, which leads to disruptions in the transaction sequences and can cause data inconsistencies.

Previously, an `AddrQueue` for an address was created right after its transactions had expired, but before pending transactions were stored. This incorrect initialization of `AddrQueue` was the root cause of the issue.

To address this issue, instead of sharing `pendingTransactionsToStoreWG`, a new shared `pendingTxsPerAddressTrackers map[common.Address]*pendingTxPerAddressTracker` has been created to keep track of the pending transactions per address. This way, we ensure to wait only for the transactions to a particular address before initiating a new `AddrQueue`. The structure `pendingTxPerAddressTracker` includes a WaitGroup (`wg`) and a counter (`count`) to track the number of pending transactions per address and delete the tracker when not needed, preventing any potential memory leaks.

The `pendingTxPerAddressTracker` struct is defined as:
```go
type pendingTxPerAddressTracker struct {
	wg    *sync.WaitGroup
	count uint
}
```
### Reviewers
- @ToniRamirezM 
- @agnusmor

### DIFF
--- a/sequencer/worker.go
+++ b/sequencer/worker.go
@@ -23,11 +23,19 @@ type Worker struct {
 	state                        stateInterface
 	batchConstraints             batchConstraintsFloat64
 	batchResourceWeights         batchResourceWeights
-	pendingTransactionsToStoreWg *sync.WaitGroup
+	pendingTxsToStoreMux         *sync.RWMutex
+	pendingTxsPerAddressTrackers map[common.Address]*pendingTxPerAddressTracker
 }
 
 // NewWorker creates an init a worker
-func NewWorker(cfg WorkerCfg, state stateInterface, constraints batchConstraints, weights batchResourceWeights, pendingTxsToStoreWg *sync.WaitGroup) *Worker {
+func NewWorker(
+	cfg WorkerCfg,
+	state stateInterface,
+	constraints batchConstraints,
+	weights batchResourceWeights,
+	pendingTxsToStoreMux *sync.RWMutex,
+	pendingTxTrackersPerAddress map[common.Address]*pendingTxPerAddressTracker,
+) *Worker {
 	w := Worker{
 		cfg:                          cfg,
 		pool:                         make(map[string]*addrQueue),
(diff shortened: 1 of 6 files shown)


sequencer/worker.go
```diff
@@ -55,6 +57,9 @@ func (w *Worker) AddTxTracker(ctx context.Context, tx *TxTracker) (replacedTx *T
 		// Unlock the worker to let execute other worker functions while creating the new AddrQueue
 		w.workerMutex.Unlock()
 
+		// Wait until all pending transactions are stored, so we can ensure getting the correct nonce and balance of the new AddrQueue
+		w.pendingTransactionsToStoreWg.Wait()
```
jesusholoo: can happen that a lot of transactions cotinue to being processed meanwhile and this never ends?

ToniRamirezM: There is a limit of txs that fit in a batch, so this will end before closing a batch.

Psykepro: Hm yep, it's possible. Maybe to create a map which to keep track of the address for which we are waiting for pending tx. And when all txs for that address are saved then it can continue. Will change now the implementation.

jesusholoo: this seems better yes ^^

ToniRamirezM (approved):

PR #2345, opened 2023-07-31
Fix debug tracer error message

### LINKED ISSUE #2344: debug trace is not returning the error message
debug trace methods are not returning the error message when a transaction fails and it is not a reverted transaction, like out of gas error.

## Explanation 
The bug consist on a create transaction that has some error, for example out of gas. The normal tracer displays the error ( out of gas), but the callTracer does not and it affects the UX on etherscan. For instance:

### Expected behaviour
Here we have the trace of an out of gas contract creation on goerli:
https://goerli.etherscan.io/tx/0xd0a111139851a0dd0619318d422f56dccec8d5054bab71dbe000bfe7d7f8a06d

trace:

```json
{
 "from": "0x0cae25c8623761783fe4ce241c9b428126a7612a",
 "gas": "0x24",
 "gasUsed": "0xd066",
 "to": "0x0000000000000000000000000000000000000000",
 "input": "0x6080604052348015600f57600080fd5b600080f3fe",
 "error": "out of gas",
 "value": "0x0",
 "type": "CREATE"
}
```

### Actual behaviour
Here we have the trace of an out of gas contract creation on zkEVM testnet:
https://testnet-zkevm.polygonscan.com/tx/0x7d6929a8b20f0af40059f4b10b93b59604aaf93e0ef8ed19b8b29921949f7ce3

```json
{
 "from": "0x0cae25c8623761783fe4ce241c9b428126a7612a",
 "gas": "0xd066",
 "gasUsed": "0xd066",
 "to": "0xe1e57b3675cae033663cd3e495be4d17e16c910c",
 "input": "0x6080604052348015600f57600080fd5b600080f3fe",
 "value": "0x0",
 "type": "CREATE"
}
```

tclemos:
Closes #2344.

### What does this PR do?

adds the error message for trace responses 

### Reviewers

Main reviewers:

@ToniRamirezM


test/e2e/debug_shared.go
```diff
@@ -163,6 +163,48 @@ func createScDeployRevertedSignedTx(t *testing.T, ctx context.Context, auth *bin
...
+// 	tx := ethTypes.NewTx(&ethTypes.LegacyTx{
+// 		Nonce:    nonce,
+// 		GasPrice: gasPrice,
+// 		Gas:      uint64(150000),
+// 		Data:     data,
+// 	})
+
+// 	return auth.Signer(auth.From, tx)
+// }
+
```
ARR552, line 207: Why do we need this commented code?

ARR552 (approved):

test/e2e/debug_shared.go
tclemos: This is the test I've mentioned in the weekly that we need to investigate with the Executor team.
This is commented at the moment to not break the E3E tests and it is going to be used by this issue: #2641

PR #2363, opened 2023-08-03
Feature/2362 state use field batch l2data of processing context

### LINKED ISSUE #2362: Improve state interface to use new field `BatchL2Data` of `ProcessingContext` struct
# Rationale

The func `ProcessAndStoreClosedBatch` have the field `encodedTxs` that is the same as `state.ProcessingContext.BatchL2Data`. This create a inconsistence in the interface. 
Related to #2357 

# Implementation
Remove `encodedTxs` of  and use the new field instead.

joanestebanr:
Closes #2362

### What does this PR do?

`state.ProcessAndStoreClosedBatch` use `processingCtx.BatchL2Data` instead of `encodedTxs` parameter

### Reviewers
@ARR552


synchronizer/synchronizer.go
```diff
@@ -935,7 +936,8 @@ func (s *ClientSynchronizer) processSequenceBatches(sequencedBatches []etherman.
 				log.Errorf("error resetting trusted state. BatchNumber: %d, BlockNumber: %d, error: %v", batch.BatchNumber, blockNumber, err)
 				return err
 			}
-			_, flushID, proverID, err := s.state.ProcessAndStoreClosedBatch(s.ctx, processCtx, batch.BatchL2Data, dbTx, stateMetrics.SynchronizerCallerLabel)
+			processCtx.BatchL2Data = &batch.BatchL2Data
```
ARR552: I think this is redundant

state/batch.go
```diff
@@ -476,7 +482,7 @@ func (s *State) ProcessAndStoreClosedBatch(ctx context.Context, processingCtx Pr
 		StateRoot:     processedBatch.NewStateRoot,
 		LocalExitRoot: processedBatch.NewLocalExitRoot,
 		AccInputHash:  processedBatch.NewAccInputHash,
-		BatchL2Data:   encodedTxs,
+		BatchL2Data:   *processingCtx.BatchL2Data,
```
ARR552:
```suggestion
		BatchL2Data:   *BatchL2Data,
```
If processingCtx.BatchL2Data is nil, then will panic

ARR552 (changes requested):
As a general comment, I see that this code will save twice the batchL2Data. When the batch is open and when It's closed. Do you think this could have an impact in the db performance? (I don't think so but its better to double check)

synchronizer/synchronizer.go
joanestebanr: You are right! I'll remove this line

ARR552 (approved):

PR #2378, opened 2023-08-05
Increment gas price with at least 10%

jbearer:
Many RPC and miner implementations automatically discard replacement transactions where the gas price is not increased by at least 10%. If this happens, the ethtxmanager can end up adding transactions to the history of a monitored transaction that have no effect, until the history size limit is quickly reached and the monitored transaction is considered failed.

### DIFF
--- a/ethtxmanager/ethtxmanager.go
+++ b/ethtxmanager/ethtxmanager.go
@@ -24,7 +24,7 @@ const (
 	// maxHistorySize           = 10
 
 	// The minimum factor by which we must increase gas price in a replacement transaction.
-	minGasPriceIncreaseNumerator = 11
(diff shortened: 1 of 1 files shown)


sveitser:
Happy to sign the CLA. Do I need to make a PR to https://github.com/0xPolygonHermez/clabot-config/ ?

tclemos:
@jbearer, thanks for your suggestion!

We had a discussion internally and we will be applying some changes to the `EthTxmanager` to handle this and other situations soon.
We have already merged a PR to help with this specific situation, as you can see here: 
https://github.com/0xPolygonHermez/zkevm-node/pull/2360

jbearer:
Ah don't know how I missed that. Glad to hear it's fixed!

PR #2379, opened 2023-08-06
updated docker compose command in Makefile for testing

Selfie21:
### What does this PR do?
The Makefile in the test folder uses the old calling convention of docker compose see with a dash, the pull request fixes this see: https://docs.docker.com/compose/migrate/ 

### Main reviewers:
    @ToniRamirezM 
    @ARR552


tclemos:
We don't plan to update the Makefile to the Docker Compose v2 syntax at the moment.
If you are facing issues with this, please check this topic:
https://stackoverflow.com/questions/72099653/how-to-alias-docker-compose-to-docker-compose

Issue #2388, opened 2023-08-08
make: `docker-compose`: No such file or directory


iprotoni:
#### System information

OS & Version: `Ubuntu 22.04`
Commit hash : `b1e8994d04e67f41c2f49829e6ff51c84e5a5c82`
Network: `Mainnet/Testnet`

#### Expected behaviour
`cd test/`
`make run`
Should be compile, but getting error with following screenshot.
$ make run
docker-compose -f docker-compose.yml up -d zkevm-state-db
make: docker-compose: No such file or directory
make: *** [Makefile:437: run] Error 127

#### Actual behaviour
should be build without getting issue

#### Steps to reproduce the behaviour
`cd test/`
`make run`

tclemos:
You need to install `docker-compose` to be able to run this command. 
Check here for more details: https://docs.docker.com/compose/install/linux/

https://github.com/0xPolygonHermez/zkevm-node/blob/b1e8994d04e67f41c2f49829e6ff51c84e5a5c82/test/docker-compose.yml#L49-L51

The, as you can see here in the `docker-compose` file, it depends on the node image to be built locally. 
Before running the command `make run`, make sure to run `make build-docker` at the root directory.



steps:
```bash
$ make build-docker
$ cd test
$ make run
```

PR #2389, opened 2023-08-08
Fix docker command

iprotoni:
https://github.com/0xPolygonHermez/zkevm-node/issues/2388


tclemos:
We don't plan to update the make file to the docker-compose v2 syntax at the moment.
If you are facing issues with this, please check this topic: 
https://stackoverflow.com/questions/72099653/how-to-alias-docker-compose-to-docker-compose

PR #2398, opened 2023-08-10
Add forced batches tx to addrQueue

### LINKED ISSUE #2395: Add forced batches txs to worker/addrQueue list
# Rationale

The forced batches txs are not store in the worker/addrQueue list, therefore we are not updating the changes (nonce/balance) of the addresses touched by the execution of the forced batch and this can cause some issues when processing future txs for those addresses

# Implementation

When processing a forced batch, before to do the execution (executor) we will add all the txs included in the forced batch in the worker list (this will create the addrQueue for each address if it doesn't already exists). After this we will execute the FB and after the execution we will get the touched addresses from the executor and we will update the nonce/balance of these addresses in the worker

ToniRamirezM:
Closes #2395 

### What does this PR do?

Add forced txs to addrqueue to have the same handling as regular ones regarding addrQueue.

### Reviewers

Main reviewers:

- @agnusmor 
- @Psykepro

### DIFF
--- a/sequencer/addrqueue.go
+++ b/sequencer/addrqueue.go
@@ -80,8 +80,8 @@ func (a *addrQueue) addTx(tx *TxTracker) (newReadyTx, prevReadyTx, replacedTx *T
 	}
 }
 
-// addForcedTX adds a forced tx to the list of forced txs
-func (a *addrQueue) addForcedTX(txHash common.Hash) {
+// addForcedTx adds a forced tx to the list of forced txs
+func (a *addrQueue) addForcedTx(txHash common.Hash) {
 	a.forcedTxs[txHash] = struct{}{}
 }
 
@@ -144,7 +144,6 @@ func (a *addrQueue) deleteTx(txHash common.Hash) (deletedReadyTx *TxTracker) {
 func (a *addrQueue) deleteForcedTx(txHash common.Hash) {
 	if _, found := a.forcedTxs[txHash]; found {
 		delete(a.forcedTxs, txHash)
-		log.Infof("DeleteForcedTx tx(%s) deleted from addrQueue", txHash.String())
 	} else {
 		log.Warnf("tx (%s) not found in forcedTxs list", txHash.String())
 	}
--- a/sequencer/addrqueue.go
+++ b/sequencer/addrqueue.go
@@ -18,6 +18,7 @@ type addrQueue struct {
 	currentBalance    *big.Int
 	readyTx           *TxTracker
 	notReadyTxs       map[uint64]*TxTracker
+	forcedTxs         map[common.Hash]struct{}
 	pendingTxsToStore map[common.Hash]struct{}
 }
 
@@ -30,6 +31,7 @@ func newAddrQueue(addr common.Address, nonce uint64, balance *big.Int) *addrQueu
 		currentBalance:    balance,
 		readyTx:           nil,
 		notReadyTxs:       make(map[uint64]*TxTracker),
+		forcedTxs:         make(map[common.Hash]struct{}),
 		pendingTxsToStore: make(map[common.Hash]struct{}),
 	}
 }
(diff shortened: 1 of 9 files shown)


sequencer/addrqueue.go
```diff
@@ -78,6 +80,11 @@ func (a *addrQueue) addTx(tx *TxTracker) (newReadyTx, prevReadyTx, replacedTx *T
 	}
 }
 
+// addForcedTX adds a forced tx to the list of forced txs
+func (a *addrQueue) addForcedTX(txHash common.Hash) {
```
agnusmor: addForcedTx (x lowercase to be consistent with addTx and deleteForcedTx function)

sequencer/addrqueue.go
```diff
@@ -133,6 +140,16 @@ func (a *addrQueue) deleteTx(txHash common.Hash) (deletedReadyTx *TxTracker) {
 	}
 }
 
+// deleteForcedTx deletes the tx from the addrQueue
+func (a *addrQueue) deleteForcedTx(txHash common.Hash) {
+	if _, found := a.forcedTxs[txHash]; found {
+		delete(a.forcedTxs, txHash)
+		log.Infof("DeleteForcedTx tx(%s) deleted from addrQueue", txHash.String())
```
agnusmor: This log is needed? We don't have it in the deletePendingTxToStore, maybe keep it but only in debug level? Anyway we have also a debug message for this in the updateWorkerAfterSuccessfulProcessing function

sequencer/finalizer.go
```diff
@@ -765,20 +792,26 @@ func (f *finalizer) storeProcessedTx(ctx context.Context, txToStore transactionT
...
 }
 
-func (f *finalizer) updateWorkerAfterSuccessfulProcessing(ctx context.Context, tx *TxTracker, result *state.ProcessBatchResponse) {
+func (f *finalizer) updateWorkerAfterSuccessfulProcessing(ctx context.Context, txHash common.Hash, txFrom common.Address, isForced bool, result *state.ProcessBatchResponse) {
 	// Delete the transaction from the txSorted list
-	f.worker.DeleteTx(tx.Hash, tx.From)
-	log.Debug("tx deleted from txSorted list", "txHash", tx.Hash.String(), "from", tx.From.Hex())
+	if isForced {
+		f.worker.DeleteForcedTx(txHash, txFrom)
+		log.Debug("forced tx deleted from addrQueue", "txHash", txHash.String(), "from", txFrom.Hex())
```
agnusmor: "forced tx delete from worker"

sequencer/finalizer.go
```diff
@@ -765,20 +792,26 @@ func (f *finalizer) storeProcessedTx(ctx context.Context, txToStore transactionT
...
+		return
+	} else {
+		f.worker.DeleteTx(txHash, txFrom)
+		log.Debug("tx deleted from txSorted list", "txHash", txHash.String(), "from", txFrom.Hex())
```
agnusmor: "tx delete from worker"

sequencer/finalizer_test.go
```diff
@@ -86,11 +89,11 @@ var (
 	newHash         = common.HexToHash("0x02")
 	newHash2        = common.HexToHash("0x03")
 	stateRootHashes = []common.Hash{oldHash, newHash, newHash2}
-	txHash          = common.BytesToHash([]byte("txHash"))
-	txHash2         = common.BytesToHash([]byte("txHash2"))
+	txHash          = common.BytesToHash([]byte("0x00000"))
```
Psykepro: Everything is good. But why do we change all hashes to be the same value? It's good to have them as different values when debugging unit tests.

Psykepro (changes requested):

sequencer/finalizer_test.go
ToniRamirezM: I will revert these changes and try to pass the tests. Also there is a test I commented because it was not passing.
Now for forced butches we get the sender from the signature, and I had to change several tests to use signed txs instead of non-related batchdata and sender, because the dummy sender was not matching, or we didn't have a signature at all.

Psykepro (approved):

agnusmor (approved):

PR #2431, opened 2023-08-21
Introduce Config Parameters to Control RPC Batch Requests

### LINKED ISSUE #2307: Add Upper Limit on `RPC` Batch Requests
# Overview:

There's a need to improve our system's handling of batched requests. Currently, the system does not enforce restrictions on the number of batched requests.

## Recommended Changes:

1. **Control Over Batch Requests:**
   - Introduce a new config parameter `RPC.BatchRequestsEnabled`, with a default value of `false`, which will allow enabling or disabling batch requests via HTTP.
   - Introduce a new config parameter `RPC.BatchRequestsLimit`, with a default value of `20`, which will limit the number of requests that can be incorporated into each batch request. If this parameter is set to zero, there will be no limit on the number of requests in a batch.

## Objective:

The recommended improvements aim to streamline the system's processing of batch requests. These changes will ensure the system's capability to efficiently manage a large number of requests within a single batch, thus enhancing overall operational efficiency. By limiting the number of requests in a batch and allowing batch requests to be enabled or disabled via configuration, we can help maintain system performance without unnecessary memory pressure.

Psykepro:
Closes #2307.

#### Changes
1. **Control Over Batch Requests:**
   - Introduce a new config parameter `RPC.BatchRequestsEnabled`, with a default value of `false`, which will allow enabling or disabling batch requests via HTTP.
   - Introduce a new config parameter `RPC.BatchRequestsLimit`, with a default value of `20`, which will limit the number of requests that can be incorporated into each batch request. If this parameter is set to zero, there will be no limit on the number of requests in a batch.

 ```toml
    [RPC]
    # ....
    BatchRequestsEnabled = false
    BatchRequestsLimit = 20
    # ...
```    

### Reviewers

Main reviewers:
- @tclemos 
- @ToniRamirezM


tclemos (approved):
I've extended the `jsonrpc/client` to be able to do `BatchCalls` and added tests to check the new validations.

I've also refactored the code, reviewing the way we were returning the errors and how we were logging it, to avoid false positive errors and have better metrics between invalid requests and requests with errors.

PR #2450, opened 2023-08-23
Feature/#2429 fork id improvement

### LINKED ISSUE #2429: ForkID sync improvement
Create a db table to store the forkIDs
Create a parallel sync to get all ForkIDs in chunks of 2000 blocks if the db table is empty
If the table is not empty, sync forkIDs during the normal synchronization

ARR552:
Closes #2429 

### What does this PR do?

This Pr adds a forkID table. If a component needs the forkID list goes to db instead of to read the info to L1. If the table is empty the synchronizer fills it until the current synced batch. During the sync process, new forkIDs are stored in db and loaded in memory.
*Important points: *
If the synchronization has not finished, do not start the aggregator because the currentForkID could be wrong.
If the node is running using microservices, other components needs to be restarted in order to be aware of the new forkID. The synchronizer is the only one that will realize that the forkID has changed. If all the components are running together, this is not a problem because the synchronizer will keep updated the memory variables shared by all the components.

Config changes:
[Etherman]
ForkIDChunkSize = 20000

### Reviewers

@agnusmor 
@ToniRamirezM 
@joanestebanr


cmd/run.go
```diff
@@ -572,3 +567,62 @@ func logVersion() {
 		"os/arch", fmt.Sprintf("%s/%s", runtime.GOOS, runtime.GOARCH),
 	)
 }
+
+func forkIDIntervals(ctx context.Context, st *state.State, etherman *etherman.Client, genesisBlockNumber uint64) ([]state.ForkIDInterval, error) {
+	dbTx, err := st.BeginStateTransaction(ctx)
```
agnusmor: I think BeginStateTransaction can be placed just before the for loop to store forkids in the db (line 598) and do the commit just after end the loop (line 605). Anyway we are not doing a rollback when returning from this function with an error. If we move the BeginStateTransaction just before the for loop, then we need only to do a rollback if we have an error when inserting a forkid in the db

cmd/run.go
```diff
@@ -572,3 +567,62 @@ func logVersion() {
...
+			forkIDIntervals = forkIntervals
+		} else {
+			log.Debug("Getting initial forkID")
+			forkIntervals, err := etherman.GetForks(ctx, genesisBlockNumber, genesisBlockNumber)
+			if err != nil {
+				return []state.ForkIDInterval{}, fmt.Errorf("error getting forks. Please check the configuration. Error: %v", err)
+			} else if len(forkIntervals) == 0 {
+				return []state.ForkIDInterval{}, fmt.Errorf("error: no forkID received. It should receive at least one, please check the configuration...")
+			}
+			forkIDIntervals = forkIntervals
```
agnusmor, line 633: We are not storing this initial forkID in the db

cmd/run.go
```diff
@@ -572,3 +567,62 @@ func logVersion() {
...
+	}
+	log.Debug("numberForkIDs: ", numberForkIDs)
+	var forkIDIntervals []state.ForkIDInterval
+	if numberForkIDs == 0 {
+		// Get last L1block Synced
+		lastBlock, err := st.GetLastBlock(ctx, dbTx)
+		if err != nil && !errors.Is(err, state.ErrStateNotSynchronized) {
+			return []state.ForkIDInterval{}, fmt.Errorf("error checking lastL1BlockSynced. Error: %v", err)
+		}
+		if lastBlock != nil {
```
agnusmor, line 586: I think we can change this if/else block code for this one, since the if and else code are the same, since I understand also that we need to store the forkid interval in the case of the initial fork (genesis):

```
lastBlockNumber := genesisBlockNumber
if lastBlock != nil {
   lastBlockNumber := lastBlock.BlockNumber
}

log.Info("Getting forkIDs intervals. Please wait...")
// Read Fork ID FROM POE SC
forkIntervals, err := etherman.GetForks(ctx, genesisBlockNumber, lastBlockNumber)
if err != nil {
   return []state.ForkIDInterval{}, fmt.Errorf("error getting forks. Please check the configuration. Error: %v", err)
} else if len(forkIntervals) == 0 {
   return []state.ForkIDInterval{}, fmt.Errorf("error: no forkID received. It should receive at least one, please check the configuration...")
}
log.Info("Storing forkID intervals into db")
// Store forkIDs
dbTx, err := st.BeginStateTransaction(ctx)
if err != nil {
   return []state.ForkIDInterval{}, fmt.Errorf("error creating dbTx. Error: %v", err)
}
for _, f := range forkIntervals {
   err := st.AddForkID(ctx, f, dbTx)
   if err != nil {
   err = dbTx.Rollback(ctx)
   if err != nil {
      return []state.ForkIDInterval{}, fmt.Errorf("error when rollback dbTx. Error: %v", err)
   }
      return []state.ForkIDInterval{}, fmt.Errorf("error adding forkID to db. Error: %v", err)
   }
}
err = dbTx.Commit(ctx)
if err != nil {
   return []state.ForkIDInterval{}, fmt.Errorf("error committing dbTx. Error: %v", err)
}
forkIDIntervals = forkIntervals

```

cmd/run.go
```diff
@@ -572,3 +567,62 @@ func logVersion() {
 		"os/arch", fmt.Sprintf("%s/%s", runtime.GOOS, runtime.GOARCH),
 	)
 }
+
+func forkIDIntervals(ctx context.Context, st *state.State, etherman *etherman.Client, genesisBlockNumber uint64) ([]state.ForkIDInterval, error) {
+	dbTx, err := st.BeginStateTransaction(ctx)
+	if err != nil {
+		return []state.ForkIDInterval{}, fmt.Errorf("error creating dbTx. Error: %v", err)
+	}
+	numberForkIDs, err := st.CountForkIDs(ctx, dbTx)
```
agnusmor: We can put here the call to the function to get the forkids intervals:

`forkIDIntervals, err = st.GetForkIDs(ctx, dbTx)`

Then if len(forkIDIntervals)==0 we can execute the code to get it from L1, else we return the forkIDIntervals

Doing in this way we don't need to use CountForkIDs function and we are removing this query to the db

config/environments/local/local.node.config.toml
```diff
@@ -34,6 +34,7 @@ PollMinAllowedGasPriceInterval = "15s"
 
 [Etherman]
 URL = "http://your.L1node.url"
+ForkIDChunkSize = 2000
```
agnusmor: I think we miss to define the default value for ForkIDCheckSize in default.go. 2000 is a good default value? or maybe we can use a higher value to reduce the amount of L1 block requests?

db/migrations/state/0008.sql
```diff
@@ -0,0 +1,13 @@
+-- +migrate Up
+CREATE TABLE IF NOT EXISTS state.fork_id
```
agnusmor, line 2: I think is good practice to define always a primary key for the tables, in this case I understand the PK is fork_id

etherman/etherman.go
```diff
@@ -228,19 +230,32 @@ func (etherMan *Client) VerifyGenBlockNumber(ctx context.Context, genBlockNumber
...
-		Topics:    [][]common.Hash{{updateZkEVMVersionSignatureHash}},
-	}
-	logs, err := etherMan.EthClient.FilterLogs(ctx, query)
-	if err != nil {
-		return []state.ForkIDInterval{}, err
+	var logs []types.Log
+	log.Debug("Using ForkIDChunkSize: ", etherMan.cfg.ForkIDChunkSize)
+	for i := genBlockNumber; i <= lastL1BlockSynced; i = i + etherMan.cfg.ForkIDChunkSize + 1 {
+		final := i + etherMan.cfg.ForkIDChunkSize
+		if i+etherMan.cfg.ForkIDChunkSize > lastL1BlockSynced {
```
agnusmor: if final > lastL1BlockSynced {

state/forkid.go
```diff
@@ -17,6 +23,37 @@ func (s *State) UpdateForkIDIntervals(intervals []ForkIDInterval) {
 	s.cfg.ForkIDIntervals = intervals
```
agnusmor, line 24: To avoid misunderstanding I suggest to rename this function to something like UpdateForkIdIntervalsStruct, as the current name seems that is updating the DB

state/forkid.go
```diff
@@ -17,6 +23,37 @@ func (s *State) UpdateForkIDIntervals(intervals []ForkIDInterval) {
...
+	// Add forkId to db and memori variable
+	oldForkIDs, err := s.GetForkIDs(ctx, dbTx)
+	if err != nil {
+		log.Error("error getting oldForkIDs. Error: ", err)
+		return err
+	}
+	if len(oldForkIDs) == 0 {
+		s.UpdateForkIDIntervals([]ForkIDInterval{newForkID})
+	} else {
+		var forkIDs []ForkIDInterval
```
agnusmor, line 38: Could be is not necessary, but we need to check that newForkID.ForkId is just the next one (+1) of oldForkIDs[len(oldForkIDs)-1].ForkId, if not we can return an error. With this we ensure we are adding the next forkid in the list

test/operations/manager.go
```diff
@@ -168,11 +169,35 @@ func (m *Manager) SetGenesis(genesisActions []*state.GenesisAction) error {
...
-	err = dbTx.Commit(m.ctx)
+	errCommit := dbTx.Commit(m.ctx)
+	if errCommit != nil {
+		return errCommit
+	}
+
+	return err
+}
+
+func (m *Manager) SetForkID(forkID uint64) error {
```
agnusmor, line 181: I understand this function is for testing propouses, isn't it? if it's the case please add a comment line just before the function signature saying this

cmd/run.go
ARR552: Yes, we are. Why do you say we are not storing it?

db/migrations/state/0008.sql
ARR552: I don't fully understand what you mean. As you can see the fork_id is defined as Primary key
`    fork_id        BIGINT NOT NULL PRIMARY KEY,
`

config/environments/local/local.node.config.toml
ARR552: WOOOW You are right. I forgot to add it there. Good catch!

state/forkid.go
ARR552: Done

test/operations/manager.go
ARR552: Yes, all functions inside test folder are for testing purposes

ToniRamirezM (approved):

PR #2476, opened 2023-08-30
remove need of decoded txs to process a batch response

ToniRamirezM:
### What does this PR do?

Removes the need for decoded transaction in order to process and convert a process batch response from the executor.

### Reviewers

Main reviewers:

- @Psykepro 
- @agnusmor 
- @ARR552

### DIFF
--- a/state/converters.go
+++ b/state/converters.go
@@ -32,12 +32,12 @@ func ConvertToCounters(resp *executor.ProcessBatchResponse) ZKCounters {
 }
 
 // TestConvertToProcessBatchResponse for test purposes
-func (s *State) TestConvertToProcessBatchResponse(txs []types.Transaction, response *executor.ProcessBatchResponse) (*ProcessBatchResponse, error) {
-	return s.convertToProcessBatchResponse(txs, response)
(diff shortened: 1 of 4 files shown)


state/converters.go
```diff
@@ -151,9 +151,8 @@ func (s *State) convertToProcessTransactionResponse(txs []types.Transaction, res
 		result.CallTrace = *callTrace
 		result.EffectiveGasPrice = response.EffectiveGasPrice
 		result.EffectivePercentage = response.EffectivePercentage
-		result.Tx = txs[i]
 
-		_, err = DecodeTx(common.Bytes2Hex(response.GetRlpTx()))
+		tx, err := DecodeTx(common.Bytes2Hex(response.GetRlpTx()))
```
ARR552, line 155: Is there any way to use the txs received from the executor?? Maybe it is better to rely on the executor to decode txs

Psykepro (approved):

state/converters.go
ToniRamirezM: Yes, this is the TX we later store in the trusted state. The executor returns it in rlp format.

arnaubennassar (commented):
lgtm

ARR552 (approved):

PR #2479, opened 2023-08-30
Cherrypick/fix convert process batch response

ToniRamirezM:
Updates develop with the contents of v0.3.1

### DIFF
--- a/sequencer/finalizer.go
+++ b/sequencer/finalizer.go
@@ -1013,6 +1013,8 @@ func (f *finalizer) processForcedBatch(ctx context.Context, lastBatchNumberInSta
 					continue
 				}
 				f.worker.AddForcedTx(txResponse.TxHash, sender)
+			} else {
+				log.Warnf("ROM_ERROR_INVALID_RLP error received from executor for forced batch %d", forcedBatch.ForcedBatchNumber)
 			}
 		}
 
--- a/sequencer/finalizer.go
+++ b/sequencer/finalizer.go
@@ -1006,12 +1006,14 @@ func (f *finalizer) processForcedBatch(ctx context.Context, lastBatchNumberInSta
 
 	if len(response.Responses) > 0 && !response.IsRomOOCError {
 		for _, txResponse := range response.Responses {
-			sender, err := state.GetSender(txResponse.Tx)
-			if err != nil {
-				log.Warnf("failed trying to add forced tx (%s) to worker. Error getting sender from tx, Err: %v", txResponse.TxHash, err)
-				continue
+			if !errors.Is(txResponse.RomError, executor.RomErr(executor.RomError_ROM_ERROR_INVALID_RLP)) {
+				sender, err := state.GetSender(txResponse.Tx)
+				if err != nil {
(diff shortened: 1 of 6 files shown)


sequencer/finalizer.go
```diff
@@ -1006,12 +1007,16 @@ func (f *finalizer) processForcedBatch(ctx context.Context, lastBatchNumberInSta
...
-				continue
+			if !errors.Is(txResponse.RomError, executor.RomErr(executor.RomError_ROM_ERROR_INVALID_RLP)) {
+				sender, err := state.GetSender(txResponse.Tx)
+				if err != nil {
+					log.Warnf("failed trying to add forced tx (%s) to worker. Error getting sender from tx, Err: %v", txResponse.TxHash, err)
+					continue
+				}
+				f.worker.AddForcedTx(txResponse.TxHash, sender)
+			} else {
+				log.Warnf("ROM_ERROR_INVALID_RLP error received from executor for forced batch %d", forcedBatch.ForcedBatchNumber)
```
agnusmor, line 1018: Could be is useful to log here also the txResponse.TxHash (I understand that we have a valid value on txResponse.TxHash although we got an error for this tx):

`log.Warnf("ROM_ERROR_INVALID_RLP error received from executor for forced tx (%s) in forced batch %d", txResponse.TxHash, forcedBatch.ForcedBatchNumber)
`

state/converters.go
```diff
@@ -151,39 +151,51 @@ func (s *State) convertToProcessTransactionResponse(txs []types.Transaction, res
...
+
+				eventErr := s.eventLog.LogEvent(context.Background(), event)
+				if eventErr != nil {
+					log.Errorf("error storing payload: %v", err)
+				}
+
+				return nil, err
+			}
+		} else {
+			log.Warnf("ROM_ERROR_INVALID_RLP returned by the executor")
```
agnusmor, line 179: Log also tx hash?

state/converters.go
```diff
@@ -151,39 +151,51 @@ func (s *State) convertToProcessTransactionResponse(txs []types.Transaction, res
...
-				log.Errorf("error storing payload: %v", err)
+				timestamp := time.Now()
+				log.Errorf("error decoding rlp returned by executor %v at %v", err, timestamp)
+
+				event := &event.Event{
+					ReceivedAt: timestamp,
+					Source:     event.Source_Node,
+					Level:      event.Level_Error,
+					EventID:    event.EventID_ExecutorRLPError,
+					Json:       string(response.GetRlpTx()),
```
agnusmor, line 168: Log also tx hash?

sequencer/finalizer.go
ToniRamirezM: If the batchdata cannot be decoded into RLP it is also not possible to calculate its hash. That's why in other point of the code we log the forced batch number, to know where this tx came from.

state/converters.go
ToniRamirezM: Same comment as before.

state/converters.go
ToniRamirezM: And again :)

agnusmor (approved):

PR #2488, opened 2023-08-31
handle invalid rlp error

ToniRamirezM:
Handles invalid rlp error returned by executor in fork id 5.

### DIFF
--- a/state/converters.go
+++ b/state/converters.go
@@ -154,7 +154,7 @@ func (s *State) convertToProcessTransactionResponse(responses []*executor.Proces
 
 		tx := new(types.Transaction)
 
-		if len(response.GetRlpTx()) > 0 {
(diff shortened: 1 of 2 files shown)


state/converters.go
```diff
@@ -176,25 +176,24 @@ func (s *State) convertToProcessTransactionResponse(responses []*executor.Proces
 				return nil, err
 			}
 		} else {
-			log.Warnf("ProcessTransactionResponse[GetRlpTx]: empty for tx %v", result.TxHash)
+			log.Warnf("Invalid RLP (tx hash may be empty) for tx %v", result.TxHash)
```
ARR552: what about printing the rlp here? Is the result.TxHash needed?? If the full rlp is wrong ithink we won't have the txHash

ARR552 (approved):

PR #2409, opened 2023-08-14
Feature/#2371 synchronizer optimization

joanestebanr:
Closes #2371

### What does this PR do?

This is a refactor of L1 synchronization to improve speed.
- It ask data in parallel  to L1 meanwhile another goroutine is execution the rollup info.
- It makes that executor be ocupied 100% of time.

## Pending to do
  - **This branch was created long time ago need to be updated to be able to merge with develop**
  - All the stuff related to updating last block on L1 could be moved to another class
  - Check context usage:
    It need a context to cancel itself and create another context to cancel workers?
  - Emit metrics
  - if nothing to update reduce  code to be executed (not sure, because functionality  to keep update beyond last block on L1)
  - Improve the unittest of all objects
  - Check all log.fatals to remove it or add a status before the panic
  - Missing **feature update beyond last block on L1**: Old syncBlocks method try to ask for blocks over last L1 block, I suppose that is to keep   synchronizing even a long the synchronization have new blocks. This is not implemented here
    This is the behaviour of ethman in that situation:
           - GetRollupInfoByBlockRange returns no errors, zero blocks...
           - EthBlockByNumber returns error:  "not found"
- Some test on ` synchronizer/synchronizer_test.go` are based on this feature, so are running against legacy code
- Move to configuration file some 'hardcoded' values

## Configuration
This feature is experimental for that reason you can configure to use old sequential one: 
```
[Synchronizer]
UseParallelModeForL1Synchronization = false
```
If you activate this feature you can configure:
- `NumberOfParallelOfEthereumClients`: how many parallel request can be done. Currently this create multiples instances of etherman over same server, in the future maybe make sense to use differents servers
- `CapacityOfBufferingRollupInfoFromL1`:  buffer of data pending to be processed
```
UseParallelModeForL1Synchronization = true
	[Synchronizer.L1ParallelSynchronization]
		NumberOfParallelOfEthereumClients = 2
		CapacityOfBufferingRollupInfoFromL1 = 10
```
### How to known the occupation of executor
To check that executor are fully ocuppied you can check next log:
```
INFO	synchronizer/l1_processor_consumer.go:110	consumer: processing rollupInfo #1291: range:[188064, 188164] num_blocks [0] wasted_time_waiting_for_data [74.17575ms] last_process_time [2.534115ms] block_per_second [0.000000]
```
The `wasted_time_waiting_for_data` show the waiting time between this call and the previous to executor. If this value (after 20 interations) are greater to 1 seconds a warning is show.

[...]

PR #2504, opened 2023-09-04
L1 synchronization in parallel

### LINKED ISSUE #2371: L1 Synchronizer optimization
The objective is parallelized the read of events from L1 nodes and process the information read from L1.
To achieve this change, a go routine is needed to read the events from L1. The results are going to be sent to the processAndStore using a channel. The channel must have a defined length by a configuration parameter to allow us to fine tune the value. The default value could be 50 for instance. If the channel is full the go routine should wait.
The processAndStore must work exactly as right now. First the trusted state sync and then L1 processAndStore. If the channel is empty, the processAndStore should do nothing.

joanestebanr:
Closes #2371
* This PR is the same as [2409](https://github.com/0xPolygonHermez/zkevm-node/pull/2409) without the refactor of trusted synchronization because it changes too much in 0.2.x and the code is outdated

### What does this PR do?

This is a refactor of L1 synchronization to improve speed.
- It ask data in parallel  to L1 meanwhile another goroutine is execution the rollup info.
- It makes that executor be ocupied 100% of time.

## Pending to do
  - Improve the unittest of all objects
  - Check all log.fatals to remove it or add a status before the panic
- Some test on ` synchronizer/synchronizer_test.go` are based on this feature, so are running against legacy code
- Move to configuration file some 'hardcoded' values

## Configuration
This PR modify default configuration (`config/default.go`) setting next values: 
```
UseParallelModeForL1Synchronization = true
	[Synchronizer.L1ParallelSynchronization]
		NumberOfParallelOfEthereumClients = 2
		CapacityOfBufferingRollupInfoFromL1 = 10
		TimeForCheckLastBlockOnL1Time = "5s"
		TimeoutForRequestLastBlockOnL1 = "5s"
		MaxNumberOfRetriesForRequestLastBlockOnL1 = 3
		TimeForShowUpStatisticsLog = "5m"
		TimeOutMainLoop = "5m"
		[Synchronizer.L1ParallelSynchronization.PerformanceCheck]
			AcceptableTimeWaitingForNewRollupInfo = "5s"
			NumIterationsBeforeStartCheckingTimeWaitinfForNewRollupInfo = 10
```

This feature is experimental for that reason you could configure to use old sequential changing that: 
```
[Synchronizer]
UseParallelModeForL1Synchronization = false
```
If you activate this feature you can configure:
- `NumberOfParallelOfEthereumClients`: how many parallel request can be done. Currently this create multiples instances of etherman over same server, in the future maybe make sense to use differents servers
- `CapacityOfBufferingRollupInfoFromL1`:  buffer of data pending to be processed
```
UseParallelModeForL1Synchronization = true
	[Synchronizer.L1ParallelSynchronization]
		NumberOfParallelOfEthereumClients = 2
		CapacityOfBufferingRollupInfoFromL1 = 10
```
### How to known the occupation of executor
To check that executor are fully ocuppied you can check next log:
```
INFO	synchronizer/l1_processor_consumer.go:110	consumer: processing rollupInfo #1291: range:[188064, 188164] num_blocks [0] wasted_time_waiting_for_data [74.17575ms] last_process_time [2.534115ms] block_per_second [0.000000]
```
The `wasted_time_waiting_for_data` show the waiting time between this call and the previous to executor. If this value (after 20 interations) are greater to 1 seconds a warning is show.

### Estimated time to be fully synchronizer with L1
This log show the estimated time (**ETA**) to reach the block goal
```
INFO	synchronizer/l1_data_retriever_producer.go:255	producer: Statistics:ETA: 3h40m1.311379085s percent:1.35  blocks_per_seconds:706.80 pending_block:127563/9458271 num_errors:0
```

## Flow of data
l1_sync_channels_flow_v2 drawio

## Class diagram
This is a class diagram of principal class an relationships.
The entry point is `synchronizer.go:276` function `syncBlocksParallel`.
- It create all objects needed and launch `l1SyncOrchestration` that wait until the job is done to return 

### The main objects are:
- `l1RollupInfoProducer`: is the object that send rollup data through the channel
- `l1RollupInfoConsumer`: that receive the data and execute it

image

### Reviewers

Main reviewers:
- @ARR552 
- @ToniRamirezM 
- @agnusmor


cmd/run.go
```diff
@@ -295,8 +295,21 @@ func runSynchronizer(cfg config.Config, etherman *etherman.Client, ethTxManager
 	}
 	zkEVMClient := client.NewClient(trustedSequencerURL)
 
+	//var etherman2 *etherman.Client = nil
```
ARR552: unnecessary code

cmd/run.go
```diff
@@ -295,8 +295,21 @@ func runSynchronizer(cfg config.Config, etherman *etherman.Client, ethTxManager
...
+	// If synchronizer are using sequential mode, we only need one etherman client
+	if cfg.Synchronizer.UseParallelModeForL1Synchronization {
+		for i := 0; i < int(cfg.Synchronizer.L1ParallelSynchronization.NumberOfParallelOfEthereumClients); i++ {
+			eth, err := newEtherman(cfg)
+			if err != nil {
+				log.Fatal(err)
+			}
+			etherManForL1 = append(etherManForL1, eth)
+		}
+	}
```
ARR552, line 309: if we run parallel mode, etherman variable won't be use for the synchronizer. You can use that instance for the first worker and create the others.

synchronizer/l1_worker.go
```diff
@@ -0,0 +1,180 @@
...
+	default:
+		return "unknown"
+	}
+}
+
+const (
+	errWorkerBusy = "worker is busy"
+)
+
+type genericResponse[T any] struct {
```
ARR552: I don't understand why a generic response is needed

synchronizer/l1_worker.go
```diff
@@ -0,0 +1,180 @@
...
+		var lastBlock *types.Block = nil
+		if err == nil && len(blocks) == 0 {
+			lastBlock, err = w.etherman.EthBlockByNumber(ctx, toBlock)
+		}
+		duration := time.Since(now)
+		result := newGenericAnswer(err, duration, typeRequestRollupInfo, &responseRollupInfoByBlockRange{blockRange, blocks, order, lastBlock})
+		w.setStatus(ethermanIdle)
+		ch <- result
+	}
+	go launch()
```
ARR552:
```suggestion
	go func() {
		if wg != nil {
			defer wg.Done()
		}
		now := time.Now()
		fromBlock := blockRange.fromBlock
		toBlock := blockRange.toBlock
		blocks, order, err := w.etherman.GetRollupInfoByBlockRange(ctx, fromBlock, &toBlock)
		var lastBlock *types.Block = nil
		if err == nil && len(blocks) == 0 {
			lastBlock, err = w.etherman.EthBlockByNumber(ctx, toBlock)
		}
		duration := time.Since(now)
		result := newGenericAnswer(err, duration, typeRequestRollupInfo, &responseRollupInfoByBlockRange{blockRange, blocks, order, lastBlock})
		w.setStatus(ethermanIdle)
		ch <- result
	}()
```

synchronizer/l1_workers.go
```diff
@@ -0,0 +1,205 @@
...
+	err := funcRequest(worker, ctx, wg)
+	if err == nil {
+		log.Debugf("workers: worker started call:[%s]", requestStrForDebug)
+	} else {
+		log.Debugf("workers: worker started failed! call:[%s] failed err:[%s]", requestStrForDebug, err.Error())
+	}
+	return err
+}
+
+func (w *workers) _launchGoroutineForRoutingResponsesIfNeed() {
```
ARR552: As we discussed, check:
https://go.dev/doc/effective_go
https://google.github.io/styleguide/go/

synchronizer/l1_sync_orchestration.go
```diff
@@ -0,0 +1,155 @@
...
+			// Producer have finish
+			log.Warnf("orchestration: consumer have finish! this never have to happen, restarting it. Error:%s", err)
+			// to avoid respawn too fast it sleeps a bit
+			time.Sleep(time.Second)
+			l.launch_producer(invalidBlockNumber, hProducer, wg)
+		case err = <-chConsumer:
+			if err != nil {
+				log.Warnf("orchestration: consumer have finish with Error: %s", err)
+			} else {
+				log.Info("orchestration: consumer have finish ok")
```
ARR552:
```suggestion
				log.Info("orchestration: consumer has processed everything and is synced")
```

synchronizer/l1_sync_orchestration.go
```diff
@@ -0,0 +1,155 @@
...
+		select {
+		case err = <-hProducer:
+			// Producer have finish
+			log.Warnf("orchestration: consumer have finish! this never have to happen, restarting it. Error:%s", err)
+			// to avoid respawn too fast it sleeps a bit
+			time.Sleep(time.Second)
+			l.launch_producer(invalidBlockNumber, hProducer, wg)
+		case err = <-chConsumer:
+			if err != nil {
+				log.Warnf("orchestration: consumer have finish with Error: %s", err)
```
ARR552:
```suggestion
				log.Warnf("orchestration: consumer has finished with Error: %s", err)
```

synchronizer/l1_sync_orchestration.go
```diff
@@ -0,0 +1,155 @@
...
+
+func (l *l1SyncOrchestration) orchestrate(wg *sync.WaitGroup, hProducer chan error, chConsumer chan error) (*state.Block, error) {
+	// Wait a cond_var for known if consumer have finish
+	var err error
+	done := false
+	for !done {
+		select {
+		case err = <-hProducer:
+			// Producer have finish
+			log.Warnf("orchestration: consumer have finish! this never have to happen, restarting it. Error:%s", err)
```
ARR552:
```suggestion
			log.Warnf("orchestration: consumer has finished! this situation shouldn't happen, restarting it. Error:%s", err)
```

synchronizer/l1_sync_orchestration.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+func (l *l1SyncOrchestration) orchestrate(wg *sync.WaitGroup, hProducer chan error, chConsumer chan error) (*state.Block, error) {
+	// Wait a cond_var for known if consumer have finish
+	var err error
+	done := false
+	for !done {
+		select {
+		case err = <-hProducer:
+			// Producer have finish
```
ARR552:
```suggestion
			// Producer has finished
```

synchronizer/l1_sync_orchestration.go
```diff
@@ -0,0 +1,155 @@
...
+	"errors"
+	"sync"
+	"time"
+
+	"github.com/0xPolygonHermez/zkevm-node/log"
+	"github.com/0xPolygonHermez/zkevm-node/state"
+)
+
+/*
+This object is to coordinate the producer process and the consumer process.
```
ARR552:
```suggestion
This object is used to coordinate the producer and the consumer process.
```

synchronizer/l1_sync_orchestration.go
```diff
@@ -0,0 +1,155 @@
...
+		producerStarted: false,
+		consumerStarted: false,
+	}
+}
+
+func (l *l1SyncOrchestration) reset(startingBlockNumber uint64) {
+	log.Warnf("Reset L1 sync process to blockNumber %d", startingBlockNumber)
+	l.mutex.Lock()
+	if l.consumerStarted {
+		log.Warnf("orchestration: Undefined behaviour,  reset (%v) and consumer is running", startingBlockNumber)
```
ARR552:
```suggestion
		log.Warnf("orchestration: Undefined behaviour, reset (%v) and consumer is running", startingBlockNumber)
```

synchronizer/l1_rollup_info_consumer_statistics.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/0xPolygonHermez/zkevm-node/log"
+	"github.com/0xPolygonHermez/zkevm-node/synchronizer/metrics"
+)
+
+type ll1RollupInfoConsumerStatistics struct {
```
ARR552:
```suggestion
type l1RollupInfoConsumerStatistics struct {
```

synchronizer/l1_rollup_info_consumer.go
```diff
@@ -0,0 +1,202 @@
...
+
+	"github.com/0xPolygonHermez/zkevm-node/etherman"
+	"github.com/0xPolygonHermez/zkevm-node/log"
+	"github.com/0xPolygonHermez/zkevm-node/state"
+	"github.com/ethereum/go-ethereum/common"
+	types "github.com/ethereum/go-ethereum/core/types"
+)
+
+const (
+	errMissingLastBlock                                             = "consumer:the received rollupinfo have no blocks and need to fill last block"
```
ARR552:
```suggestion
	errMissingLastBlock                                             = "consumer:the received rollupinfo has no blocks and needs to fill last block"
```

synchronizer/l1_rollup_info_consumer.go
```diff
@@ -0,0 +1,202 @@
...
+	// Send a stop to the channel to stop the consumer when reach this point
+	l.chIncommingRollupInfo <- *newL1SyncMessageControl(eventStop)
+}
+
+func (l *l1RollupInfoConsumer) _Process(rollupInfo responseRollupInfoByBlockRange) (*state.Block, error) {
+	blocks := rollupInfo.blocks
+	order := rollupInfo.order
+	err := l.synchronizer.processBlockRange(blocks, order)
+	var lastEthBlockSynced *state.Block
+	if err != nil {
```
ARR552:
```suggestion
	var lastEthBlockSynced *state.Block
	err := l.synchronizer.processBlockRange(blocks, order)
	if err != nil {
```

synchronizer/l1_rollup_info_consumer.go
```diff
@@ -0,0 +1,202 @@
...
+		log.Error("consumer: Error processing block range: ", rollupInfo.blockRange, " err:", err)
+		return nil, err
+	}
+	if len(blocks) > 0 {
+		tmpStateBlock := convertEthmanBlockToStateBlock(&blocks[len(blocks)-1])
+		lastEthBlockSynced = &tmpStateBlock
+		logBlocks(blocks)
+	}
+	if len(blocks) == 0 {
+		fb := rollupInfo.lastBlockOfRange
```
ARR552:
```suggestion
		lastBlock := rollupInfo.lastBlockOfRange
```
or
```suggestion
		lb := rollupInfo.lastBlockOfRange
```

synchronizer/synchronizer.go
```diff
@@ -59,12 +60,14 @@ type ClientSynchronizer struct {
...
 	previousExecutorFlushID uint64
+	l1SyncOrchestration     *l1SyncOrchestration
 }
 
 // NewSynchronizer creates and initializes an instance of Synchronizer
 func NewSynchronizer(
 	isTrustedSequencer bool,
-	ethMan ethermanInterface,
+	ethMan EthermanInterface,
+	etherManForL1 []EthermanInterface,
```
ARR552, line 71: I think with `etherManForL1` is enough. If we use the legacy method, the array would have only one position. If not, would have many

synchronizer/synchronizer.go
```diff
@@ -259,14 +280,26 @@ func (s *ClientSynchronizer) Sync() error {
...
-			lastEthBlockSynced, err = s.syncBlocks(lastEthBlockSynced)
+			if s.l1SyncOrchestration != nil {
+				log.Infof("Syncing L1 blocks in parallel lastEthBlockSynced=%d", lastEthBlockSynced.BlockNumber)
+				lastEthBlockSynced, err = s.syncBlocksParallel(lastEthBlockSynced)
+			} else {
+				log.Infof("Syncing L1 blocks sequentially lastEthBlockSynced=%d", lastEthBlockSynced.BlockNumber)
+				lastEthBlockSynced, err = s.syncBlocksSequential(lastEthBlockSynced)
+			}
 			metrics.FullL1SyncTime(time.Since(startL1))
+			// Process of syncBlocksXXXX fails?
```
ARR552: could we delete it?

synchronizer/synchronizer.go
```diff
@@ -259,14 +280,26 @@ func (s *ClientSynchronizer) Sync() error {
...
 			if err != nil {
 				log.Warn("error syncing blocks: ", err)
 				lastEthBlockSynced, err = s.state.GetLastBlock(s.ctx, nil)
 				if err != nil {
 					log.Fatal("error getting lastEthBlockSynced to resume the synchronization... Error: ", err)
 				}
+				if s.l1SyncOrchestration != nil {
+					// If have failed execution and get starting point from DB, we must reset parallel sync to this point
```
ARR552, line 314: What do you want to say here?

synchronizer/l1_rollup_info_producer.go
```diff
@@ -0,0 +1,365 @@
...
+		log.Fatal(err)
+	}
+	err = l.workers.initialize()
+	if err != nil {
+		log.Fatal(err)
+	}
+	if l.syncStatus.needToRenewLastBlockOnL1() {
+		log.Infof("producer: Need a initial value for Last Block On L1, doing the request (maxRetries:%v, timeRequest:%v)",
+			maxRetriesForRequestnitialValueOfLastBlock, timeRequestInitialValueOfLastBlock)
+		//result := l.retrieveInitialValueOfLastBlock(maxRetriesForRequestnitialValueOfLastBlock, timeRequestInitialValueOfLastBlock)
```
ARR552, line 210: i think we could delete it

synchronizer/l1_syncstatus.go
```diff
@@ -0,0 +1,263 @@
...
+	s.mutex.Lock()
+	defer s.mutex.Unlock()
+	return s.lastBlockOnL1 == invalidLastBlock
+}
+
+func (s *syncStatus) verify() error {
+	if s.amountOfBlocksInEachRange == 0 {
+		return fmt.Errorf("SyncChunkSize must be greater than 0")
+	}
+	if s.lastBlockStoreOnStateDB == 0 {
```
ARR552, line 258:
```suggestion
	} else if s.lastBlockStoreOnStateDB == 0 {
```

ARR552 (changes requested):

cmd/run.go
joanestebanr: synchronizer still need a isntance of etherman because is used in some places like: 
* `func Sync`: VerifyGenBlockNumber, HeaderByNumber
* To get forkID info...
To avoid any conflict I think that is better to create a new instances

synchronizer/l1_worker.go
joanestebanr: Refactor the code to avoid the generic response

synchronizer/synchronizer.go
joanestebanr: I agree, but to recude the changes in `synchronizer.go` is better to separece both. In a long term it could be removed

synchronizer/synchronizer.go
joanestebanr: Instead of checking the config struct I see if is created the object, so I reduce the dependency with the config and avoid a situation that config says something but the object have not been created

synchronizer/l1_rollup_info_producer.go
joanestebanr: It's needed because in the first run need to ask for the las block on L1. 
- I renamed the function `needToRenewLastBlockOnL1` for `isSetLastBlockOnL1Value` that I think that make more sense

synchronizer/l1_workers.go
joanestebanr: I have check the provided documentation, and I see that the initial underscore in the name of the funcion is a bad idea, but I don't the way to say that this function need the mutex locked before the call...
Any idea?

joanestebanr: Renamed to XXXUnsafe() instead of _XXX()

synchronizer/l1_syncstatus.go
joanestebanr: I think that is better to keep without the else because the condition are not related

config/gen_json_schema.go
```diff
@@ -44,7 +44,7 @@ func NewNodeConfigJsonSchemaGenerater() ConfigJsonSchemaGenerater[Config] {
 	config_default_values, err := Default()
 	res.defaultValues = config_default_values
 	if err != nil {
-		panic("can't create default values for config file")
+		panic("can't create default values for config file: " + err.Error())
```
ARR552: why don't we use log.Fatal instead of panic?

synchronizer/config.go
```diff
@@ -12,4 +12,50 @@ type Config struct {
...
+	// TimeForShowUpStatisticsLog how ofter show a log with statistics (0 is disabled)
+	TimeForShowUpStatisticsLog types.Duration `mapstructure:"TimeForShowUpStatisticsLog"`
+	// TimeOutMainLoop is the timeout for the main loop of the L1 synchronizer when is not updated
+	TimeOutMainLoop types.Duration `mapstructure:"TimeOutMainLoop"`
+}
+
+// L1PerformanceCheckConfig Configuration for the consumer of rollup information from L1
+type L1PerformanceCheckConfig struct {
+	// AcceptableTimeWaitingForNewRollupInfo is the expected maximum time that the consumer
+	// could wait until new data are produced. If the time is greater it emmit a log to warn about
```
ARR552:
```suggestion
	// could wait until new data is produced. If the time is greater it emmit a log to warn about
```

synchronizer/l1_filter_send_orderer_results_to_synchronizer.go
```diff
@@ -0,0 +1,117 @@
...
+		}
+		newPendingResults = append(newPendingResults, s.pendingResults[j])
+	}
+	s.pendingResults = newPendingResults
+}
+
+func (s *filterToSendOrdererResultsToConsumer) setLastBlockOnSynchronizerCorrespondingLatBlockRangeSendUnsafe(lastBlock blockRange) {
+	newVaule := lastBlock.toBlock
+	log.Debug("Moving lastBlockSend from ", s.lastBlockOnSynchronizer, " to ", newVaule)
+	s.lastBlockOnSynchronizer = newVaule
```
ARR552:
```suggestion
	log.Debug("Moving lastBlockSend from ", s.lastBlockOnSynchronizer, " to ", lastBlock.toBlock)
	s.lastBlockOnSynchronizer = lastBlock.toBlock
```

synchronizer/l1_rollup_info_consumer.go
```diff
@@ -0,0 +1,225 @@
...
+)
+
+type configConsumer struct {
+	numIterationsBeforeStartCheckingTimeWaitinfForNewRollupInfoData int
+	acceptableTimeWaitingForNewRollupInfoData                       time.Duration
+}
+
+// synchronizerProcessBlockRangeInterface is the interface with synchronizer
+// to execute blocks. This interface is used to mock the synchronizer in the tests
+type synchronizerProcessBlockRangeInterface interface {
```
ARR552, line 35: Time ago, we agreed to define the interfaces in the consumer side. I think this interface (and maybe others) should be in interfaces.go

synchronizer/l1_rollup_info_consumer.go
ARR552: Can yo change the order here?

synchronizer/l1_rollup_info_consumer.go
ARR552: Can you modify this var?? It can be confused by forcedbatch

synchronizer/l1_rollup_info_consumer_statistics.go
ARR552: Can you fix this name?

synchronizer/l1_rollup_info_consumer_test.go
```diff
@@ -0,0 +1,80 @@
...
+)
+
+type consumerTestData struct {
+	sut      *l1RollupInfoConsumer
+	syncMock *synchronizerProcessBlockRangeMock
+	ch       chan l1SyncMessage
+	ctx      context.Context
+}
+
+func Test_Given_Consumer_When_ReceiveAFullSyncAndChannelIsEmpty_Then_StopOk(t *testing.T) {
```
ARR552, line 22: Lets try to Use the same function name format in the whole proyect

synchronizer/l1_rollup_info_producer.go
```diff
@@ -0,0 +1,401 @@
...
+// verify: test params and status without if not allowModify avoid doing connection or modification of objects
+func (l *l1RollupInfoProducer) verify() error {
+	return l.syncStatus.verify()
+}
+
+func (l *l1RollupInfoProducer) initialize() error {
+	err := l.verify()
+	if err != nil {
+		log.Fatal(err)
+	}
```
ARR552: If you returns an error in this function why you just log a fatal instead of return the error? We should return the error if possible and try to avoid fatals as much as possible

synchronizer/l1_rollup_info_producer.go
ARR552: I'm talking about the commented code. Can we delete it?? If it is necessary please uncomment it

synchronizer/l1_sync_orchestration.go
ARR552: Still pending

synchronizer/l1_syncstatus.go
```diff
@@ -0,0 +1,260 @@
...
+	return s.lastBlockOnL1 <= s.highestBlockRequested && s.errorRanges.len() == 0
+}
+
+// isNodeFullySynchronizedWithL1 returns true if the node is fully synchronized with L1
+// it means that all blocks until the last block on L1 are requested (maybe not finish yet) and there are no pending errors
+func (s *syncStatus) isNodeFullySynchronizedWithL1() bool {
+	s.mutex.Lock()
+	defer s.mutex.Unlock()
+	if s.lastBlockOnL1 == invalidLastBlock {
+		log.Warnf("Can't decide if it's fully sync because Last block on L1  is no valid: %v", s.lastBlockOnL1)
```
ARR552:
```suggestion
		log.Warnf("Can't decide if it's fully synced because last block on L1  is no valid: %d", s.lastBlockOnL1)
```

synchronizer/l1_syncstatus.go
```diff
@@ -0,0 +1,260 @@
...
+		s.setLastBlockOnL1Unsafe(lastBlock)
+		return response
+	}
+	if lastBlock == oldLastBlock {
+		response.extendedRange = nil
+		s.setLastBlockOnL1Unsafe(lastBlock)
+		return response
+	}
+	if lastBlock < oldLastBlock {
+		log.Warnf("new block [%v] is less than old block [%v]!", lastBlock, oldLastBlock)
```
ARR552:
```suggestion
		log.Warnf("new block [%d] is less than old block [%d]!", lastBlock, oldLastBlock)
```

synchronizer/l1_syncstatus.go
```diff
@@ -0,0 +1,260 @@
...
+		lastBlock = oldLastBlock
+		response.fullRange = blockRange{fromBlock: s.lastBlockStoreOnStateDB, toBlock: lastBlock}
+		return response
+	}
+	log.Debugf("onNewLastBlockOnL1(%v) final_status: %s", lastBlock, s.toStringBrief())
```
ARR552:
```suggestion
	log.Debugf("onNewLastBlockOnL1(%d) final_status: %s", lastBlock, s.toStringBrief())
```

synchronizer/l1_worker.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+	return result
+}
+
+type rollupInfoByBlockRangeResult struct {
+	blockRange blockRange
+	blocks     []etherman.Block
+	order      map[common.Hash][]etherman.Order
+	// If there are no blocks in this range get get the last one
+	// so it could be nil if there are blocks.
```
ARR552:
```suggestion
	// If there are no blocks in this range, it gets the last one
	// so it could be nil if there are no blocks.
```

synchronizer/synchronizer_test.go
```diff
@@ -40,52 +40,66 @@ type mocks struct {
...
 	genesis, cfg, m := setupGenericTest(t)
-	syncInterface, err := NewSynchronizer(false, m.Etherman, m.State, m.Pool, m.EthTxManager, m.ZKEVMClient, nil, *genesis, *cfg)
+	ethermanForL1 := []EthermanInterface{m.Etherman}
+	syncInterface, err := NewSynchronizer(false, m.Etherman, ethermanForL1, m.State, m.Pool, m.EthTxManager, m.ZKEVMClient, nil, *genesis, *cfg)
 	require.NoError(t, err)
 	sync, ok := syncInterface.(*ClientSynchronizer)
 	require.EqualValues(t, true, ok, "Can't convert to underlaying struct the interface of syncronizer")
 	lastBatchNumber := uint64(10)
-	batch10With1Tx := createBatch(t, lastBatchNumber, 1)
+	//batch10With1Tx := createBatch(t, lastBatchNumber, 1)
```
ARR552: please, remove the commented code

synchronizer/synchronizer_test.go
```diff
@@ -668,68 +700,72 @@ func expectedCallsForsyncTrustedState(t *testing.T, m *mocks, sync *ClientSynchr
 		Once()
 
 	m.State.
-		On("BeginStateTransaction", sync.ctx).
+		On("BeginStateTransaction", mock.Anything).
```
ARR552, line 702: I don't fully understand this change

ARR552 (changes requested):

synchronizer/l1_sync_orchestration.go
```diff
@@ -129,16 +129,16 @@ func (l *l1SyncOrchestration) orchestrate(wg *sync.WaitGroup, hProducer chan err
 	for !done {
 		select {
 		case err = <-hProducer:
-			// Producer have finish
-			log.Warnf("orchestration: consumer have finish! this never have to happen, restarting it. Error:%s", err)
+			// Producer have finished
```
ARR552:
```suggestion
			// Producer has finished
```

ARR552 (changes requested):

synchronizer/synchronizer_test.go
joanestebanr: Do you mind the change of `sync.ctx` to `mock.Anything`? Is because I think that It's not important which ctx are you sending, and forcing to be this one could make that it breaks in the future if we decide to use another one.

synchronizer/l1_rollup_info_consumer_test.go
joanestebanr: I have adopted the format: `Test_Given_XXX_When_YYY_Then_ZZZ` because, from my point of view, it's more easy to read. 
But if you disagree I can rename to this format `TestGivenXXWhenYYYThenZZZ`

synchronizer/l1_rollup_info_consumer.go
joanestebanr: I think that is better here, because this is the interface require by consumer. Another object could need another interface over the same object

config/gen_json_schema.go
joanestebanr: no reason, changed!

synchronizer/l1_rollup_info_consumer_test.go
joanestebanr: We agree to decide later

synchronizer/l1_rollup_info_producer.go
joanestebanr: changed!

ARR552 (approved):

PR #2517, opened 2023-09-06
Add sequencer data to stream server

### LINKED ISSUE #2515: Add Stream Server to Sequencer

ToniRamirezM:
Closes #2515

### What does this PR do?

Adds sequencer data to stream server.

New config params:

```
[Sequencer.StreamServer]
      Port = 6900
      Filename = "/datastreamer/datastream.bin"
      Enabled = true
```

### Reviewers

Main reviewers:

- @dpunish3r 
- @Psykepro 
- @agnusmor


config/default.go
```diff
@@ -110,6 +110,9 @@ MaxTxLifetime = "3h"
 		ByteGasCost = 16
 		MarginFactor = 1
 		Enabled = false
+	[Sequencer.StreamServer]
```
agnusmor, line 113: We can add an "Enabled" parameter to specify if the StreamServer must be initiated or not. Instead to check if Port and Filename are set. I think in this way is more clear when the StreamServer is enabled or not

sequencer/datastream.go
```diff
@@ -0,0 +1,48 @@
...
+	L2BlockNumber  uint64
+	Timestamp      time.Time
+	GlobalExitRoot common.Hash
+	Coinbase       common.Address
+}
+
+// Encode returns the encoded L2Block as a byte slice
+func (b L2Block) Encode() []byte {
+	const size = int(unsafe.Sizeof(L2Block{}))
+	return (*(*[size]byte)(unsafe.Pointer(&b)))[:]
```
ARR552: Are you sure about using this package right?? Isn't there other way to do it? It is a bit scary
Maybe this is helpful https://stackoverflow.com/questions/16330490/in-go-how-can-i-convert-a-struct-to-a-byte-array

ToniRamirezM: This is the easiest and faster way to convert a struct into a []byte. I know unsafe sounds "unsafe" but it is there for a reason and it only means YOU must take care of making it safe, in the same way unsafe works in Rust.

IMHO in this case where the only thing we use it for is to get a slice of consecutive memory into a slice. It is safe. There is no arithmetic involved or anything harmful.

Also there are parts of the code in the library that use reflection and it is though to be used also with real (unsafe) pointers.

config/default.go
ToniRamirezM: If you look closely you will see the Config type for this section is a datastreamer.Config It means it comes from the library so it only has information needed for the data stream server to work. It makes no sense adding a "enabled" parameter into de library. This could be solved by creating a new Config in the Node with aggregation, adding the Enabled boolean, put IMHO I think is easier to just do that if I did.

sequencer/dbmanager.go
```diff
@@ -200,9 +202,59 @@ func (d *dbManager) StoreProcessedTxAndDeleteFromPool(ctx context.Context, tx tr
 		}
 	}
 
+	// Send tx data to data stream server
+	if d.streamServer != nil {
+		l2BlochHeader, err := d.GetLastL2BlockHeader(ctx, dbTx)
```
agnusmor: The d.state.StoreTransaction function can return the header of the L2BLock added, with this we avoid to do another query in GetLastL2BlockHeader to get it and improve the performance

sequencer/dbmanager.go
```diff
@@ -200,9 +202,59 @@ func (d *dbManager) StoreProcessedTxAndDeleteFromPool(ctx context.Context, tx tr
...
+
+		_, err = d.streamServer.AddStreamEntry(EntryTypeL2Tx, l2Transaction.Encode())
+		if err != nil {
+			return err
+		}
+	}
+
 	err = dbTx.Commit(ctx)
 	if err != nil {
+		err2 := d.streamServer.RollbackAtomicOp()
```
agnusmor: In line 244 I see `err = dbTx.Commit(ctx)` but I don't see where we do the `dbTx.Rollback` in case we get an error

agnusmor: My proposal for this "commit" block is first do the CommitAtomicOp() if it goes fine then to the dbTx.Commit, if not do the dbTx.Rollback. I think that if all db queries are executed without errors the risk to have an error when doing the dbTx.Commit is very low and maybe there is some more risk when doing the CommitAtomicOp()
```
if d.streamServer != nil {
   err = d.streamServer.CommitAtomicOp()
   if err != nil {
      log.Error("failed to commit atomic op: %v", err)
      err2 = dbTx.Rollback(ctx)
      if err2 != nil {
         return err2
      }      
      return err
   }
   err = dbTx.Commit(ctx)
   if err != nil {
      return err
   }      
} else {
   err = dbTx.Commit(ctx)
   if err != nil {
      return err
   }      
}
```

config/default.go
ToniRamirezM: I finally added the enabled config param

sequencer/dbmanager.go
```diff
@@ -175,14 +177,18 @@ func (d *dbManager) StoreProcessedTxAndDeleteFromPool(ctx context.Context, tx tr
...
-	err = d.state.StoreTransaction(ctx, tx.batchNumber, tx.response, tx.coinbase, uint64(tx.timestamp.Unix()), dbTx)
+	l2BlochHeader, err := d.state.StoreTransaction(ctx, tx.batchNumber, tx.response, tx.coinbase, uint64(tx.timestamp.Unix()), dbTx)
 	if err != nil {
 		return err
 	}
 
 	// Update batch l2 data
 	batch, err := d.state.GetBatchByNumber(ctx, tx.batchNumber, dbTx)
 	if err != nil {
+		err2 := dbTx.Rollback(ctx)
```
agnusmor, line 188: This rollback should not be done in line 182? as in the line 182 we are returning with an open db transaction (started in line 175)

sequencer/dbmanager.go
```diff
@@ -196,13 +202,67 @@ func (d *dbManager) StoreProcessedTxAndDeleteFromPool(ctx context.Context, tx tr
...
+			}
 			return err
 		}
 	}
 
-	err = dbTx.Commit(ctx)
-	if err != nil {
-		return err
+	// Send tx data to data stream server
+	if d.streamServer != nil {
```
agnusmor, line 215: I think this code does the same:
```
if d.streamServer != nil {
		l2Block := DSL2Block{
			BatchNumber:    tx.batchNumber,
			L2BlockNumber:  l2BlochHeader.Number.Uint64(),
			Timestamp:      tx.timestamp,
			GlobalExitRoot: batch.GlobalExitRoot,
			Coinbase:       tx.coinbase,
		}

		l2Transaction := DSL2Transaction{
			BatchNumber:                 batch.BatchNumber,
			EffectiveGasPricePercentage: uint8(tx.response.EffectivePercentage),
			IsValid:                     1,
			EncodedLength:               uint32(len(txData)),
			Encoded:                     txData,
		}

		err = d.streamServer.StartAtomicOp()
		if err != nil {
			return err
		}

		_, err = d.streamServer.AddStreamEntry(EntryTypeL2Block, l2Block.Encode())
		if err != nil {
			return err
		}

		_, err = d.streamServer.AddStreamEntry(EntryTypeL2Tx, l2Transaction.Encode())
		if err != nil {
			return err
		}

		err = d.streamServer.CommitAtomicOp()
		if err != nil {
			log.Error("failed to commit atomic op: %v", err)
			err2 := dbTx.Rollback(ctx)
			if err2 != nil {
				log.Errorf("failed to rollback dbTx when committing atomic op that gave err: %v. Rollback err: %v", err2, err)
			}
			return err
		}
	} 

	err = dbTx.Commit(ctx)
	if err != nil {
		return err
	}
```

config/default.go
```diff
@@ -110,6 +110,10 @@ MaxTxLifetime = "3h"
...
```
agnusmor, line 113: StreamServer or DataStreamer? As we have used DataStreamerCfg for the config struct

sequencer/dbmanager.go
ToniRamirezM: If we do the error check this way, we would return without rolling back the stream transaction.

config/environments/local/local.node.config.toml
```diff
@@ -93,6 +93,10 @@ MaxTxLifetime = "3h"
 		ByteGasCost = 16
 		MarginFactor = 1
 		Enabled = false
+	[Sequencer.StreamServer]
+		Port = 0
```
Psykepro, line 97: Is this `Port = 0` here correct?

ToniRamirezM: It is disabled / not configured, so it does not matter

agnusmor (approved):

PR #2521, opened 2023-09-07
feat(db): l2block add block_hash index for method getBlockByHash

zibuyu28:
Closes #<issue number>.

### What does this PR do?

Optimize the rpc method eth_getBlockByHash method

### Reviewers

Main reviewers:

- @John
- @Doe

Codeowner reviewers:

- @-Alice
- @-Bob


tclemos:
Hi, @zibuyu28, thanks for your contribution.

Migration files can't be changed, a new one needs to be created in order to add this new index.

Also, I'm working on a task at the moment that aims to improve some indexes, in which we will also add the hash index for l2 blocks.

I'll close this PR to avoid conflicts, but don't worry, this index will be added in a future version.

PR #2524, opened 2023-09-07
Use ZKEVM_NETWORK variable to specify location of config files in docker compose file

agnusmor:
### What does this PR do?

Changes the docker-compose file used when deploying a node ([production-setup.md](https://github.com/0xPolygonHermez/zkevm-node/blob/develop/docs/production-setup.md)) to use the variable $ZKEVM_NETWORK when specifying the location of the config files. Before it used always the "testnet" path for the location of the config files and this doesn't works when deploying the node for mainnet

### Reviewers

Main reviewers:

@ToniRamirezM 
@ARR552 
@joanestebanr


ARR552 (approved):

ToniRamirezM (approved):

ToniRamirezM:
Maybe this should also be added to v0.3.1

PR #2536, opened 2023-09-12
avoid halt syncronization in permissionless nodes

### LINKED ISSUE #2535: Avoid block Permissionless sync

ARR552:
Closes #2535

### What does this PR do?

Removes the halt for permissionless nodes and adds some logs

### Reviewers

- @joanestebanr 
- @agnusmor

### DIFF
--- a/synchronizer/synchronizer.go
+++ b/synchronizer/synchronizer.go
@@ -1320,14 +1320,12 @@ func (s *ClientSynchronizer) processTrustedBatch(trustedBatch *types.Batch, dbTx
 				if isBatchClosed {
 					//Sanity check
 					if s.trustedState.lastStateRoot != nil && trustedBatch.StateRoot != *s.trustedState.lastStateRoot {
-						if trustedBatchL2Data.Hex() != "0x"+common.Bytes2Hex(batches[0].BatchL2Data) {
-							log.Errorf("batch %d, different batchL2Datas (trustedBatchL2Data: %s, batches[0].BatchL2Data: %s). Decoded txs are len(storedTxs): %d, len(syncedTxs): %d", uint64(trustedBatch.Number), trustedBatchL2Data.Hex(), "0x"+common.Bytes2Hex(batches[0].BatchL2Data), len(storedTxs), len(syncedTxs))
-							for _, tx := range storedTxs {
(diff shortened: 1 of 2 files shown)


joanestebanr (approved):

synchronizer/synchronizer.go
```diff
@@ -1309,7 +1320,17 @@ func (s *ClientSynchronizer) processTrustedBatch(trustedBatch *types.Batch, dbTx
 				if isBatchClosed {
 					//Sanity check
 					if s.trustedState.lastStateRoot != nil && trustedBatch.StateRoot != *s.trustedState.lastStateRoot {
-						s.halt(s.ctx, fmt.Errorf("stateRoot calculated (%s) is different from the stateRoot (%s) received during the trustedState synchronization", *s.trustedState.lastStateRoot, trustedBatch.StateRoot))
+						if trustedBatchL2Data.Hex() != "0x"+common.Bytes2Hex(batches[0].BatchL2Data) {
```
agnusmor: I think is better to log always trustedBatchL2Data, batches[0].BatchL2Data and contents of storedTxs and syncedTxs, not only to log it if trustedBatchL2Data != batches[0].BatchL2Data, in this way we will have all the log info whatever is the scenario

synchronizer/synchronizer.go
```diff
@@ -1361,7 +1382,7 @@ func (s *ClientSynchronizer) processTrustedBatch(trustedBatch *types.Batch, dbTx
 		// Update batchL2Data
 		err := s.state.UpdateBatchL2Data(s.ctx, batches[0].BatchNumber, trustedBatchL2Data, dbTx)
 		if err != nil {
-			log.Errorf("error opening batch %d", trustedBatch.Number)
+			log.Errorf("error opening batch %v", trustedBatch.Number)
```
agnusmor: Why we use here %v and in lines 1346 and 1349 we use %d with the uint64 casting? I think we can use the same "solution" in all the places

ARR552: In this case the type it is a custom one called "types.ArgUint64". Not exactly uint64.  %v is the generic so will work always, thats why I decided to use %v

PR #2552, opened 2023-09-15
update contracts image

ToniRamirezM:
updates contracts image


ARR552 (approved):

ARR552:
Using the latest opcode of the smc, Push0 compatibility is required and the unit tests of the etherman don't work because of the backends. I have opened an issue on geth repo (https://github.com/ethereum/go-ethereum/issues/28144) where I explain the problem

joanestebanr (approved):

PR #2572, opened 2023-09-20
Feature/jrpc log limit

### LINKED ISSUE #2502: review and add limits to eth_getLogs and eth_newFilter

tclemos:
Closes #2502.

### What does this PR do?

Add limits when creating new filters or getting logs from state.
New configurations were created to allow the limits to be defined by the node, but not configuration changes are required if the default values are enough for your use case.

`MaxLogsCount` is a configuration to set the maximum number of logs that can be returned
in a single call to the state, if zero it means no limit

`MaxLogsBlockRange` is a configuration to set the max range for block number when querying TXs
logs in a single call to the state, if zero it means no limit

### Default Values: 

```toml
[RPC]
MaxLogsCount = 10000
MaxLogsBlockRange = 10000
```

### Reviewers

@ToniRamirezM 
@Psykepro


Psykepro (approved):

Psykepro (commented):
LGTM just fix the conflicts

db/migrations/state/0010.sql
```diff
@@ -0,0 +1,19 @@
+-- +migrate Up
```
ARR552: Can you add the test for this migration?

tclemos: yes, done!

ARR552 (approved):

PR #2575, opened 2023-09-20
add random ContextId when calling executor.ProcessBatch

### LINKED ISSUE #2564: update executor proto to context_id

tclemos:
Closes #2564.

### What does this PR do?

add the `ContextId` to the `executor proto file` in order to allow the executor to track all the logs of a specific call from the node.

### Reviewers

@agnusmor 
@ToniRamirezM


state/state_test.go
```diff
@@ -958,6 +962,7 @@ func TestExecutorRevert(t *testing.T) {
 //		Db:               genesisDB,
 //		ChainId:          stateCfg.ChainID,
 //		ForkId:           forkID,
+//		ContextId:        uuid.NewString(),
```
ARR552, line 965: Why this test is commented?

state/state_test.go
```diff
@@ -2180,7 +2192,8 @@ func TestExecutorGasRefund(t *testing.T) {
 		EthTimestamp:     uint64(time.Now().Unix()),
 		UpdateMerkleTree: 1,
 		ChainId:          stateCfg.ChainID,
-		ForkId: 		 forkID,
+		ForkId:           forkID,
```
ARR552, line 2195: Do you think this test can be uncommented?? I think the executor is currently returning the correct gas refund

state/state_test.go
tclemos: I don't know, I just added the new field in case we uncomment this in the future.

state/state_test.go
tclemos: I don't know, maybe it would be better to create an issue to review commented tests in the state and try to find in the history why they are commented at the moment. I just added the fields to the commented tests to help us in case we want to uncomment them in the future.

ARR552 (approved):

PR #2593, opened 2023-09-22
add unit test to validate jRPC max request limit per ip and second

tclemos:
Closes #<issue number>.

### What does this PR do?

add unit test to validate jRPC max request limit per ip and second

### Reviewers

@ToniRamirezM 
@Psykepro


jsonrpc/server_test.go
```diff
@@ -439,3 +440,83 @@ func TestRequestValidation(t *testing.T) {
...
+				// registers in the wait group a request was executed and has returned
+				wg.Done()
+			}
+		}()
+	}
+
+	// add jobs to notify workers accordingly to the number
+	// of requests the test wants to send to the server
+	for i := 0; i < numberOfRequests; i++ {
+		jobs <- i
```
ToniRamirezM, line 520: `Job` is not a buffered channel, so these requests will not be concurrent. Is that the desired behaviour?

ToniRamirezM (approved):

jsonrpc/server_test.go
tclemos: good catch, I'm updating the test to buffer the channel and to assert the number of requests limited by the server

PR #2600, opened 2023-09-29
fix deadlock

### LINKED ISSUE #2595: Synchronizer L1 parallel: get stuck after a new ForkId
# Rationale
Fix deadlock

# Implementation

joanestebanr:
Closes #2595 

### What does this PR do?
Fix a problem of deadlock when a forkid produce a reset of state

### Reviewers
- @ARR552


synchronizer/ext_control.go
```diff
@@ -0,0 +1,112 @@
+package synchronizer
```
ARR552, line 1: Can you add somewhere a readme with an example of the file content to send commands?

synchronizer/ext_control.go
```diff
@@ -0,0 +1,112 @@
...
+	"os"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/0xPolygonHermez/zkevm-node/log"
+)
+
+const (
+	externalControlFilename = "/tmp/synchronizer_in"
```
ARR552, line 15: if this path must be created and filled with the commands, why don't we convert it in a config param?
I understand that if we runt the docker, first we should copy the file into the container. Am I right?

synchronizer/l1_rollup_info_consumer.go
```diff
@@ -14,7 +14,7 @@ import (
 )
 
 const (
-	minNumIterationsBeforeStartCheckingTimeWaitinfForNewRollupInfoData = 20
+	minNumIterationsBeforeStartCheckingTimeWaitinfForNewRollupInfoData = 5
```
ARR552:
```suggestion
	minNumIterationsBeforeStartCheckingTimeWaitingForNewRollupInfoData = 5
```

synchronizer/l1_rollup_info_consumer.go
```diff
@@ -92,7 +92,7 @@ func (l *l1RollupInfoConsumer) step() error {
 		}
 		if rollupInfo.ctrlIsValid {
 			err = l.processIncommingRollupControlData(rollupInfo.ctrl)
-			if err != nil && !errors.Is(err, errConsumerStoppedBecauseIsSynchronized) {
+			if err != nil && !errors.Is(err, errConsumerStoppedBecauseIsSynchronized) && !errors.Is(err, errConsumerStopped) {
 				log.Error("consumer: error processing package.ControlData. Error: ", err)
```
ARR552, line 96: In this case what happens?? would it retries to get the package in the next iteration?

synchronizer/l1_rollup_info_producer.go
```diff
@@ -164,13 +164,14 @@ func newL1DataRetriever(cfg configProducer, ethermans []EthermanInterface, outgo
 // ResetAndStop: reset the object and stop the current process. Set first block to be retrieved
 func (l *l1RollupInfoProducer) ResetAndStop(startingBlockNumber uint64) {
 	log.Infof("producer: Reset L1 sync process to blockNumber %d st=%s", startingBlockNumber, l.toStringBrief())
-	l.mutex.Lock()
-	defer l.mutex.Unlock()
 	log.Debugf("producer: Reset(%d): stop previous run (state=%s)", startingBlockNumber, l.status.String())
-	l.stopUnsafe()
+	l.Stop()
```
ARR552, line 168: Is the stop called from somewhere without needing the lock?? Could we just have the stop() function, instead the stop() and stopUnsafe()???

synchronizer/l1_sync_orchestration.go
```diff
@@ -58,20 +60,15 @@ func newL1SyncOrchestration(ctx context.Context, producer l1RollupProducerInterf
 
 func (l *l1SyncOrchestration) reset(startingBlockNumber uint64) {
 	log.Warnf("Reset L1 sync process to blockNumber %d", startingBlockNumber)
-	l.mutex.Lock()
-	defer l.mutex.Unlock()
-	comuserWasRunning := l.consumerStarted
-	if comuserWasRunning {
+	if l.isStarted {
```
ARR552:
```suggestion
	if l.isRunning {
```
Just to use the same nomenclature than in line 53 and 54.

synchronizer/l1_sync_orchestration_test.go
```diff
@@ -31,8 +31,8 @@ func TestGivenOrquestrationWhenHappyPathThenReturnsBlockAndNoErrorAndProducerIsR
 	returnedBlock, err := sut.start()
 	require.NoError(t, err)
 	require.Equal(t, block, *returnedBlock)
-	require.Equal(t, true, sut.producerStarted)
```
ARR552, line 34: Why sut?? Shouldn't it be called sot??? (**S**etup**O**rchestration**T**est)

ARR552 (changes requested):

synchronizer/ext_control.go
joanestebanr: in fact no, the code create the file and if exists skip all contents (is for live commands). I think that is not worth to allow configure it becuase is for debug/test propouses. If you want to use inside a docker you can map the volume of /tmp/ or the file by itself

synchronizer/l1_sync_orchestration_test.go
joanestebanr: sut stand for Subject Under Test... is a common way to name the object that you are testing

synchronizer/l1_sync_orchestration.go
joanestebanr: good point, done

synchronizer/l1_rollup_info_producer.go
joanestebanr: Yes, the function stopUnsafe is already deleted!

synchronizer/l1_rollup_info_consumer.go
joanestebanr: It returns to main Sync function and depending of the lastblock returned is going to reset the synchronization process. This if is just to log an error or not. If the consumer finish by a controled situation we dont log a error

ARR552 (approved):

PR #2616, opened 2023-10-04
Feature/2595 l1 sync deadlock

### LINKED ISSUE #2595: Synchronizer L1 parallel: get stuck after a new ForkId

joanestebanr:
Closes #2595
### What does this PR do?
- It keep running producer instead of stopping to do the reset
- Fix some errors due this change of behaviour

### Reviewers

Main reviewers:
@ARR552


synchronizer/l1_syncstatus.go
```diff
@@ -60,7 +60,7 @@ func (s *syncStatus) reset(lastBlockStoreOnStateDB uint64) {
 	s.lastBlockStoreOnStateDB = lastBlockStoreOnStateDB
 	s.highestBlockRequested = lastBlockStoreOnStateDB
 	s.processingRanges = newLiveBlockRanges()
-	s.lastBlockOnL1 = invalidLastBlock
+	//s.lastBlockOnL1 = invalidLastBlock
```
ARR552, line 63: Isn't this code needed?

synchronizer/l1_workers.go
```diff
@@ -84,15 +91,15 @@ func (w *workers) initialize() error {
...
+	log.Infof("workers: stopping workers %s", w.String())
 	for i := range w.workers {
 		wd := &w.workers[i]
 		if !wd.worker.isIdle() {
 			log.Debugf("workers: stopping worker[%d] %s", i, wd.String())
 		}
 		wd.ctx.cancel()
 	}
-	w.waitFinishAllWorkers()
+	//w.waitFinishAllWorkers()
```
ARR552, line 102: Why don't you wait here?

ARR552 (approved):

synchronizer/l1_syncstatus.go
joanestebanr: I was thinking about that... previously I discard the lastBlockOnL1 after a reset to force to retrieve it again, but I think that is not needed, because the block if change is going to increase never decrease. So it would be retrieve as the regular flow of the code

synchronizer/l1_workers.go
joanestebanr: Because you don't need to wait to finish all workers, the workers are going to be canceled eventually and you can go on with the regular flow. Just to speed up process

PR #2620, opened 2023-10-05
Bring validium home

### LINKED ISSUE #2187: Remove unnecessary interface functions from sequencer
After segregating the `sequencesender` from the `sequencer`, there are a bunch of methods that are no longer required

arnaubennassar:
# 🏘️ Bring the Validium home 🏘️ 

I will add comments on the parts of the code that I think are more sensible and need to be reviewed with attention

### Etherman

- Added the validium (`polygonzkevmvalidium` and `cdkdatacommittee`) contracts, and renamed the previous one: `polygonzkevm` => `polygonzkevmrollup`. The client struct now has this 3 mentioned contracts (instead of only the `ZkEVM`)
- Added methods to interact with the DAC contract
- Splitted `EstimateGasSequenceBatches` into `EstimateGasSequenceBatchesRollup` and `EstimateGasSequenceBatchesValidium`.
- Splitted `BuildSequenceBatchesTxData` into `BuildSequenceBatchesRollupTxData` and `BuildSequenceBatchesValidiumTxData`.
- Splitted `decodeSequences` into `decodeSequencesRollup` and `decodeSequencesValidium`.
- Modified `SequencedBatch` so it does not inherit the type from the smart contract, as now it's used for two different contracts

### E2E tests

- Added a mechanism to run tests twice for validium/rollup (if the test is somehow relevant for the consensus) or only once (using rollup by default). A rule of thumb is that if the test only involves the trusted state, it should only be run once. This involves changes on docker-compose and makefile.
- Renamed some of the config files in order to:
    - Get rid of the redundant `.test.config` on the names (this files are under `test/config` directory 🙃 )
    - Separated config files for rollup and validium when needed (node and genesis config files)
- Added an E2E test for the DAC, which required to create to new services on the docker compose
- Added a L1 mock network service for the validium

### Sequence Sender

- Now it needs the private key of the sequencer, to sign messages to be sent to the DAC. As a side effect, the signature of `etherman.LoadAuthFromKeyStore` is changed to return the raw private key
- Add methods to interact with the DAC (request to sign data before sending to L1)
- Decide how/when to send data to L1 based on validium vs rollup

### Policy

@christophercampbell please check that everything is fine here

- Added a new mechanism to block interactions with RPC based on address
- Added command in CLI to interact
- Added table on the pool schema to store the policy rules
- Added necessary methods and errors on the `pool` and `pool/pgstorage`

### Config

- Added `IsRollup` bool for [`Synchronizer`, `SequenceSender`, `Etherman`]
- Added `MaxBatchesForL1` for  `SequenceSender`. This is relevant for both rollup and validium, although it's unlikely to be relevant fro rollup
- Added `NetworkConfig.L1Config.DataCommitteeAddr` to the genesis file, to be able to load the address of the DAC contract

### Sequencer

- Remove unnecessary interfaces, close #2187 (@ToniRamirezM , many weeks ago I said that I will do that, and you asked me to add you as a reviewer...)

### Synchronizer

- Add methods to (in the case of validium) get batch data from:
    - Local DB (in case already synced from trusted state
    - Sequencer
    - DAC
- When syncing the virtual state, if is validium, get batch data using the previously described methods (get hash of the batch data from L1, try to get data that is the image hash from [DB, sequencer, DAC])


pool/pool.go
```diff
@@ -81,6 +81,14 @@ func NewPool(cfg Config, batchConstraintsCfg state.BatchConstraintsCfg, s storag
 		gasPricesMux:            new(sync.RWMutex),
 	}
 
+	p.refreshBlockedAddresses()
+	go func(cfg *Config, p *Pool) {
+		for {
+			time.Sleep(cfg.IntervalToRefreshBlockedAddresses.Duration)
+			p.refreshBlockedAddresses()
+		}
+	}(&cfg, p)
```
arnaubennassar: This needs to be reviewed @tclemos

sequencesender/sequencesender.go
```diff
@@ -107,10 +114,27 @@ func (s *SequenceSender) tryToSendSequence(ctx context.Context, ticker *time.Tic
 	metrics.SequencesSentToL1(float64(sequenceCount))
 
 	// add sequence to be monitored
-	to, data, err := s.etherman.BuildSequenceBatchesTxData(s.cfg.SenderAddress, sequences, s.cfg.L2Coinbase)
-	if err != nil {
-		log.Error("error estimating new sequenceBatches to add to eth tx manager: ", err)
-		return
+	var (
```
arnaubennassar, line 117: The changes on this file need to be reviewed carefully @tclemos @ToniRamirezM

synchronizer/synchronizer.go
```diff
@@ -872,6 +882,14 @@ func (s *ClientSynchronizer) processSequenceBatches(sequencedBatches []etherman.
...
 	}
 	for _, sbatch := range sequencedBatches {
+		if !s.cfg.IsRollup {
+			// Recover data that is not on-chain for Validiums
+			batchL2Data, err := s.getBatchL2Data(sbatch.BatchNumber, sbatch.TransactionsHash)
+			if err != nil {
+				return err
+			}
+			sbatch.Transactions = batchL2Data
+		}
```
arnaubennassar, line 893: @ARR552 @joanestebanr please review this

ToniRamirezM (approved):

sequencesender/sequencesender.go
ToniRamirezM: LGTM

pool/pool.go
tclemos: this code needs to be removed from this part, it was moved to here: https://github.com/0xPolygonHermez/zkevm-node/blob/1b8ba45686295dad14891a36bb37811c2af139c5/pool/pool.go#L110-L118

config/default.go
```diff
@@ -54,6 +54,7 @@ GlobalQueue = 1024
 	MaxConns = 200
 
 [Etherman]
+IsRollup = true
```
tclemos: I see `Etherman`, `Synchronizer` and `SequenceSender` have a new config parameter called `IsRollup`, do you think we could centralize and share this config in the root config the same way as `IsTrustedSequencer`?

This way we would have a single point to define this and this would be replicated to all the components running this configuration.

etherman/etherman.go
tclemos, line 1: Probably the best approach to the Etherman interface would be having a client interface, a shared implementation for everything that is common between clients, and two different implementations, one for Rollup and another one for Validium. This would avoid some `If Rollup` or `If Validium`

arnaubennassar: Something like this:

```go
// etherman/etherman.go

type ethermanBase struct {/*...*/}

// implementation of all common methods under the ethermanBase struct

// etherman/rollup.go and etherman/validium.go
type Etherman{Rollup/Validum} struct {
    ethermanBase
    // ...
}

func New{Rollup/Validium}() {}

// implementation of the rollup/validium specifics under EthermanRollup and EthermanValidium
```
?

config/default.go
arnaubennassar: The problem of doing such a thing is that then all the constructors will need a new parameter `isRollup` and this will need to be stored on the root of each struct (or elsewhere, but not in the config). I still thing that it's a good idea to apply the change since this will make the config file better which is already complex enough, and it's what the users will see at the end of the day

test/e2e/permissionlessrpc_test.go
```diff
@@ -21,7 +22,9 @@ func TestPermissionlessJRPC(t *testing.T) {
 	// Initial setup:
 	// - permissionless RPC + Sync
 	// - trusted node with everything minus EthTxMan (to prevent the trusted state from being virtualized)
-	if testing.Short() {
+
+	os.Setenv(operations.TestConcensusENV, operations.Validium)
```
tclemos: should we force validium here?

config/default.go
tclemos: you can maintain the config structure you already created, just create a new field in the root config with the `mapstrucutre` tag, remove all the `mapstructure` tags from the internal configs and, do the set mapping in the `run.go`, here is an example:

state configs inherited from other components configs:
https://github.com/0xPolygonHermez/zkevm-node/blob/496d4a13f76b69fbb356a24019d336cccbf26d33/cmd/run.go#L482-L492

RPC configs with `mapstructure` tags to read data from config:
https://github.com/0xPolygonHermez/zkevm-node/blob/496d4a13f76b69fbb356a24019d336cccbf26d33/jsonrpc/config.go#L50-L56

State configs without `mapstrucutre` tags to be set via `run.go`:
https://github.com/0xPolygonHermez/zkevm-node/blob/496d4a13f76b69fbb356a24019d336cccbf26d33/state/config.go#L37-L43

etherman/etherman.go
tclemos: Yes, and the caller would have these 2 instances, injecting the correct one for the methods calling the client methods ... this would make the internal code agnostic to the client and the code cleaner and extensible if needed.

pool/pool.go
arnaubennassar: code removed 🖖

test/e2e/permissionlessrpc_test.go
arnaubennassar: fixed

synchronizer/synchronizer.go
```diff
@@ -872,6 +883,14 @@ func (s *ClientSynchronizer) processSequenceBatches(sequencedBatches []etherman.
 		return nil
 	}
 	for _, sbatch := range sequencedBatches {
+		if !s.cfg.IsRollup {
+			// Recover data that is not on-chain for Validiums
+			batchL2Data, err := s.getBatchL2Data(sbatch.BatchNumber, sbatch.TransactionsHash)
+			if err != nil {
+				return err
+			}
+			sbatch.Transactions = batchL2Data
```
ARR552, line 892: If it is a validium I assume the etherman will leave the value empty right?

ARR552 (approved):
The PR make sense to me ( I haven't reviewed validium functions deeply). The only thing I don't like is that my PR will face many conflicts with this one TT

PR #2642, opened 2023-10-10
Add bookmarks to streamer

ToniRamirezM:
Add bookmarks to streamer


.gitignore
```diff
@@ -11,7 +11,9 @@
 /test/contracts/bin/**/*.abi
 
 /tools/datastreamer/*.bin
+/test/datastreamer/*.db/*
 /test/datastream.bin
+/test/*.db/**/*
```
dpunish3r: Why not ignore all *.bin (not just the datastream.bin) also for the path `/test/` like in the path `/tools/`. 
Additionally the mask `/test/*.db/*` is enough, no need to set like `/test/*.db/**/*`

dpunish3r (approved):

dpunish3r (approved):

PR #2661, opened 2023-10-14
add new endpoint zkevm_getNativeBlockHashesInRange

### LINKED ISSUE #2651: new endpoint zkevm_getNativeBlockHashesInRange
# Rationale

- to provide a simple and direct way to access the block state roots.

What are the use-cases?
- native block hash synchronization.

# Implementation

Provide a new endpoint called `zkevm_getNativeBlockHashesInRange` that allows filtering the block range by block number and provides a response with the block hashes.

The hashes must be ordered by the block number.

The endpoint should have a configurable limitation to define the maximum range allowed. The default value should provide a response with a size between 5mb to 10mb, it is still pending to define.

request example: 
```bash
curl --location 'https://zkevm-rpc.com/' \
--header 'Content-Type: application/json' \
--data '{
	"jsonrpc":"2.0",
	"method":"zkevm_getNativeBlockHashesInRange",
    "params": [
        {
            "fromBlock": "0x308",
            "toBlock": "0x3E8"
        }
    ],
	"id":1
}'
```

response example:
```json
{
    "jsonrpc": "2.0",
    "id": 83,
    "result": [
        "0x4f29d9b9bc2a12528a4fb2aff747613084de9d1411c99d60aa66e7e2f57eb5a1",
        "0x4f29d9b9bc2a12528a4fb2aff747613084de9d1411c99d60aa66e7e2f57eb5a2",
        "0x4f29d9b9bc2a12528a4fb2aff747613084de9d1411c99d60aa66e7e2f57eb5a3",
        "0x4f29d9b9bc2a12528a4fb2aff747613084de9d1411c99d60aa66e7e2f57eb5a3",
        ....
        "0x4f29d9b9bc2a12528a4fb2aff747613084de9d1411c99d60aa66e7e2f57eb5a9",
        "0x4f29d9b9bc2a12528a4fb2aff747613084de9d1411c99d60aa66e7e2f57eb5a9",
    ]
}
```

tclemos:
Closes #2651.

### What does this PR do?

add new endpoint `zkevm_getNativeBlockHashesInRange` to return the native block hashes (a.k.a state root)

`MaxNativeBlockHashBlockRange` is a configuration to set the max range for block number when querying native block hashes in a single call to the state, if zero it means no limit

### Default Values: 

```toml
[RPC]
MaxNativeBlockHashBlockRange = 60000
```


state/pgstatestorage.go
```diff
@@ -2697,6 +2697,34 @@ func (p *PostgresStorage) GetDSL2Transactions(ctx context.Context, minL2Block, m
 	return l2Txs, nil
 }
 
+// GetNativeBlockHashesInRange return the state root for the blocks in range
+func (p *PostgresStorage) GetNativeBlockHashesInRange(ctx context.Context, fromBlock, toBlock uint64, dbTx pgx.Tx) ([]common.Hash, error) {
+	const l2TxSQL = `
+    SELECT l2b.state_root
+      FROM state.l2block l2b
+     WHERE l2_block_num BETWEEN $1 AND $2
```
joanestebanr: The name of the field is not l2_block_num is block_num

joanestebanr (changes requested):

jsonrpc/endpoints_zkevm.go
```diff
@@ -266,3 +266,22 @@ func (z *ZKEVMEndpoints) GetFullBlockByHash(hash types.ArgHash, fullTx bool) (in
...
+
+// GetNativeBlockHashesInRange return the state root for the blocks in range
+func (z *ZKEVMEndpoints) GetNativeBlockHashesInRange(filter NativeBlockHashBlockRangeFilter) (interface{}, types.Error) {
+	return z.txMan.NewDbTxScope(z.state, func(ctx context.Context, dbTx pgx.Tx) (interface{}, types.Error) {
+		fromBlockNumber, toBlockNumber, rpcErr := filter.GetNumericBlockNumbers(ctx, z.cfg, z.state, z.etherman, dbTx)
+		if rpcErr != nil {
+			return nil, rpcErr
+		}
+
+		nativeBlockHashes, err := z.state.GetNativeBlockHashesInRange(ctx, fromBlockNumber, toBlockNumber, dbTx)
```
joanestebanr, line 278: If toBlockNumber doesn't exist it returns less values than expected. The same if fromBlockNumber doesnt exists.

 I don't known if it's the desired behaviour. Maybe it needs a comments some place

joanestebanr (changes requested):

jsonrpc/endpoints_zkevm.go
tclemos: What do you mean by doesn't exist? Can you please provide an example?

state/pgstatestorage.go
tclemos: good catch, I missed the unit test for this method. I'll add it.

jsonrpc/endpoints_zkevm.go
joanestebanr: I have l2block from 1-200 and I ask for 100-300, It'll return 100 hashes

joanestebanr (approved):

PR #2672, opened 2023-10-16
L1 parallel sync: avoid get stuck when there are a lot of errors

### LINKED ISSUE #2371: L1 Synchronizer optimization

joanestebanr:
…consumer checks that the rollupinfo is the expected or ignore it

Closes #2371

Main reviewers:

@ARR552

### DIFF
--- a/synchronizer/l1_worker_etherman.go
+++ b/synchronizer/l1_worker_etherman.go
@@ -161,9 +161,15 @@ func (w *workerEtherman) asyncRequestRollupInfoByBlockRange(ctx contextWithCance
 			case <-time.After(sleepBefore):
 			}
 		}
+
+		// Uncomment these lines to respond with a nil result to generate fast responses (just for develop!)
+		//w.setStatus(ethermanIdle)
+		//ch <- newResponseRollupInfo(nil, time.Second, typeRequestRollupInfo, &rollupInfoByBlockRangeResult{blockRange, nil, nil, nil})
+
 		now := time.Now()
 		fromBlock := blockRange.fromBlock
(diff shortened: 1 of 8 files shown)


synchronizer/l1_worker_etherman.go
```diff
@@ -161,9 +161,15 @@ func (w *workerEtherman) asyncRequestRollupInfoByBlockRange(ctx contextWithCance
 			case <-time.After(sleepBefore):
 			}
 		}
+
+		// Uncomment these lines to respond with a nil result to generate fast responses (just for develop!)
+		//w.setStatus(ethermanIdle)
+		//ch <- newResponseRollupInfo(nil, time.Second, typeRequestRollupInfo, &rollupInfoByBlockRangeResult{blockRange, nil, nil, nil})
+
```
ARR552, line 168: Do we need this code?

ARR552 (approved):

synchronizer/l1_worker_etherman.go
joanestebanr: is a fast reminder to do some tests to the synchronization methods

PR #2685, opened 2023-10-18
+ rollupinfo request for lastest block, add previous block to check L…

### LINKED ISSUE #2682: Synchronizer L1 parallel: fails if L1 blocks re-org
# Rationale
After testing in testnet appears an error that was not able to verfiy a batch because does'nt exist the an entry in virtual_batch for that batch.

After a deep investigation the conclusion is that the block that contains the info changed. Asking for the same block (9847396) in permissionless5 (parellel) and permissionless1 (sequential) we got different block_hash:

`0x4189d9f8fd3328dddd6f8fde6ee2b4db4e782ff0611cc2ff09ba25267dbcde2a` != `0xf149e4517bde754535c15828d06e1857192b2dfa3fbc302084855e46ae8c1822`

Logs [here](https://app.datadoghq.com/logs?query=service%3Azkevm-public%20source%3Apermissionless5-synchronizer%20-status%3Ainfo&cols=host%2Cservice&index=%2A&messageDisplay=inline&refresh_mode=paused&stream_sort=desc&viz=stream&from_ts=1697017980000&to_ts=1697019960000&live=false)

# Implementation

joanestebanr:
check L1 reogs

Closes #2682

This commit introduce a new config parameter:
```
[Synchronizer.L1ParallelSynchronization.PerformanceCheck]
SwitchToSequentialModeIfIsSynchronized = false
```
If `SwitchToSequentialModeIfIsSynchronized` is true, after reaching the synchronization with L1 swith to sequential mode.

Modify this default values (`config/default.go`): 
```
[Synchronizer.L1ParallelSynchronization]
                NumberOfParallelOfEthereumClients = 10
		CapacityOfBufferingRollupInfoFromL1 = 25
```
- It increase from 2 parallel client, that is a really poor performance,  to 10 
- It increase from 10 pending result on channel to 25
- 
### What does this PR do?

Main reviewers:

- @ARR552


synchronizer/block_range.go
```diff
@@ -0,0 +1,60 @@
+package synchronizer
+
+import (
+	"errors"
+	"fmt"
+)
+
+const (
+	latestBlockNumber  uint64 = ^uint64(0)
```
ARR552, line 9:
```suggestion
	latestBlockNumber  uint64 = uint64(0)
```

synchronizer/l1_worker_etherman_test.go
```diff
@@ -18,27 +19,48 @@ import (
 func TestExploratoryWorker(t *testing.T) {
 	t.Skip("no real test, just exploratory")
 	cfg := etherman.Config{
-		URL: "http://localhost:8545",
```
ARR552: It think we should avoid this infura url

synchronizer/l1_worker_etherman_test.go
```diff
@@ -18,27 +19,48 @@ import (
...
+
 	l1Config := etherman.L1Config{
-		L1ChainID:                 1337,
-		ZkEVMAddr:                 common.HexToAddress("0x610178dA211FEF7D417bC0e6FeD39F05609AD788"),
-		MaticAddr:                 common.HexToAddress("0x5FbDB2315678afecb367f032d93F642f64180aa3"),
-		GlobalExitRootManagerAddr: common.HexToAddress("0x2279B7A0a67DB372996a5FaB50D91eAA73d2eBe6"),
+		L1ChainID:                 1442,
+		ZkEVMAddr:                 common.HexToAddress("0xa997cfD539E703921fD1e3Cf25b4c241a27a4c7A"),
+		MaticAddr:                 common.HexToAddress("0x1319D23c2F7034F52Eb07399702B040bA278Ca49"),
+		GlobalExitRootManagerAddr: common.HexToAddress("0x4d9427DCA0406358445bC0a8F88C26b704004f74"),
```
ARR552: We shouldn't commit this addresses

synchronizer/time_provider.go
```diff
@@ -0,0 +1,19 @@
...
+	Now() time.Time
+}
+
+// DefaultTimeProvider is the default implementation of TimeProvider
+type DefaultTimeProvider struct{}
+
+// Now returns current time
+func (d DefaultTimeProvider) Now() time.Time {
+	return time.Now()
+}
```
ARR552, line 19: I think we could avoid this. I mean I think it doesn't bring any value to the project

ARR552 (changes requested):

synchronizer/block_range.go
joanestebanr: In fact is not an error is the negation of 0 that is 0xff......ffff

synchronizer/time_provider.go
joanestebanr: I haven't write the unittest yet for the cache, but I'll need that to be able to test it fast an not depend on real time

synchronizer/l1_worker_etherman_test.go
joanestebanr: Changed

synchronizer/l1_worker_etherman_test.go
joanestebanr: Changed

ARR552 (approved):

PR #2733, opened 2023-10-30
delete failed txs during pool cleanup

### LINKED ISSUE #2728: Failed txs are not being deleted from the pool
It looks like the query to delete transactions from the pool only takes into account those that are in the trusted state.
Failed tx never get into the trusted state and so they are not deleted.

ToniRamirezM:
Closes #2728

### What does this PR do?

Deletes failed txs during pool cleanup

### Reviewers

Main reviewers:

- @agnusmor 
- @dpunish3r 

Codeowner reviewers:

- @tclemos


sequencer/sequencer.go
```diff
@@ -213,6 +213,13 @@ func (s *Sequencer) purgeOldPoolTxs(ctx context.Context) {
 			continue
 		}
 		log.Infof("deleted %d selected txs from the pool", len(txHashes))
+
+		// Delete failed txs older than a certain date (14 seconds per L1 block)
+		err = s.pool.DeleteFailedTransactionsOlderThan(ctx, time.Now().Add(-time.Duration(s.cfg.BlocksAmountForTxsToBeDeleted*14)*time.Second)) //nolint:gomnd
```
agnusmor, line 219: Add log.Infof("deleting failed txs from the pool")

sequencer/sequencer.go
```diff
@@ -213,6 +213,13 @@ func (s *Sequencer) purgeOldPoolTxs(ctx context.Context) {
...
+		if err != nil {
+			log.Errorf("failed to delete txs from the pool, err: %v", err)
```
agnusmor: "failed to delete failed txs from the pool" (as now is the same message as we have in line 212, in this way we can differentiate both)

agnusmor (approved):

PR #2750, opened 2023-11-03
Fix e2e group 10 (forced batches)

### LINKED ISSUE #2748: Improve e2e group 10 (forced batches) to reduce the failing rate
# Rationale

The highest rate of false failing of the e2e groups are: 
- group 10: 42%
- group 8: 12%
- group  9: 10%
- group  5: 10%

So, the most affecting group is 10 (forced batches)

joanestebanr:
Closes #2748

### What does this PR do?

* Import fixes from ARR552 in PR #2694  (wrong address for tests)
* Fix a bad usage of function `GetLatestGer` that provoke an error if the test takes longer to finish
* Refactor some code to avoid duplicated code and reduce the complexity

### Reviewers

Main reviewers:

- @ARR552 
- @tclemos 
-


test/output_from_scratch_debug_v2.log
ToniRamirezM, line 1: I think this file is not needed

test/e2e/forced_batches_test.go
```diff
@@ -23,64 +23,112 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+const (
+	// dockersArePreLaunched is a flag that indicates if dockers are pre-launched, used for local development
+	// avoiding launch time and reset database time at end (so you can check the database after the test)
+	dockersArePreLaunched = false
+	gerFinalityBlocks     = uint64(9223372036854775807) // The biggeset uint64
```
ToniRamirezM, line 30: This will cause the GER to never be considered

joanestebanr: The opposite. This value is send to the state function `GetLatestGer` that use this value to check the blocknumber below that. It also could affect to sequencer in `finalizer.go`. 
Maybe the bug is in the state funcion instead of this constant

ToniRamirezM: You are right. Consider using math.MaxUint64 if it works for you.

ToniRamirezM (approved):

PR #2774, opened 2023-11-09
improve OOC and OOG errors

### LINKED ISSUE #2483: Review out of counters user message to make it clear
we must review the behavior and message returned when calling each of these endpoints:
- eth_estimateGas
- eth_call
- eth_sendRawTransaction

tclemos:
Closes #2483.

improve the way the methods `eth_estimateGas`, `eth_call` and `eth_sendRawTransaction` handle `Out Of Counters` and `Out of Gas` errors in order to provider a better communication to the users.


pool/pool.go
```diff
@@ -208,7 +209,11 @@ func (p *Pool) StoreTx(ctx context.Context, tx types.Transaction, ip string, isW
 			log.Errorf("error adding event: %v", err)
 		}
 		// Do not add tx to the pool
-		return ErrOutOfCounters
+		err = ErrOutOfCounters
+		if preExecutionResponse.OOCError != nil {
+			err = fmt.Errorf("failed to add tx to the pool: %v", preExecutionResponse.OOCError.Error())
+		}
+		return err
```
ARR552: How is it possible to have a ErrOutOfCounters if the preExecutionResponse.OOCError is nil?? I mean if the executor didn't return the OOCError it shouldn't be possible to have this situation

pool/pool.go
```diff
@@ -56,6 +56,7 @@ type preExecutionResponse struct {
 	usedZkCounters       state.ZKCounters
 	isExecutorLevelError bool
 	isOOC                bool
+	OOCError             error
```
ToniRamirezM, line 58: Maybe it is cleaner if we change both isOOC and isOOG for errors en remove both booleans.

ToniRamirezM (approved):

pool/pool.go
ToniRamirezM: I understand OOC is the default case and then it gets overwritten if there is an OOC

tclemos: > How is it possible to have a ErrOutOfCounters if the preExecutionResponse.OOCError is nil?? I mean if the executor didn't return the OOCError it shouldn't be possible to have this situation

yes, I just added this check just in case a code in the future would change the logic, to keep this code protected, just in case a new OOC code is added, we change part of the code and forget to change some other part, this could break the pool.

But I'll be refactoring the code to avoid this double check.

pool/pool.go
tclemos: Yes, I think this is a good idea and I'll try to do this refactoring

PR #2777, opened 2023-11-10
Added L2GasPriceSuggesterFactor config parameter to simulate estimated L2 gas price (EGP disabled)

agnusmor:
Closes #<issue number>.

### What does this PR do?

Adds L2GasPriceSuggesterFactor config parameter to "simulate" estimated L2 gas price when EGP is disabled.

The following configuration parameter has been added:

> [Pool]
> [Pool.EffectiveGasPrice]
> L2GasPriceSuggesterFactor = 0.5

### Reviewers

Main reviewers:

@ToniRamirezM 
@joan-esteban


ToniRamirezM (approved):

config/environments/local/local.node.config.toml
```diff
@@ -46,6 +46,7 @@ GlobalQueue = 1024
 		NetProfit = 1
 	    BreakEvenFactor = 1.1			
 		FinalDeviationPct = 10
+		L2GasPriceSuggesterFactor = 0.5
```
tclemos, line 49: as a suggestion, since this value is the same as the default, we can avoid adding this line.
it would be good to have in this file only the values different from the default, it would also be good to do a cleanup in the future, but for now we can keep the new values out of the file to make it easier to clean when we are ready to do it.

pool/config.go
```diff
@@ -64,4 +64,8 @@ type EffectiveGasPriceCfg struct {
 
 	// FinalDeviationPct is the max allowed deviation percentage BreakEvenGasPrice on re-calculation
 	FinalDeviationPct uint64 `mapstructure:"FinalDeviationPct"`
+
+	// L2GasPriceSuggesterFactor is the factor to apply to L1 gas price to get the suggested L2 gas price used in the
+	// calculations when the effective gas price is disabled (testing/metrics purposes)
+	L2GasPriceSuggesterFactor float64 `mapstructure:"L2GasPriceSuggesterFactor"`
```
tclemos, line 70: if this parameter is used only for metrics and testing, is it possible to not be a configuration but a mocked price suggester somewhere that would be injected the same way we do for the real one, but only the situations we need it?

Also, if this parameter is used only when the EGP is disabled, I understand that it means we are always going to have an EGP "enabled", a fixed value and a dinamic value. Not sure if this config is conflicting with the `Enabled` config.

test/config/debug.node.config.toml
```diff
@@ -46,6 +46,7 @@ GlobalQueue = 1024
 		NetProfit = 1
 	    BreakEvenFactor = 1.1			
 		FinalDeviationPct = 10
+		L2GasPriceSuggesterFactor = 0.5
```
tclemos, line 49: please check the comment in the local config file.

test/config/test.node.config.toml
```diff
@@ -46,6 +46,7 @@ GlobalQueue = 1024
 		NetProfit = 1
 	    BreakEvenFactor = 1.1			
 		FinalDeviationPct = 10
+		L2GasPriceSuggesterFactor = 0.5
```
tclemos, line 49: please check the comment in the local config file.

tclemos (approved):
I've added some comments as questions and suggestions, nothing that would possibly block the merging, but to be considered and check it they make sense to change something. Feel free to discard all of them if you think they don't apply.

pool/config.go
agnusmor: It is not used for the tests (CI) it is used in production (internal, testnet, mainnet) when the EGP is disabled to do anyway the calculation and logs the results about what would have happened if the EGP had been enabled. With this info (that is showed in the logs but also store in the state.transaction table) we can do a post analisys to review how the EGP is working an get metrics about the amount of txs executed

config/environments/local/local.node.config.toml
agnusmor: Ok, but until now we have added always to this files all the config parameters. I think we need to continue working in this way (to be consistent) and have a meeting to review if we want to do this in the way you are commenting and if we agree create an issue to cleanup the config files and put only the values that are different from the default (as you propose)

PR #2782, opened 2023-11-11
replace atomic wsConn by concurrentWsConn

### LINKED ISSUE #2781: WS writing concurrency error
#### System information

zkEVM Node version: `v0.3.4`
Network: `Mainnet`

```log
panic: concurrent write to websocket connection

goroutine 47197 [running]:
github.com/gorilla/websocket.(*messageWriter).flushFrame(0xc00226ade8, 0x1, {0xc001bd3200?, 0xc001bd2e00?, 0x0?})
        /go/pkg/mod/github.com/gorilla/websocket@v1.5.1/conn.go:632 +0x4b8
github.com/gorilla/websocket.(*Conn).WriteMessage(0xc000c60580, 0xc0033021e0?, {0xc001bd2e00, 0x61b, 0x700})
        /go/pkg/mod/github.com/gorilla/websocket@v1.5.1/conn.go:785 +0x127
github.com/0xPolygonHermez/zkevm-node/jsonrpc.(*Filter).sendSubscriptionResponse(0xc001136780, {0xc001aba600, 0x5a5, 0x600})
        /src/jsonrpc/query.go:90 +0x15d
github.com/0xPolygonHermez/zkevm-node/jsonrpc.(*Filter).SendEnqueuedSubscriptionData(0xc001136780)
        /src/jsonrpc/query.go:65 +0xc5
created by github.com/0xPolygonHermez/zkevm-node/jsonrpc.(*EthEndpoints).notifyNewHeads in goroutine 46911
        /src/jsonrpc/endpoints_eth.go:1057 +0x27d
```

tclemos:
Closes #2781.

### What does this PR do?

Replace the way the node handle the concurrent access to the web socket connection to handle concurrent writings

### DIFF
--- a/jsonrpc/endpoints_debug.go
+++ b/jsonrpc/endpoints_debug.go
@@ -218,7 +218,7 @@ func (d *DebugEndpoints) TraceBatchByNumber(httpRequest *http.Request, number ty
 
 		buffer := make(chan byte, bufferSize)
 
-		mu := sync.Mutex{}
+		mu := &sync.Mutex{}
 		wg := sync.WaitGroup{}
 		wg.Add(len(receipts))
 		responses := make([]traceResponse, 0, len(receipts))
--- a/jsonrpc/server.go
+++ b/jsonrpc/server.go
@@ -423,9 +423,7 @@ func (s *Server) decreaseWsConnCounter() {
 }
 
 func (s *Server) logConnCounters() {
-	httpConnCounter := s.httpConnCounter
-	wsConnCounter := s.wsConnCounter
-	totalConnCounter := httpConnCounter + wsConnCounter
+	totalConnCounter := s.httpConnCounter + s.wsConnCounter
 	log.Infof("[ HTTP conns: %v | WS conns: %v | Total conns: %v ]", httpConnCounter, wsConnCounter, totalConnCounter)
 }
 
--- a/jsonrpc/server.go
+++ b/jsonrpc/server.go
@@ -10,7 +10,6 @@ import (
 	"net"
 	"net/http"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	"github.com/0xPolygonHermez/zkevm-node/jsonrpc/metrics"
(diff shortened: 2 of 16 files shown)


jsonrpc/server.go
```diff
@@ -387,47 +386,47 @@ func (s *Server) handleWs(w http.ResponseWriter, req *http.Request) {
...
-	atomic.AddInt64(&s.wsConnCounter, -1)
+	s.wsConnCounter--
 	s.logConnCounters()
 	s.connCounterMutex.Unlock()
 }
 
 func (s *Server) logConnCounters() {
-	httpConnCounter := atomic.LoadInt64(&s.httpConnCounter)
-	wsConnCounter := atomic.LoadInt64(&s.wsConnCounter)
+	httpConnCounter := s.httpConnCounter
```
agnusmor: You can use directly s.httpConnCounter and s.wsConnCounter, not needed to assign the values to local variables, only to simplify the code

jsonrpc/endpoints_debug.go
```diff
@@ -218,7 +218,7 @@ func (d *DebugEndpoints) TraceBatchByNumber(httpRequest *http.Request, number ty
 
 		buffer := make(chan byte, bufferSize)
 
-		mu := sync.Mutex{}
+		mu := &sync.Mutex{}
```
agnusmor, line 221: This is something I viewed when I started to review the code, why this needs to be a pointer? I think it should work doing the assignement without &, isn't it?

tclemos: I have faced a race condition issue when running the tests and I've found this explanation: https://stackoverflow.com/questions/64735267/race-condition-when-using-mutex-and-reflect

When I've moved the mutex to be a pointer it fixed the problem. Them I decided to review the mutex we had and made them pointers to avoid the problem, since the RPC uses many go routines to handle the requests.

agnusmor (approved):

PR #2799, opened 2023-11-16
Fix tx size check

### LINKED ISSUE #2798: Properly validate max tx size before adding it to the pool

ToniRamirezM:
Closes #2798

### What does this PR do?

Change the way the tx size is checked

### Reviewers

Main reviewers:

- @agnusmor 
- @dpunish3r


pool/pool.go
```diff
@@ -426,7 +426,12 @@ func (p *Pool) validateTx(ctx context.Context, poolTx Transaction) error {
 	}
 
 	// Reject transactions over defined size to prevent DOS attacks
-	if poolTx.Size() > p.cfg.MaxTxBytesSize {
+	decodedTx, err := state.EncodeTransaction(poolTx.Transaction, 0xFF, p.cfg.ForkID) //nolint: gomnd
+	if err != nil {
+		return ErrTxTypeNotSupported
```
agnusmor, line 431: isn't it better to return err?

sequencer/txtracker.go
```diff
@@ -53,7 +53,7 @@ func newTxTracker(tx types.Transaction, counters state.ZKCounters, ip string) (*
 		GasPrice: tx.GasPrice(),
 		Cost:     tx.Cost(),
 		BatchResources: state.BatchResources{
-			Bytes:      tx.Size(),
+			Bytes:      uint64(len(rawTx)) + state.EfficiencyPercentageByteLength,
```
agnusmor, line 56: maybe we need to use state.EncodeTransaction to get the rawTx and pass the forkid to the txTracker, to be coherent

pool/pool.go
ToniRamirezM: This error es returned to the user by the RPC, it should be a standard error if possible.

agnusmor (approved):

PR #2808, opened 2023-11-20
set synchronizer sequential mode as default (#2807) cherry-pick

joanestebanr:
* set sequential mode as default

### Reviewers

Main reviewers:

- @agnusmor 
- @ARR552


joanestebanr:
Next version is going to have by default 'parallel' so we don't need to cherry-pick this change in develop

PR #2809, opened 2023-11-21
etherman: use pointer instead of value reduce memory alloc

jsvisa:
### What does this PR do?

Use Pointer instead of value to reduce memory copy

### DIFF
--- a/etherman/etherman.go
+++ b/etherman/etherman.go
@@ -456,8 +456,9 @@ func (etherMan *Client) readEvents(ctx context.Context, query ethereum.FilterQue
 	blocksOrder := make(map[common.Hash][]Order)
 	startProcess := time.Now()
 	for _, vLog := range logs {
+		vLog := vLog
 		startProcessSingleEvent := time.Now()
-		err := etherMan.processEvent(ctx, vLog, &blocks, &blocksOrder)
+		err := etherMan.processEvent(ctx, &vLog, &blocks, &blocksOrder)
 		metrics.ProcessSingleEventTime(time.Since(startProcessSingleEvent))
 		metrics.EventCounter()
 		if err != nil {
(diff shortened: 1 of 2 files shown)


etherman/etherman.go
```diff
@@ -324,7 +324,7 @@ func (etherMan *Client) readEvents(ctx context.Context, query ethereum.FilterQue
 	startProcess := time.Now()
 	for _, vLog := range logs {
 		startProcessSingleEvent := time.Now()
-		err := etherMan.processEvent(ctx, vLog, &blocks, &blocksOrder)
+		err := etherMan.processEvent(ctx, &vLog, &blocks, &blocksOrder)
```
ToniRamirezM, line 461: As the linter warns, this would introduce a bug where all the variables used in the loop point to the same value.

ToniRamirezM (changes requested):

etherman/etherman.go
jsvisa: fixed, please take another look.

tclemos (commented):
what's exactly the problem with the copy? are you facing memory allocation problems specifically because of these methods?

Passing a pointer means the called method has control over the data stored by the caller, which is not needed but can lead to bugs in the future. The processEvent method doesn't need to change anything regarding the log, just read, passing a copy guarantee this behavior will be respected, because it is impossible to the called method to access the memory pointer from the caller data.

I understand the suggestion, but both solutions work, for me it is a trade-off decision and in case we are note facing memory allocation due to this and the PR is just an early optimization suggestion, I recommend keeping it as a copy in favor of the code maintenance.

ToniRamirezM:
Should we close this PR?

jsvisa:
> what's exactly the problem with the copy? are you facing memory allocation problems specifically because of these methods?
> 
> Passing a pointer means the called method has control over the data stored by the caller, which is not needed but can lead to bugs in the future. The processEvent method doesn't need to change anything regarding the log, just read, passing a copy guarantee this behavior will be respected, because it is impossible to the called method to access the memory pointer from the caller data.
> 
> I understand the suggestion, but both solutions work, for me it is a trade-off decision and in case we are note facing memory allocation due to this and the PR is just an early optimization suggestion, I recommend keeping it as a copy in favor of the code maintenance.

In my opinion, too much value copying can put a lot of pressure on Golang's GC, and as you mentioned, here we are just reading without making any modifications to the data inside. So, copying a large struct in this way is not efficient.

PR #2823, opened 2023-11-22
Feat/txpool_status rpc endpoint

IdrisHanafi:
### What does this PR do?

In this PR, I introduce the JSON RPC endpoint `txpool_status`. This is a endpoint that `geth` [implements](https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-txpool#txpool-status), which essentially returns the # of pending and queued transactions in the transaction pool. The response is of the format:
```
$ curl http://localhost:8123 -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"txpool_status","params":[],"id":1}' | jq
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "pending": "0x0",
    "queued": "0x0"
  }
}
```

In the case of a permissionless node that isn't connected to the source Pool DB, it will just return 0 for both values. 

**What's the usecase for this endpoint?**

The primary usecase is to be able to conduct [loadtests with `polygon-cli`](https://github.com/maticnetwork/polygon-cli/blob/main/doc/polycli_loadtest.md) on a zkEVM net using adaptive rate limiting. More specifically, we need a way to query the queue size to effectively apply the load to the system in an [AIMD fashion](https://ocw.mit.edu/courses/6-033-computer-system-engineering-spring-2018/d071f3da1e8e5b646e220525783eb2ba_MIT6_033S18lec11.pdf).

### Testing

There were 3 main tests included for this endpoint to ensure it works as intended:

1. Mock tests for the endpoint:
```
$ cd jsonrpc
$ go test
```

2. e2e test integrated into `e2e-test-group-4`:
```
$ make build-docker
$ cd test
$ make test-e2e-group-4
```

3. Environment test (i.e. running the endpoint from a devnet):
```
$ make build-docker
$ cd test && make run

# send transaction or generate load with cast or polycli:
$ cast send --legacy --private-key 0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80 --rpc-url http://localhost:8123 --value 0.01ether 0x85dA99c8a7C2C95964c8EfD687E95E632Fc533D6
$ polycli loadtest --legacy --verbosity 700 --mode t --rate-limit 10 --concurrency 10 --requests 10 --rpc-url http://localhost:8123 --private-key ac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80 --gas-price 100000000000

# then watch pool status increase:
$ curl http://localhost:8123 -X POST -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"txpool_status","params":[],"id":1}' | jq
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": {
    "pending": "0x4",
    "queued": "0x56"
  }
}
```

### Reviewers

Main reviewers:

- @arnaubennassar 
- @ARR552 
- @ToniRamirezM


tclemos (changes requested):
Hey, thanks for the help with the RPC!

Indeed we need to add tests before merging it, unit test is a must have, but it would be very important to also have E2E tests.

I would recommend reverting the changes in the `.gitignore` and also the ones related to the `docker-compose`, making the PR strict to the new endpoint, only.

It's also important to list the new endpoint and the permissionless observation in this file: ./docs/json-rpc-endpoints.md

IdrisHanafi:
Perfect, I'll be updating this. Thanks for the fast feedback!

PR #2827, opened 2023-11-23
etrog Integration with executor

ToniRamirezM:
### What does this PR do?

First iteration on integration with executor.

### Reviewers

Main reviewers:

- @agnusmor 
- @tclemos


tclemos (approved):

state/types.go
```diff
@@ -41,6 +46,26 @@ type ProcessBatchResponse struct {
 	FlushID              uint64
 	StoredFlushID        uint64
 	ProverID             string
+	GasUsed_V2           uint64
+	SMTKeys_V2           []merkletree.Key
+	ProgramKeys_V2       []merkletree.Key
+	ForkID               uint64
+}
```
joanestebanr, line 53: Some struct are duplicated for V2, maybe make sense to create a `ProcessBatchResponseV2` with the legacy fields
like: 
```
type ProcessBatchResponseV2 struct {
     ProcessBatchResponse
     GasUsed_V2           uint64
(...)
}
```

state/batchV2.go
```diff
@@ -0,0 +1,230 @@
...
+	log.Debugf("ExecuteBatchV2[processBatchRequest.GlobalExitRoot]: %v", hex.EncodeToHex(processBatchRequest.GlobalExitRoot))
+	log.Debugf("ExecuteBatchV2[processBatchRequest.OldAccInputHash]: %v", hex.EncodeToHex(processBatchRequest.OldAccInputHash))
+	log.Debugf("ExecuteBatchV2[processBatchRequest.EthTimestamp]: %v", processBatchRequest.EthTimestamp)
+	log.Debugf("ExecuteBatchV2[processBatchRequest.Coinbase]: %v", processBatchRequest.Coinbase)
+	log.Debugf("ExecuteBatchV2[processBatchRequest.UpdateMerkleTree]: %v", processBatchRequest.UpdateMerkleTree)
+	log.Debugf("ExecuteBatchV2[processBatchRequest.ChainId]: %v", processBatchRequest.ChainId)
+	log.Debugf("ExecuteBatchV2[processBatchRequest.ForkId]: %v", processBatchRequest.ForkId)
+	log.Debugf("ExecuteBatchV2[processBatchRequest.ContextId]: %v", processBatchRequest.ContextId)
+
+	processBatchResponse, err := s.executorClient.ProcessBatch(ctx, processBatchRequest)
```
joanestebanr: It's calling incaberry executor, may be have to call to `executorClient.ProcessBatchV2`

ToniRamirezM: You are right. We still don't have the l1inforoot or the timestamp limit, but this call should be to ProcessBatchV2.

state/types.go
ToniRamirezM: We have tried to encapsulate both responses in just one, inside the state

joanestebanr (approved):

PR #2865, opened 2023-11-29
Feature/#2840 l1 info root

### LINKED ISSUE #2840: Improve l1InfoTree to keep siblings in memory
Improve l1InfoTree to keep siblings in memory to avoid recompute the full mt each time a ger is received

ARR552:
Closes #2840

### What does this PR do?

This Pr adds the functionality to keep siblings in memory and not compute the full tree all the time

### Reviewers

- @agnusmor 
- @joanestebanr


cmd/run.go
```diff
@@ -488,8 +489,20 @@ func newState(ctx context.Context, c *config.Config, l2ChainID uint64, forkIDInt
...
+		log.Fatal("error getting all leaves. Error: ", err)
+	}
+	var leaves [][32]byte
+	for _, leaf := range allLeaves {
+		leaves = append(leaves, leaf.Hash())
+	}
+	mt, err := l1infotree.NewL1InfoTree(uint8(32), leaves) //nolint:gomnd
+	if err != nil {
+		log.Fatal("error creating L1InfoTree. Error: ", err)
+	}
```
joanestebanr, line 503: Why this code is here instead in a function of state to initialize the cache of the L1InfoTree?

state/l1infotree.go
```diff
@@ -19,53 +20,40 @@ type L1InfoTreeLeaf struct {
 type L1InfoTreeExitRootStorageEntry struct {
 	L1InfoTreeLeaf
 	L1InfoTreeRoot  common.Hash
-	L1InfoTreeIndex uint64
+	L1InfoTreeIndex uint
 }
```
joanestebanr, line 24: L1InfoTreeIndex is `uint32` there are a type defined `L1InfoTreeIndexType` for that

state/l1infotree.go
```diff
@@ -19,53 +20,40 @@ type L1InfoTreeLeaf struct {
...
+	gerIndex, err := s.GetLatestIndex(ctx, dbTx)
+	if err != nil && !errors.Is(err, ErrNotFound) {
+		log.Error("error getting latest l1InfoTree index. Error: ", err)
 		return nil, err
+	} else if err == nil {
+		newIndex = gerIndex + 1
 	}
-	root, err := buildL1InfoTree(allLeaves)
+	log.Debug("latestIndex: ", gerIndex)
+	root, err := s.l1InfoTree.AddLeaf(newIndex, l1InfoTreeLeaf.Hash())
```
joanestebanr, line 43: Instead of looking for next index in DB, doesnt make more sense that the `s.l1InfoTree.AddLeaf`  returns the leafIndex?

state/pgstatestorage/l1infotree.go
```diff
@@ -49,3 +45,18 @@ func (p *PostgresStorage) GetAllL1InfoRootEntries(ctx context.Context, dbTx pgx.
...
+		WHERE l1_info_tree_index IS NOT NULL`
+	var l1InfoTreeIndex *uint
+	e := p.getExecQuerier(dbTx)
+	err := e.QueryRow(ctx, getLatestIndexSQL).Scan(&l1InfoTreeIndex)
+	if err != nil {
+		return 0, err
+	}
+	if l1InfoTreeIndex == nil {
+		return 0, state.ErrNotFound
+	}
```
joanestebanr, line 60: In case of no rows on DB, maybe is better to return a 0 , nil

state/l1infotree.go
ARR552: There is a check in the AddLeaf function to use the right one. If there is a mismatch for some reason an error will be returned

state/l1infotree.go
ARR552: I now, It was completely unusefull and it was deleted

cmd/run.go
ARR552: I thought to do it in the way you suggests but finally I decided to pass the component instance in the same way we are doing with other components

state/pgstatestorage/l1infotree.go
ARR552: If it returns 0, then the l1InfoTree will fail because index 0 is used as the first element. So it needs to detect the initial situation to compute the tree successfully

state/l1infotree.go
joanestebanr: This type was validated in the [PR 2844](https://github.com/0xPolygonHermez/zkevm-node/pull/2844).
I think that is a good idea to have types for this kind of values. 
Also keep in mind that `uint` depend of the compilation platform

joanestebanr (approved):

PR #2875, opened 2023-12-04
Add StoreL2Block (ETROG) function to the state

agnusmor:
### What does this PR do?

Adds the StoreL2Block funcion to the state. This function is used to store L2 blocks generated with the fork ETROG (but it also should work for previous forks, if needed)

### Reviewers

Main reviewers:

@joanestebanr
@ToniRamirezM 
@ARR552

### DIFF
--- a/state/transaction.go
+++ b/state/transaction.go
@@ -298,8 +298,8 @@ func (s *State) StoreL2Block(ctx context.Context, batchNumber uint64, l2Block *P
 			continue
 		}
 
-		transaction := &txResponse.Tx
-		transactions = append(transactions, transaction)
+		txResp := *txResponse
+		transactions = append(transactions, &txResp.Tx)
 
 		storeTxsEGPData = append(storeTxsEGPData, StoreTxEGPData{EGPLog: nil, EffectivePercentage: uint8(txResponse.EffectivePercentage)})
 		if txsEGPLog != nil {
--- a/state/transaction.go
+++ b/state/transaction.go
@@ -258,6 +258,75 @@ func (s *State) StoreTransactions(ctx context.Context, batchNumber uint64, proce
 	return nil
 }
 
+// StoreL2Block stores a l2 block into the state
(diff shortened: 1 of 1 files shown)


state/transaction.go
```diff
@@ -258,6 +258,76 @@ func (s *State) StoreTransactions(ctx context.Context, batchNumber uint64, proce
...
+// StoreL2Block stores a l2 block into the state
+func (s *State) StoreL2Block(ctx context.Context, batchNumber uint64, l2Block *ProcessBlockResponse, txsEGPLog []*EffectiveGasPriceLog, dbTx pgx.Tx) error {
+	if dbTx == nil {
+		return ErrDBTxNil
+	}
+
+	log.Debugf("storing l2 block %d, txs %d, hash %d", l2Block.BlockNumber, len(l2Block.TransactionResponses), l2Block.BlockHash.String())
+	start := time.Now()
+
+	dbTx, err := s.BeginStateTransaction(ctx)
```
joanestebanr, line 270: If you receive a dbTx usually means that somebody else have started a transaction

state/transaction.go
```diff
@@ -258,6 +258,76 @@ func (s *State) StoreTransactions(ctx context.Context, batchNumber uint64, proce
...
+		receipt := generateReceipt(header.Number, txResponse)
+		receipts = append(receipts, receipt)
+	}
+
+	// Create block to be able to calculate its hash
+	block := NewL2Block(l2Header, transactions, []*L2Header{}, receipts, &trie.StackTrie{})
+	block.ReceivedAt = time.Unix(int64(l2Block.Timestamp), 0)
+
+	for _, receipt := range receipts {
+		receipt.BlockHash = block.Hash()
```
joanestebanr, line 318: It must be the computed hash or the stateRoot?

agnusmor: It must by the state root if it is ETROG, but this is replaced later inside the AddL2Block function if the fork ks ETROG, maybe we can add also the code here to change the hash if it is ETROG, but we are doing it twice

state/transaction.go
agnusmor: I missed this, I will change it

joanestebanr (approved):

PR #2892, opened 2023-12-11
fix inconsistent status of trustedState

### LINKED ISSUE #2890: The sync module hangs due to inconsistent status of trustedState
in syncTrustedState func
 Step 1 :  update L2Data for  batch[0] ;  [link](https://github.com/0xPolygonHermez/zkevm-node/blob/5576e50b42f264881aec06c83502f246c53f9e19/synchronizer/synchronizer.go#L989)
 Step 2 :  processAndStoreTxs  ; [link](https://github.com/0xPolygonHermez/zkevm-node/blob/5576e50b42f264881aec06c83502f246c53f9e19/synchronizer/synchronizer.go#L1004)
 Step 3 : update trustedState.lastStateRoot ; [link](https://github.com/0xPolygonHermez/zkevm-node/blob/5576e50b42f264881aec06c83502f246c53f9e19/synchronizer/synchronizer.go#L611)

RootCause：If the second step fails, the first step should not be updated.

The problem is: if the first step is updated but the third step is not updated, then the status of trustStated will be inconsistent (for example, the current stateRoot should be the execution result of the first five transactions, but because the third step has not been executed, stateRoot is still is the execution result of the first 4 transactions), it will cause the final stateRoot to be wrong.

scf0220:
Closes #2890 .

### Reviewers

Main reviewers:

- @John
- @Doe

Codeowner reviewers:

- @-Alice
- @-Bob

### DIFF
--- a/synchronizer/synchronizer.go
+++ b/synchronizer/synchronizer.go
@@ -831,6 +831,7 @@ func (s *ClientSynchronizer) Stop() {
 
 func (s *ClientSynchronizer) processTrustedBatch(trustedBatch *types.Batch, dbTx pgx.Tx) ([]*state.Batch, *common.Hash, error) {
 	log.Debugf("Processing trusted batch: %d", uint64(trustedBatch.Number))
+	var needUpdatL2DataForBatch0 = false
 	trustedBatchL2Data := trustedBatch.BatchL2Data
 	batches := s.trustedState.lastTrustedBatches
 	log.Debug("len(batches): ", len(batches))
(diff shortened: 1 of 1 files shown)


ARR552:
There is no inconsistency. Memory variables are reset in line 378 and 379

scf0220:
> There is no inconsistency. Memory variables are reset in line 378 and 379

Thank you for your reply. The version we are using is v0.3.0. The code you mentioned did not exist at that time. Adding this code will eliminate this problem.

PR #2904, opened 2023-12-13
etrog: batch timestamp is set when is virtualized

### LINKED ISSUE #2903: etrog: batch timestamp is unknown until the batch is synchronized from L1
# Rationale

- The batches got the timestamp from the L1 block that contains the event
- When a batch is process we give the `timestampLimit` that is not the batch tstamp

# Implementation

- db.batch.timestamp -> put the timestampLimit used to generate the batch
- Add timestamp to db.virtual_batch to keep real timestamp of a batch
- Change zkevm-RPC to response both timestamps

joanestebanr:
Closes #2903

### What does this PR do?
- Add field virtual_batch.batch_timestamp
- Modify RPC to return the right batch_timestamp
### Reviewers
@ToniRamirezM 
@ARR552 
@tclemos

### DIFF
--- a/db/migrations/state/0013.sql
+++ b/db/migrations/state/0013.sql
@@ -14,6 +14,9 @@ CREATE INDEX IF NOT EXISTS idx_transaction_l2_hash ON state.transaction (l2_hash
 ALTER TABLE state.batch
     ADD COLUMN wip BOOLEAN NOT NULL;
 
+ALTER TABLE state.virtual_batch
+    ADD COLUMN batch_timestamp TIMESTAMP WITH TIME ZONE NULL;
+
 -- +migrate Down
 ALTER TABLE state.exit_root
     DROP COLUMN prev_block_hash,
@@ -29,4 +32,7 @@ DROP INDEX IF EXISTS state.idx_transaction_l2_hash;
 
 ALTER TABLE state.batch
     DROP COLUMN wip;
+
+ALTER TABLE state.virtual_batch
+    DROP COLUMN batch_timestamp;
 
--- a/db/migrations/state/0013.sql
+++ b/db/migrations/state/0013.sql
@@ -15,7 +15,7 @@ ALTER TABLE state.batch
     ADD COLUMN wip BOOLEAN NOT NULL;
 
 ALTER TABLE state.virtual_batch
-    ADD COLUMN batch_timestamp TIMESTAMP WITH TIME ZONE NULL;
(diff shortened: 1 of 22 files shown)


db/migrations/state/0013.sql
```diff
@@ -14,6 +14,9 @@ CREATE INDEX IF NOT EXISTS idx_transaction_l2_hash ON state.transaction (l2_hash
 ALTER TABLE state.batch
     ADD COLUMN wip BOOLEAN NOT NULL;
 
+ALTER TABLE state.virtual_batch
+    ADD COLUMN batch_timestamp TIMESTAMP WITH TIME ZONE NULL;
```
ARR552:
```suggestion
ALTER TABLE state.virtual_batch
    ADD COLUMN final_timestamp TIMESTAMP WITH TIME ZONE NULL;
```
The name could be l1_timestamp as well
what about renaming the current timestamp column to `trusted_timestamp` or `timestamp_limit`?
Doing it, we'd have `trusted_timestamp or timestamp_limit` and `final_timestamp or l1_timestamp`

state/batch.go
```diff
@@ -525,3 +540,27 @@ func (s *State) GetLastBatch(ctx context.Context, dbTx pgx.Tx) (*Batch, error) {
...
+	forkid := s.GetForkIDByBatchNumber(batchNumber)
+	batchTimestamp, virtualTimestamp, err := s.GetRawBatchTimestamps(ctx, batchNumber, dbTx)
+	if err != nil {
+		return nil, err
+	}
+	if forkid >= FORKID_ETROG {
+		return &BatchTimestampData{
+			BatchTimestamp:   virtualTimestamp,
+			LimitTimestampV2: batchTimestamp,
+			IsV2:             true,
```
ARR552: Why do we need the isV2 param??. If LimitTimestampV2 is nil we are in v1 and if not, we are in v2. I think we could delete it

joanestebanr: I have delete that, because I'm not able to use timeLimit in syncronizer. So I left only the batch_timestamp

db/migrations/state/0013.sql
joanestebanr: The previous batches the batch.timestamp is the final tstamp, so I think that is confusing also

joanestebanr: We agree to rename it to `timestamp_batch_etrog`

ARR552 (approved):

PR #2927, opened 2023-12-15
add WIP to state.batch struct

### LINKED ISSUE #2926: etrog: RPC: add the flag WIP to the returned Batches in zkevm end-point
# Rationale

Before etrog you could check if a batch is open using StateRoot != 0 but now it keeps the intermediate stateRoot and is no longer to known if a batch is close or open

# Implementation

Add Wip flag to the response of `zkevm_getBatchByNumber`

joanestebanr:
Closes #2926

### What does this PR do?
Add WIP flag to response of RPC

### Reviewers

Main reviewers:

- @ToniRamirezM 
- @agnusmor 
- @tclemos 

Codeowner reviewers:

- @-Alice
- @-Bob

### DIFF
--- a/jsonrpc/types/types.go
+++ b/jsonrpc/types/types.go
@@ -352,12 +352,14 @@ type Batch struct {
 	Blocks              []BlockOrHash       `json:"blocks"`
 	Transactions        []TransactionOrHash `json:"transactions"`
 	BatchL2Data         ArgBytes            `json:"batchL2Data"`
+	WIP                 bool                `json:"wip"`
 }
 
 // NewBatch creates a Batch instance
 func NewBatch(batch *state.Batch, virtualBatch *state.VirtualBatch, verifiedBatch *state.VerifiedBatch, blocks []state.L2Block, receipts []types.Receipt, fullTx, includeReceipts bool, ger *state.GlobalExitRoot) (*Batch, error) {
 	batchL2Data := batch.BatchL2Data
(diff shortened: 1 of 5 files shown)


ToniRamirezM (approved):

jsonrpc/types/types.go
```diff
@@ -352,12 +352,14 @@ type Batch struct {
 	Blocks              []BlockOrHash       `json:"blocks"`
 	Transactions        []TransactionOrHash `json:"transactions"`
 	BatchL2Data         ArgBytes            `json:"batchL2Data"`
+	WIP                 bool                `json:"wip"`
```
tclemos: Do we really need both closed and wip fields? As far as I understood WIP is used exclusively to know if a batch is closed or not, in theory we don't need to add this field to the RPC response. It opens an opportunity to have inconsistent data in case closed and wip were equals for any reason. can we remove WIP and continue only with closed?

tclemos (changes requested):
scanBatch method is a shared method among different other methods loading batches from the storage, changing it requires all the methods using it have the wip field in the query, please review them and add the wip field to the missing ones

jsonrpc/types/types.go
joanestebanr: then I proposed to remove 'closed' and left wip that is going to map the state.batch field

tclemos (approved):

PR #2983, opened 2023-12-22
Fix data streamer out-of-order with the fork upgrading

zjg555543:
### What does this PR do?

GenerateDataStreamerFile(data streamer file) will be out of order batches at the batch of fork5 to fork 6


ToniRamirezM:
Thanks @zjg555543

PR #3002, opened 2023-12-29
Security - Add SHA Pin to Third Party Actions/Checkout Action (fixes #3001)

joethechicken:
Closes #3001 (https://github.com/0xPolygonHermez/zkevm-node/issues/3001).

### What does this PR do?

Currently, the third party actions/checkout action specifies a version (v3) rather than pinning the action via a full length commit SHA. GitHub recommends using the full length SHA for security reasons because "pinning to a particular SHA helps mitigate the risk of a bad actor adding a backdoor to the action's repository, as they would need to generate a SHA-1 collision for a valid Git object payload."

This PR pings the third party actions/checkout action to the latest commit of the v3 version. The action will behave exactly the same as it does now, as it uses the same commit, but the SHA pin will protect this repository if the third party action is ever compromised in the future.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -40,7 +40,7 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v3
+      uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744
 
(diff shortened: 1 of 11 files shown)


tclemos:
@cla-bot check

tclemos (approved):

PR #3016, opened 2024-01-04
set eth_getLogs default value of fromBlock to "latest"

### LINKED ISSUE #3015: eth_getLogs api behavior different to default ETH API
#### System information

zkEVM Node version: `v0.0.X-RCXX`
OS & Version: `Windows/Linux/OSX`
Commit hash : (if `develop`)
Network: `Mainnet/Testnet`

The default ETH_GetLogs Api param for `fromBlock` is "latest", but in zkevm, the default value is zero
#### Expected behaviour
image

#### Actual behaviour
image

#### Backtrace

````
[backtrace]
````

When submitting logs: please submit them as text and not screenshots.

louisliu2048:
Closes #3015 .

### What does this PR do?

set eth_getLogs default value of fromBlock to "latest"


tclemos:
even though the statement in the issue is right and fromBlock must be the latest when not provided, we cannot do the fix in the getNumericBlockNumbers, we must do it isolated to the eth_getLogs method. I'll be fixing it

PR #3083, opened 2024-01-16
[WIP] AggLayer integration

nivida:
### What does this PR do?

It does add the option to settle the final proof over AggLayer with a soundness check. 

### Reviewers

Main reviewers:
- TBD

Codeowner reviewers:

- TBD


nivida:
FYI: Beethoven is still a private repository and until we put it out to the public will those CI tasks fail.

aggregator/aggregator.go
```diff
@@ -267,34 +279,139 @@ func (a *Aggregator) sendFinalProof() {
...
+			switch a.cfg.SettlementBackend {
+			case Beethoven:
+				if success := a.settleWithBeethoven(ctx, proof, inputs); !success {
+					continue
+				}
+			default:
+				if success := a.settleDirect(ctx, proof, inputs); !success {
+					continue
+				}
 			}
```
Stefan-Ethernal: WDYT about abstracting these different settlement strategies into some interface. Let's call it e.g. `ISettler` with the following function:

```go
type ISettler interface {
	Settle(ctx context.Context, proof *state.Proof, inputs types.FinalProofInputs) bool
}
```
You would have two implementations e.g. `L1Settler` and `BeethovenSettler` and instantiate the corresponding one based on the `SettlementBackend` "enumeration". The constructor function of the aggregator (`New`) would receive `ISettler` as a parameter and you could omit the trusted sequencer private key (`sequencerPrivateKey`) parameter, because it is not really needed by the aggregator per se, but only by Beethoven settling strategy. In that light, only `BeethovenSettler` would receive sequencer private key.

tclemos:
Since the PR is WIP, I'm converting it to draft until it is ready to be reviewed and merged.

aggregator/aggregator.go
nivida: Yep, totally agree. This is the correct way to implement that. On it.

nivida: I've looked closer at the aggregator code and I think we can go with that switch/case cause this will otherwise end up in a full refactoring of it. Many basic architectural principles are violated here.

config/config.go
```diff
@@ -183,3 +187,19 @@ func Load(ctx *cli.Context, loadNetworkConfig bool) (*Config, error) {
 	}
 	return cfg, nil
 }
+
+// NewKeyFromKeystore creates a private key from a keystore file
+func NewKeyFromKeystore(cfg types.KeystoreFileConfig) (*ecdsa.PrivateKey, error) {
```
nivida, line 192: We could also already apply the related changes from Arnau here. [See here.](https://github.com/0xPolygon/cdk-validium-node/blob/a03bdeb526891222e22e138d2a83408b43559c25/etherman/etherman.go#L1739)

PR #3086, opened 2024-01-16
reprocess trusted batch, use GER of leaf 0 of L1InfoRoot

### LINKED ISSUE #3085: etrog: synchronizer. If a trustedBatch don't have more L2Data but differ in something reprocess
# Rationale

If a trusted batch have same L2Data but differ in something from local one reprocess if it's a WIP .

### LINKED ISSUE #3087: etrog: Use GlobalExitRoot 0x000000 to process L1 batches that are using just the index 0 of L1InfoTree
# Rationale

First we implement that batches using leaf 0 the GER is the corresponding to leaf 0 (a leaf that are not usable for deposits). The problem is the sequencer when create a batch maybe don't have received yet the first GER and use 0, but when the batch is received by synchronizer the leaf 0 have another value and doesnt match the trustedBatch

So we agree to use GER 0x0000 for batches that use index 0 of L1InfoTree

**We known that the value of leave 0 is not use by executor, neither smc!**

joanestebanr:
Closes #3085
Closes #3087 

### Reviewers

Main reviewers:
- @ARR552


synchronizer/l2_sync/l2_shared/batch_compare.go
```diff
@@ -75,3 +109,9 @@ func AreEqualStateBatchAndTrustedBatch(stateBatch *state.Batch, trustedBatch *ty
 	}
 	return false, debugStrResult
 }
+
+func hash(data []byte) common.Hash {
+	sha := sha3.NewLegacyKeccak256()
+	sha.Write(data)
+	return common.BytesToHash(sha.Sum(nil))
+}
```
ARR552: What about putting this function in the state helper file?

ARR552 (approved):

synchronizer/l2_sync/l2_shared/batch_compare.go
joanestebanr: I agree, good idea. I'll create another PR

joanestebanr: Done in current PR

PR #3088, opened 2024-01-16
synchronizer from L1 miss match GER

### LINKED ISSUE #3087: etrog: Use GlobalExitRoot 0x000000 to process L1 batches that are using just the index 0 of L1InfoTree

joanestebanr:
Closes #3087

### Reviewers

Main reviewers:

- @ARR552


agnusmor (approved):

joanestebanr:
This change have been discarded

PR #3090, opened 2024-01-16
Return zero GER when L1IntroTreeIndex is 0

agnusmor:
### What does this PR do?

- Fixes return a zero GER when sequencer gets the last L1InfoRoot and the index is 0

### Reviewers

Main reviewers:

@joanestebanr 
@ToniRamirezM 
@ARR552

### DIFF
--- a/state/pgstatestorage/l1infotree.go
+++ b/state/pgstatestorage/l1infotree.go
@@ -64,12 +64,6 @@ func (p *PostgresStorage) GetLatestL1InfoRoot(ctx context.Context, maxBlockNumbe
 	if !errors.Is(err, pgx.ErrNoRows) {
 		return entry, err
 	}
-
-	// L1InfoTreeIndex = 0 is a special case (empty tree) therefore we will set GER as zero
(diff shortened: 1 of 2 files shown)


ToniRamirezM (approved):

ARR552 (approved):

state/pgstatestorage/l1infotree.go
```diff
@@ -65,6 +65,11 @@ func (p *PostgresStorage) GetLatestL1InfoRoot(ctx context.Context, maxBlockNumbe
 		return entry, err
 	}
 
+	// L1InfoTreeIndex = 0 is a special case (empty tree) therefore we will set GER as zero
+	if entry.L1InfoTreeIndex == 0 {
+		entry.GlobalExitRoot.GlobalExitRoot = state.ZeroHash
+	}
+
```
joanestebanr: This business logic make more sense to put in state layer instead of storage.

ToniRamirezM: I moved it into the finalizer.

joanestebanr (approved):

PR #3128, opened 2024-01-23
fix state storage initialization

### LINKED ISSUE #3125: fix state storage initialization to load the configuration properly

tclemos:
Closes #3125.

### What does this PR do?

fix state storage initialization


joanestebanr (approved):

joanestebanr:
This bug affects to current 0.5.x, isn't? because the RPC configuration pass to state is going to affect the behaviour?

tclemos:
> This bug affects to current 0.5.x, isn't? because the RPC configuration pass to state is going to affect the behaviour?

yes, but there was other PR fixing it, because the PR depended on this fix.
This is the `v0.5.3`: https://github.com/0xPolygonHermez/zkevm-node/blob/291ac8580f90fb3fa403c1443ba98fba7eba1234/cmd/run.go#L463

and here is the fix in the `v0.5.4`: https://github.com/0xPolygonHermez/zkevm-node/blob/95465ca0bdef1090431359cc2fb84b41e30484af/cmd/run.go#L483

and after cherry picking the fix from other PRs to develop, this PR is not needed anymore, so I'll close it.

PR #3166, opened 2024-01-29
Sync old network compatibility

ARR552:
### What does this PR do?

Allows sync mainnet before the upgrade with latest version

### Reviewers

- @tclemos 
- @joanestebanr

### DIFF
--- a/etherman/etherman.go
+++ b/etherman/etherman.go
@@ -246,8 +246,8 @@ func NewClient(cfg Config, l1Config L1Config) (*Client, error) {
 	// Get RollupID
 	rollupID, err := rollupManager.RollupAddressToID(&bind.CallOpts{Pending: false}, l1Config.ZkEVMAddr)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


etherman/etherman.go
```diff
@@ -1553,11 +1561,18 @@ func (etherMan *Client) EthBlockByNumber(ctx context.Context, blockNumber uint64
 
 // GetLatestBatchNumber function allows to retrieve the latest proposed batch in the smc
 func (etherMan *Client) GetLatestBatchNumber() (uint64, error) {
+	var latestBatchNum uint64
 	rollupData, err := etherMan.RollupManager.RollupIDToRollupData(&bind.CallOpts{Pending: false}, etherMan.RollupID)
 	if err != nil {
-		return 0, err
+		log.Warn("error getting latestBatchNum from rollupManager. Trying old zkevm smc... Error: ", err)
```
joanestebanr: Using old contracts are going to fill logs with Warning, maybe we can decide which one to use (configuration or in runtime) and just call to the right one

joanestebanr (approved):

PR #3169, opened 2024-01-30
synchronizer: update fromTrusted fix cache update

joanestebanr:
Closes #<issue number>.

### What does this PR do?
- It was showing some warning when trusted batch is closed and there are no update
- Also fix update cache that was not working as expected

### Reviewers

Main reviewers:

- @ARR552 
- @tclemos 
- 
Codeowner reviewers:

- @-Alice
- @-Bob


ARR552 (approved):

ARR552:
When I read the word cache I thought it was a real cache but I see it is the same mechanism the synchronizer used to use, keeping in memory the array of the block and prevBlock

PR #3182, opened 2024-01-31
fix migration down; check for greater forkIDs instead of the next +1

### LINKED ISSUE #3176: Enable non consecutive ForkID jumps
# Rationale

Right now the node prevents fork ID upgrades that are not consecutive in number:

```go
                // Check to detect forkID inconsistencies
		if forkIDs[len(forkIDs)-1].ForkId+1 != newForkID.ForkId {
			log.Errorf("error checking forkID sequence. Last ForkID stored: %d. New ForkID received: %d", forkIDs[len(forkIDs)-1].ForkId, newForkID.ForkId)
			return fmt.Errorf("error checking forkID sequence. Last ForkID stored: %d. New ForkID received: %d", forkIDs[len(forkIDs)-1].ForkId, newForkID.ForkId)
		}
```

To allow CDKs that want to move from forkID X to fork ID Y where `Y !== X+1`

# Implementation

Just remove the code mentioned above, or maybe change log level to warning and do not through error

tclemos:
Closes #3176.

### What does this PR do?

fix migration down; check for greater forkIDs instead of the next +1

### DIFF
--- a/state/pgstatestorage/forkid.go
+++ b/state/pgstatestorage/forkid.go
@@ -81,9 +81,11 @@ func (p *PostgresStorage) AddForkIDInterval(ctx context.Context, newForkID state
 		var forkIDs []state.ForkIDInterval
 		forkIDs = oldForkIDs
 		// Check to detect forkID inconsistencies
-		if forkIDs[len(forkIDs)-1].ForkId+1 != newForkID.ForkId {
-			log.Errorf("error checking forkID sequence. Last ForkID stored: %d. New ForkID received: %d", forkIDs[len(forkIDs)-1].ForkId, newForkID.ForkId)
-			return fmt.Errorf("error checking forkID sequence. Last ForkID stored: %d. New ForkID received: %d", forkIDs[len(forkIDs)-1].ForkId, newForkID.ForkId)
+		if forkIDs[len(forkIDs)-1].ForkId >= newForkID.ForkId {
+			errMsg := "error checking forkID sequence. Last ForkID stored: %d. New ForkID received: %d"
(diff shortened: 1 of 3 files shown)


state/pgstatestorage/forkid.go
```diff
@@ -81,9 +81,11 @@ func (p *PostgresStorage) AddForkIDInterval(ctx context.Context, newForkID state
 		var forkIDs []state.ForkIDInterval
 		forkIDs = oldForkIDs
 		// Check to detect forkID inconsistencies
-		if forkIDs[len(forkIDs)-1].ForkId+1 != newForkID.ForkId {
-			log.Errorf("error checking forkID sequence. Last ForkID stored: %d. New ForkID received: %d", forkIDs[len(forkIDs)-1].ForkId, newForkID.ForkId)
-			return fmt.Errorf("error checking forkID sequence. Last ForkID stored: %d. New ForkID received: %d", forkIDs[len(forkIDs)-1].ForkId, newForkID.ForkId)
+		if forkIDs[len(forkIDs)-1].ForkId >= newForkID.ForkId {
```
arnaubennassar, line 84: While this fixes "my" problem, I believe that this restriction shouldn't be here. Conceptually, the node is an implementation of the protocol, and the protocol allow to move from ForkID X to Y, where Y < X, and even X == Y.

FYI @invocamanman

arnaubennassar (approved):
Approving, but would conisder implementing this in a different way (just logging that the ForkID is not +1 from previous FID)

ARR552 (approved):

PR #3190, opened 2024-02-02
fix l2 block gas limit to use value returned by the executor (#3186)

### LINKED ISSUE #3187: cherry-pick PR 3186 into release/v0.5.4
cherry-pick PR #3186 into release/v0.5.4

tclemos:
Closes #3187.

### What does this PR do?

fix l2 block gas limit to use value returned by the executor (#3186)


test/benchmarks/sequencer/common/setup/setup.go
```diff
@@ -31,7 +32,7 @@ var (
 	bc = state.BatchConstraintsCfg{
 		MaxTxsPerBatch:       300,
 		MaxBatchBytesSize:    120000,
-		MaxCumulativeGasUsed: 30000000,
+		MaxCumulativeGasUsed: uint64(math.MaxInt64),
```
agnusmor, line 35: Isn't it better to set the constant value 9223372036854775807 and thus be aligned with the rest? This may help in the future if you look for where that value 9223372036854775807 is used

tclemos: well, I understand the reasoning and yes, we can do that, but this is a cerry-pick of the develop branch and fixing it here would be pointless if we don't fix it also in the develop branch, also this change is not part of the node itself, it's just a test benchmark script, so I don't feel we need to do that also considering how urgent the fix is...

I mean, yes, we could, but the constant is exactly the same value and I think it doesn't worth our time invested to fix in here and in the develop branch.

My suggestion for now is that we just merge it and move forward.

agnusmor: Ok!

agnusmor (approved):

PR #3191, opened 2024-02-02
sync: fix getting URL from contract in compatibility mode, and set sequential by default

### LINKED ISSUE #2806: Synchronized must use by default the sequential mode
# Rationale

The parallel mode is still pending to be verified, for that reason the default must be the sequential mode

# Implementation

Change config

joanestebanr:
Closes #2806

### What does this PR do?
- If are not able to get trustedURL from contract ask to oldContract
- Set sync from L1 mode to `sequential` by default

### Reviewers

Main reviewers:

-- @ARR552 

Codeowner reviewers:

- @-Alice
- @-Bob


etherman/etherman.go
```diff
@@ -1665,7 +1665,13 @@ func (etherMan *Client) ApprovePol(ctx context.Context, account common.Address,
 
 // GetTrustedSequencerURL Gets the trusted sequencer url from rollup smc
 func (etherMan *Client) GetTrustedSequencerURL() (string, error) {
-	return etherMan.ZkEVM.TrustedSequencerURL(&bind.CallOpts{Pending: false})
+	url, err := etherMan.ZkEVM.TrustedSequencerURL(&bind.CallOpts{Pending: false})
+	//TODO: remove this code because is for compatibility with oldZkEVM
+	if err != nil || (len(url) == 0) {
+		// Getting from oldZkEVM Contract
```
ARR552, line 1671: I think it would be nice to add a log here (debug level)

etherman/etherman.go
```diff
@@ -1665,7 +1665,13 @@ func (etherMan *Client) ApprovePol(ctx context.Context, account common.Address,
...
+		return etherMan.OldZkEVM.TrustedSequencerURL(&bind.CallOpts{Pending: false})
+	}
+	return url, err
```
ARR552: this err will always be nil

etherman/etherman.go
```diff
@@ -1665,7 +1665,13 @@ func (etherMan *Client) ApprovePol(ctx context.Context, account common.Address,
 
 // GetTrustedSequencerURL Gets the trusted sequencer url from rollup smc
 func (etherMan *Client) GetTrustedSequencerURL() (string, error) {
-	return etherMan.ZkEVM.TrustedSequencerURL(&bind.CallOpts{Pending: false})
+	url, err := etherMan.ZkEVM.TrustedSequencerURL(&bind.CallOpts{Pending: false})
+	//TODO: remove this code because is for compatibility with oldZkEVM
+	if err != nil || (len(url) == 0) {
```
ARR552:
```suggestion
	if err != nil || url == "" {
```

ARR552 (changes requested):

etherman/etherman.go
joanestebanr: what do you suggest: `return url, nil`?

ARR552 (approved):

PR #3320, opened 2024-02-20
Nonce too low fix

goran-ethernal:
### What does this PR do?

While testing the CDK, a `nonce too low` error was quite frequent when agglayer was under load.

Basically, what happened was, there was a test account that was sending a number of transaction to the agglayer rpc in a short period of time. Since the agglayer uses the `ethTxManager` from the zkevm node, the behavior was like this:
- a transaction is sent to the agglayer rpc
- some validation is done on the agglayer
- agglayer adds the transaction to the `ethTxManager` using the `Add` function.
- `ethTxManager` sets the nonce of all the newly arrived transactions from the same account to the same nonce (CurrentNonce gotten from the L1 node), and those transactions are stored in postgress db to be then sent to the L1 node.
- a timer ticks and those newly stored transactions are now sent to the L1 node, but only the first one was successful, since the nonce was not updated for the other transactions from that account.

The solution is to always update the nonce of pending transactions (transactions to be either sent or resent), by using a local map of nonces. Only if the account does not exist in the map, the L1 node will be queried to return the current nonce for that account. This is of course only done for the one iteration of timer for monitoring transactions. The local nonce map is always recreated for each iteration of monitoring to avoid unnecessary memory consumption.

### Reviewers

Main reviewers:

- @John
- @Doe

Codeowner reviewers:

- @-Alice
- @-Bob

### DIFF
--- a/ethtxmanager/ethtxmanager.go
+++ b/ethtxmanager/ethtxmanager.go
@@ -36,39 +36,37 @@ var (
 	ErrExecutionReverted = errors.New("execution reverted")
 )
 
-// nonceMap is a struct that represents a mapping of Ethereum addresses to nonces.
-type nonceMap struct {
+// nonceStore is a struct that represents a mapping of Ethereum addresses to nonces.
+type nonceStore struct {
 	m map[common.Address]uint64
 
 	lock sync.RWMutex
 }
 
-// newNonceMap creates a new instance of nonceMap.
-// It returns a pointer to the newly created nonceMap.
-func newNonceMap() *nonceMap {
-	return &nonceMap{
+// newNonceStore creates a new instance of nonceStore.
+// It returns a pointer to the newly created nonceStore.
+func newNonceStore() *nonceStore {
+	return &nonceStore{
 		m: make(map[common.Address]uint64, 0),
 	}
 }
 
 // get retrieves the nonce value for the given address.
 // It returns the nonce value and a boolean indicating whether the nonce exists in the map.
-func (n *nonceMap) get(addr common.Address) (uint64, bool) {
+func (n *nonceStore) get(addr common.Address) (uint64, bool) {
 	n.lock.RLock()
-	defer n.lock.RUnlock()
-
 	nonce, hasNonce := n.m[addr]
+	n.lock.RUnlock()
 
 	return nonce, hasNonce
 }
 
 // set updates the nonce value for the given address in the nonceMap.
 // It acquires a lock to ensure thread safety and then updates the nonce value.
-func (n *nonceMap) set(addr common.Address, nonce uint64) {
+func (n *nonceStore) set(addr common.Address, nonce uint64) {
 	n.lock.Lock()
-	defer n.lock.Unlock()
-
 	n.m[addr] = nonce
+	n.lock.Unlock()
 }
 
 // Client for eth tx manager
@@ -282,7 +280,7 @@ func (c *Client) monitorTxs(ctx context.Context) error {
 
 	log.Infof("found %v monitored tx to process", len(mTxs))
 
-	nonceMap := newNonceMap()
+	nonceMap := newNonceStore()
 
 	wg := sync.WaitGroup{}
 	wg.Add(len(mTxs))
@@ -305,7 +303,7 @@ func (c *Client) monitorTxs(ctx context.Context) error {
 }
 
 // monitorTx does all the monitoring steps to the monitored tx
-func (c *Client) monitorTx(ctx context.Context, mTx monitoredTx, logger *log.Logger, nonceMap *nonceMap) {
+func (c *Client) monitorTx(ctx context.Context, mTx monitoredTx, logger *log.Logger, nonceMap *nonceStore) {
 	var err error
 	logger.Info("processing")
 	// check if any of the txs in the history was confirmed
@@ -585,7 +583,7 @@ func (c *Client) reviewMonitoredTx(ctx context.Context, mTx *monitoredTx, mTxLog
 // IMPORTANT: Nonce is reviewed apart from the other fields because it is a very
 // sensible information and can make duplicated data to be sent to the blockchain,
 // causing possible side effects and wasting resources.
-func (c *Client) reviewMonitoredTxNonce(ctx context.Context, mTx *monitoredTx, mTxLogger *log.Logger, nonceMap *nonceMap) error {
+func (c *Client) reviewMonitoredTxNonce(ctx context.Context, mTx *monitoredTx, mTxLogger *log.Logger, nonceMap *nonceStore) error {
 	mTxLogger.Debug("reviewing nonce")
 
 	var (
--- a/ethtxmanager/ethtxmanager.go
+++ b/ethtxmanager/ethtxmanager.go
@@ -36,6 +36,41 @@ var (
 	ErrExecutionReverted = errors.New("execution reverted")
 )
 
+// nonceMap is a struct that represents a mapping of Ethereum addresses to nonces.
+type nonceMap struct {
+	m map[common.Address]uint64
+
+	lock sync.RWMutex
+}
+
+// newNonceMap creates a new instance of nonceMap.
+// It returns a pointer to the newly created nonceMap.
+func newNonceMap() *nonceMap {
+	return &nonceMap{
+		m: make(map[common.Address]uint64, 0),
+	}
+}
+
+// get retrieves the nonce value for the given address.
+// It returns the nonce value and a boolean indicating whether the nonce exists in the map.
+func (n *nonceMap) get(addr common.Address) (uint64, bool) {
+	n.lock.RLock()
+	defer n.lock.RUnlock()
+
+	nonce, hasNonce := n.m[addr]
+
+	return nonce, hasNonce
+}
+
+// set updates the nonce value for the given address in the nonceMap.
+// It acquires a lock to ensure thread safety and then updates the nonce value.
+func (n *nonceMap) set(addr common.Address, nonce uint64) {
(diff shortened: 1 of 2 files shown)


goran-ethernal:
I have read the CLA Document and I hereby sign the CLA

ethtxmanager/ethtxmanager.go
```diff
@@ -36,6 +36,41 @@ var (
 	ErrExecutionReverted = errors.New("execution reverted")
 )
 
+// nonceMap is a struct that represents a mapping of Ethereum addresses to nonces.
+type nonceMap struct {
```
begmaroman: "nonceStore"

ethtxmanager/ethtxmanager.go
```diff
@@ -36,6 +36,41 @@ var (
...
+	return &nonceMap{
+		m: make(map[common.Address]uint64, 0),
+	}
+}
+
+// get retrieves the nonce value for the given address.
+// It returns the nonce value and a boolean indicating whether the nonce exists in the map.
+func (n *nonceMap) get(addr common.Address) (uint64, bool) {
+	n.lock.RLock()
+	defer n.lock.RUnlock()
```
begmaroman: The logic is very straight forward, no needed to defer here, you can just: lock, retrieve, unlock

ethtxmanager/ethtxmanager.go
```diff
@@ -36,6 +36,41 @@ var (
...
+	nonce, hasNonce := n.m[addr]
+
+	return nonce, hasNonce
+}
+
+// set updates the nonce value for the given address in the nonceMap.
+// It acquires a lock to ensure thread safety and then updates the nonce value.
+func (n *nonceMap) set(addr common.Address, nonce uint64) {
+	n.lock.Lock()
+	defer n.lock.Unlock()
```
begmaroman: same here

ethtxmanager/ethtxmanager.go
goran-ethernal: Fix: https://github.com/0xPolygonHermez/zkevm-node/pull/3320/commits/f6e92fb3753524d74123597a818980499f29ec10

ethtxmanager/ethtxmanager.go
goran-ethernal: Fix: https://github.com/0xPolygonHermez/zkevm-node/pull/3320/commits/f6e92fb3753524d74123597a818980499f29ec10

ethtxmanager/ethtxmanager.go
goran-ethernal: Fix: https://github.com/0xPolygonHermez/zkevm-node/pull/3320/commits/f6e92fb3753524d74123597a818980499f29ec10

ethtxmanager/ethtxmanager.go
```diff
@@ -254,6 +280,8 @@ func (c *Client) monitorTxs(ctx context.Context) error {
 
 	log.Infof("found %v monitored tx to process", len(mTxs))
 
+	nonceMap := newNonceStore()
```
begmaroman: Do you think it should be created per Client struct rather than passing it as a function argument

goran-ethernal: That is something to discuss. To me it is probably enough to create this `nonceStore` per monitoring cycle.
If we are to store the nonce map in the `Client` struct itself, then, I am not sure when is the right time to clean it? It might grow, and take memory space unnecessarily.

ethtxmanager/ethtxmanager.go
```diff
@@ -275,19 +303,13 @@ func (c *Client) monitorTxs(ctx context.Context) error {
 }
 
 // monitorTx does all the monitoring steps to the monitored tx
-func (c *Client) monitorTx(ctx context.Context, mTx monitoredTx, logger *log.Logger) {
+func (c *Client) monitorTx(ctx context.Context, mTx monitoredTx, logger *log.Logger, nonceMap *nonceStore) {
```
begmaroman: `nonce *nonceStore` or `nonceStore *nonceStore` or `ns *nonceStore`? 
And the same in the rest of places

begmaroman (approved):
Minor suggestion and then LGTM

ethtxmanager/ethtxmanager.go
goran-ethernal: Fix: https://github.com/0xPolygonHermez/zkevm-node/pull/3320/commits/920d167e4e3e95a0d8c0ba26d74aa387c0452354

tclemos:
Let me give you a bit of context here, so we can think more about this suggestion.

The EthTxManager was created to make sure txs would be succefully mined when facing issues with the network issues, for example, adjusting the gas limit and gas price accordingly to the current situation of the network that might be different from the moment when the tx was created.

The EthTxManager stores the raw tx data instead of storing a signed tx, this allows the system to create new transactions with the same data, but with different parameters to make the tx to be able to get mined.

In order to help the callers, all the metadata required by the transaction is calculated when the tx is added to be monitored, at this moment the component compute the suggested gas price, the gas estimation, the current nonce for the account being used to send the tx and stores all the metadata to later be able to create a signed tx.

Bear in mind that the EthTxManager doesn't care about what the txs is doing, this is the caller responsibility and the sequence of transactions being enqueued by the caller to be monitored should be controlled by the caller.

It's also important to mention that the EthTxManger doesn't "fix" the nonce automatically, because this is exaclty the mechanism it uses to protect duplicated data to be sent to the network and also to avoid balance drain in case of a malfunction. The EthTxManager has a special rule to "fix" the nonce only when it detects the tx was reverted due to the metadata computed but its own rules. Let's say the EthTxManager computed a gasLimit that was not able to fullfill the needs to pay the tx and it gets reverted due the `Out Of Gas` issue, in this case the nonce used in the TX was consumed, but the EthTxManager still can increase the gasLimit and nonce to generate a new tx with the new nonce in order to allow this TX to get mined. 

Talking about the zkEVM trusted node, the Sequencer and Aggregator implementations today just add new txs to be monitored when the previous one was already mined, this is the way these components work and this is probably why we don't face this situation in the network very often.

Theoretically we could just change the nonce calculation to take into account the txs in the pool when a new TX is added to be monitored, this would make each tx to get a different nonce. It's basically the same solution as you proposed, but using the network as the address map instead of keeping it in memory. Basically we need to expose this functionality via Etherman and instead of calling the CurrentNonce method, we would call the PendingNonce for that address.

The point here is that the caller still needs to be able to recover from the situation where one of the nonces is consumed by a reverted tx, causing the other transactions to be able to get mined without having the previous one to get mined. This is very important, let's create and example to understand better what can help: 
- lets say the Sequencer created the sequences 1 and 2
- the sequence 2 should be added to the network only after the sequence 1 is added, but since the caller is the responsible of controlling actions, let's assume the sequencer already created both sequence transactions and added them to be monitored without waiting for one of them to get mined.
  - the sequence 1 has the nonce 1
  - the sequence 2 has the nonce 2
- since both txs have different nonces, the EthTxManager can add both to the network, so both are sent succesfully to the network pool and now are waiting to get mined.
- during the mining process, the sequence 1 gets reverted, the nonce 1 is consumed, the gas is paid and the balance of the account is consumed.
- since both txs were sent to the network and nonce 1 was already mined, the sequence 2 is now going to be able to get mined, and will probably get reverted also, because it depends on sequence 1.

The most important thing is, we need to think about how the callers are managing the transactions in order to be able react to side cases, which is the most important feature of the EthTxManager, the system is prepared to react when something goes wrong, this is the main reason to have this component, not the opposite.

For callers that have txs that are not dependent, it would be ok to add multiple txs to be monitored at the same time, because if one of them revert it doesn't affect the other txs already in the pool, but for callers that have dependant txs, they need to be controlled by the caller in order to keep the sequence and fullfill the dependencies between them.

As mentioned in the PR description and also pointed here, everytime a tx is added to be monitored, the EthTxManager computes the current nonce, and I think this is the point that is causing the issues due to the way we compute it. Currently we compute this using the current nonce without considering txs that are still in the pool. With that said, if the caller adds 2 transactions to be monitored before the first one gets mined, both will be added with the same nonce, because the current nonce for the account at the moment they were added to be monitored was the same.

In theory we must update the EthTxManager to use the PendingNonce instead of the CurrentNonce when adding a new tx to be monitored, but as mentioned before, the caller should be able to handle this situation if it decided to send multiple transactions to be monitored and one of them gets reverted for any reason.

goran-ethernal:
> For callers that have txs that are not dependent, it would be ok to add multiple txs to be monitored at the same time, because if one of them revert it doesn't affect the other txs already in the pool, but for callers that have dependant txs, they need to be controlled by the caller in order to keep the sequence and fullfill the dependencies between them.

Completely agree with this.

> In theory we must update the EthTxManager to use the PendingNonce instead of the CurrentNonce when adding a new tx to be monitored, but as mentioned before, the caller should be able to handle this situation if it decided to send multiple transactions to be monitored and one of them gets reverted for any reason.

Thank you @tclemos for this detailed comment. I agree with you, `PendingNonce` should be checked either way. Our concern was, and the use case we stumbled upon on the testnet, was that the same user can send unrelated txs to be monitored in the same time. So let's say two transactions from the same user get added to the eth tx manager before the another monitoring cycle gets executed, even with pending nonce fix, they will still face the same issue (getting the same nonce).

Maybe the good idea would be this, similar to what you suggested:
- Change the `Add` function to call the `PendingNonce` instead of `CurrentNonce`.
- Have some function in storage that returns all the monitored txs from given sender that have status `MonitoredTxStatusCreated` and update the nonce according to them (not sure if this would be expensive, but since it is a local storage, it probably isn't).

PR #3338, opened 2024-02-21
Nonce too low fix

begmaroman:
### What does this PR do?

While testing the CDK, a `nonce too low` error was quite frequent when agglayer was under load.

Basically, what happened was, there was a test account that was sending a number of transaction to the agglayer rpc in a short period of time. Since the agglayer uses the `ethTxManager` from the zkevm node, the behavior was like this:
- a transaction is sent to the agglayer rpc
- some validation is done on the agglayer
- agglayer adds the transaction to the `ethTxManager` using the `Add` function.
- `ethTxManager` sets the nonce of all the newly arrived transactions from the same account to the same nonce (CurrentNonce gotten from the L1 node), and those transactions are stored in postgress db to be then sent to the L1 node.
- a timer ticks and those newly stored transactions are now sent to the L1 node, but only the first one was successful, since the nonce was not updated for the other transactions from that account.

The solution is to always update the nonce of pending transactions (transactions to be either sent or resent), by using a local map of nonces. Only if the account does not exist in the map, the L1 node will be queried to return the current nonce for that account. This is of course only done for the one iteration of timer for monitoring transactions. The local nonce map is always recreated for each iteration of monitoring to avoid unnecessary memory consumption.


tclemos:
please, check the explanation here: https://github.com/0xPolygonHermez/zkevm-node/pull/3320#issuecomment-1956535402

ethtxmanager/pgstorage.go
```diff
@@ -123,6 +123,50 @@ func (s *PostgresStorage) GetByStatus(ctx context.Context, owner *string, status
 	return mTxs, nil
 }
 
+// GetBySenderAndStatus loads all monitored txs of the given sender that match the provided status
+func (s *PostgresStorage) GetBySenderAndStatus(ctx context.Context, sender string, statuses []MonitoredTxStatus, dbTx pgx.Tx) ([]monitoredTx, error) {
```
tclemos: would you mind replacing the `sender` type from `string` to `common.Address`? 
this would force the parameter to always be an Address instead of any string, helping the consumers of this method to understand this is an Address.

tclemos (changes requested):
here is the code that updates the monitored tx nonce in case the EthTxManager detects it needs to be updated:
https://github.com/0xPolygonHermez/zkevm-node/blob/0290a295e936fcfe273658838ae9405c6c632160/ethtxmanager/ethtxmanager.go#L595-L610

We must to replace the call to the `etherman.CurrentNoce` by the new `getNonce` method, but keep it in mind that it will make the current transaction to be moved to the last position of the transactions in the pool for this sender. Depending on the scenario, if the transactions are dependant, this will probably not work, but the caller still have the control to not add multiple txs, so I think it's ok to fix the code in this direction for now.

ethtxmanager/pgstorage.go
begmaroman: Done

tclemos (approved):

ToniRamirezM (approved):

PR #3370, opened 2024-02-26
synchronizer accepts empty closed batches from trusted 

### LINKED ISSUE #3265: If synchronizer receive a empty closed batch  (0 bytes) from trustedNode must be store without checking anything
#### System information

zkEVM Node version: `v0.5.8`
Network: `Mainnet`

#### Backtrace

````
[backtrace]
````

When submitting logs: please submit them as text and not screenshots.

### PR DESCRIPTION
**Pull request by joanestebanr**
Closes #3265

If synchronizer receive a empty closed batch (0 bytes) from trustedNode must be store without checking anything

It could be configure in the node config file:

```
[Synchronizer]
(...)
	[Synchronizer.L2Synchronization]
		AcceptEmptyClosedBatches = false
```

### What does this PR do?

Main reviewers:

- @tclemos 
- @agnusmor 
- @ToniRamirezM 

Codeowner reviewers:

- @-Alice
- @-Bob

### CODE REVIEW

File: jsonrpc/client/zkevm_test.go
Comment (tclemos) [inline jsonrpc/client/zkevm_test.go:1]: I understand this test needs to be either converted into a unit, e2e test or removed.
Acccessing the public URL forces the CI to have access to the internet in order to be able to run the test, also the test is not asserting the results, but printing it to the console. Looks like to me it was using as a debug tool during the development, but shouldn't be pushed to the repo.

File: synchronizer/l2_sync/l2_shared/processor_trusted_batch_sync.go [hunk 1]
```diff
@@ -308,23 +309,39 @@ func (s *ProcessorTrustedBatchSync) GetModeForProcessBatch(trustedNodeBatch *typ
...
 	result.OldAccInputHash = statePreviousBatch.AccInputHash
 	result.Now = s.timeProvider.Now()
 	result.DebugPrefix = fmt.Sprintf("%s mode %s:", debugPrefix, result.Mode)
+
+	if isTrustedBatchEmptyAndClosed(trustedNodeBatch) {
+		if s.Cfg.AcceptEmptyClosedBatches {
+			log.Infof("%s Batch %v: TrustedBatch Empty and closed, accepted due configuration", result.DebugPrefix, trustedNodeBatch.Number)
+		} else {
+			log.Infof("%s Batch %v: TrustedBatch Empty and closed, rejected due configuration", result.DebugPrefix, trustedNodeBatch.Number)
+			return result, fmt.Errorf("%s Batch %v: TrustedBatch Empty and closed, rejected due configuration", result.DebugPrefix, trustedNodeBatch.Number)
```
tclemos: shouldn't we return `nil` instead of `result` here?

synchronizer/l2_sync/l2_shared/processor_trusted_batch_sync.go
```diff
@@ -339,7 +356,12 @@ func checkProcessBatchResultMatchExpected(data *ProcessData, processBatchResp *s
 	var err error = nil
 	var trustedBatch = data.TrustedBatch
 	if trustedBatch == nil {
-		panic("trustedBatch is nil")
+		log.Error("trustedBatch is nil, it never should be nil")
+		return fmt.Errorf("trustedBatch is nil, it never should be nil")
```
tclemos: just to avoid duplicated string, we could replace it by: 
```go
err := fmt.Errorf("trustedBatch is nil, it never should be nil")
log.Error(err.Error())
return err
```

synchronizer/synchronizer.go
```diff
@@ -109,16 +109,19 @@ func NewSynchronizer(
 		l1EventProcessors:       nil,
 		halter:                  syncCommon.NewCriticalErrorHalt(eventLog, 5*time.Second), //nolint:gomnd
 	}
-	L1SyncChecker := l2_sync_etrog.NewCheckSyncStatusToProcessBatch(res.zkEVMClient, res.state)
 
-	syncTrustedStateEtrog := l2_sync_etrog.NewSyncTrustedBatchExecutorForEtrog(res.zkEVMClient, res.state, res.state, res,
-		syncCommon.DefaultTimeProvider{}, L1SyncChecker)
+	if !isTrustedSequencer {
+		log.Info("Permissionless: creating L2 synchronization stuff")
```
tclemos: not sure if `stuff` word is a good one to put here in an `Info` level log message. Maybe we could be more explicit in what's being prepared, or something like:
"Permissionless: creating and Initializing L2 synchronization components"

tclemos (changes requested):

synchronizer/l2_sync/l2_shared/processor_trusted_batch_sync.go
joanestebanr: The returned type is not a pointer, so is just to return something

jsonrpc/client/zkevm_test.go
joanestebanr: The test is never execute: `t.Skip("This test is exploratory")` is just used to check the values of the RPC and make sense to keep there for future use

tclemos (approved):

PR #3463, opened 2024-03-16
Fix the out of order issue by the getLogs api

zjg555543:
Closes #<issue number>.

### Reviewers

Main reviewers:

- @John
- @Doe

Codeowner reviewers:

- @-Alice
- @-Bob


tclemos:
The fix helps with a bug introduced on Etrog where the log_index was reset for each tx in the same block instead of referencing the whole block.

This issue was already fixed on Elderberry, but since this query is one of the most expensive queries in the node, we must make sure this new field in the `order by` clause will not negatively affect the query time.

Also, it's essential to have a test that makes sure this will continue to work for all the forkIDs.

tclemos:
fix implemented here: #3476

PR #3497, opened 2024-03-25
fix aggregator l1 info tree (#3490) (#3491) (#3484)  (#3495)

ToniRamirezM:
Cherry pick #3495

### DIFF
--- a/l1infotree/tree.go
+++ b/l1infotree/tree.go
@@ -84,16 +84,13 @@ func (mt *L1InfoTree) ComputeMerkleProof(gerIndex uint32, leaves [][32]byte) ([]
 		if len(leaves)%2 == 1 {
 			leaves = append(leaves, mt.zeroHashes[h])
 		}
-		if index%2 == 1 { //If it is odd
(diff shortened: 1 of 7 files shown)


l1infotree/tree.go
```diff
@@ -84,16 +84,13 @@ func (mt *L1InfoTree) ComputeMerkleProof(gerIndex uint32, leaves [][32]byte) ([]
...
-			if len(leaves) > 1 {
-				if index >= uint32(len(leaves)) {
-					// siblings = append(siblings, mt.zeroHashes[h])
-					siblings = append(siblings, leaves[index-1])
-				} else {
-					siblings = append(siblings, leaves[index+1])
-				}
+		if index >= uint32(len(leaves)) {
+			siblings = append(siblings, mt.zeroHashes[h])
+		} else {
```
ARR552, line 89: i'm not sure about this. If the index is 7 and the tree only have 3 leaves, this will break the function. I think you should check that you are in the even branch first

agnusmor (approved):

PR #3543, opened 2024-04-08
add column checked on state.block

### LINKED ISSUE #3541: synchronizer: each block on L1 stored on DB must have a bool that says if the block need to be checked or not
# Rationale

When insert a block on node DB must mark:
- If, in the moment of the request, the block_number is over **finalized** block need to be checked
- If a second process verifiy the hash of this block when the block_number is below L1 **finalized**  block mark it

# Implementation

Add a column called **checked**:
- If is created below **L1 finalized** mark as **true** because don't need to be verified

joanestebanr:
Closes #3541
### What does this PR do?

add column **checked** to state.block

### Reviewers

Main reviewers:

- @tclemos 
- @ARR552 
- @agnusmor 

Codeowner reviewers:

- @-Alice
- @-Bob

### DIFF
--- a/db/migrations/state/0018.sql
+++ b/db/migrations/state/0018.sql
@@ -1,0 +1,7 @@
+-- +migrate Up
+ALTER TABLE state.block
+    ADD COLUMN IF NOT EXISTS checked BOOL NOT NULL DEFAULT TRUE;
+
+-- +migrate Down
+ALTER TABLE state.receipt
+    DROP COLUMN IF EXISTS checked;
--- a/db/migrations/state/0018.sql
+++ b/db/migrations/state/0018.sql
@@ -1,7 +1,11 @@
 -- +migrate Up
 ALTER TABLE state.block
-    ADD COLUMN IF NOT EXISTS checked BOOL NOT NULL DEFAULT TRUE;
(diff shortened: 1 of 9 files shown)


db/migrations/state/0018.sql
```diff
@@ -0,0 +1,7 @@
+-- +migrate Up
+ALTER TABLE state.block
+    ADD COLUMN IF NOT EXISTS checked BOOL NOT NULL DEFAULT TRUE;
```
ARR552: In a full resync it is perfect but in the current env, If we do so, maybe we miss some previous mismatch. becuase we were assuming tha it was ok. Wdyt?

db/migrations/state/0018.sql
```diff
@@ -0,0 +1,7 @@
...
+
+-- +migrate Down
+ALTER TABLE state.receipt
+    DROP COLUMN IF EXISTS checked;
```
ARR552: Can you give more context about this change?

ARR552 (approved):

db/migrations/state/0018.sql
joanestebanr: my bad, wrong table!

db/migrations/state/0018.sql
joanestebanr: I agree, I changed and the latest 1000 blocks are checked=false

PR #3546, opened 2024-04-08
synchronizer:  check l1blocks

### LINKED ISSUE #3540: synchronizer: double-check old L1 block to detect reorgs
# Rationale
Synchronizer is going to check in background the L1block after the block has been consolidated by ethereum. 

You can choose what do you consider 'consolidated' (the safety level): 
- type of block: finalized, safe or latest
- add or substract a offset

# Implementation
There are 2 goruintes that are checking in background blocks:
- One check from 'consolidated' point and if it's ok it mark as checked the block and is no longer verified
- The other check from 'consolidated' point to newest blocks, it just verify without marking and do it continously until the block is marked as checked.

So you can configure next scenario: 
- From 0  to safe: validate and mark as *checked* if match L1 RPC
- From safe to latest -5 : check in memory

## Configuration
```
	[Synchronizer.L1BlockCheck]
		Enable = true
		L1SafeBlockPoint = "finalized"
		L1SafeBlockOffset = 0
		ForceCheckBeforeStart = true
		PreCheckEnable = true
		L1PreSafeBlockPoint = "latest"
		L1PreSafeBlockOffset = -32
```
* **Enable** : it enable of disable  full feature
* **ForceCheckBeforeStart**: Before start anything it check the pending block up to 'consolidated' point 
* **L1SafeBlockPoint** + **L1SafeBlockOffset** : define what we consider as 'consolidated' point
* **PreCheckEnable**: enables second gorutine that check newest blocks continuously
* **L1PreSafeBlockPoint** + **L1PreSafeBlockOffset** : define the maximum block to check for the second gorutine

joanestebanr:
Closes #3540 #3561

### Reviewers

Main reviewers:

- @ARR552 
- @tclemos 
- @agnusmor 

Codeowner reviewers:

- @-Alice
- @-Bob


etherman/etherman.go
```diff
@@ -287,6 +287,11 @@ func NewClient(cfg Config, l1Config L1Config) (*Client, error) {
 	}, nil
 }
 
+// GetL1EthereumClient return geth client
```
tclemos: The idea of Etherman is to abstract L1 integration while keeping the responsibility to abstract Ethereum to the node in a single point.

Whether this is a good or bad decision, this is how it was planned, and exposing EthClient goes directly against the purpose of this library.

I recommend reviewing this and exposing only the necessary methods, as a plus I recommend trying to abstract as much as possible the Ethereum integration needs and expose a function that is self-explanatory instead of only exposing Geth methods directly, which doesn't make much sense, but sometimes it's hard to avoid.

synchronizer/common/syncinterfaces/async_l1_block_checker.go
```diff
@@ -0,0 +1,41 @@
...
+}
+
+type AsyncL1BlockChecker interface {
+	Run(ctx context.Context, onFinish func())
+	RunSynchronous(ctx context.Context) IterationResult
+	Stop()
+	GetResult() *IterationResult
+}
+
+type L1BlockCheckerIntegrater interface {
```
tclemos: ~~Integrater~~ -> Integrator

synchronizer/l1_check_block/async.go
```diff
@@ -0,0 +1,186 @@
...
+	}
+}
+
+// NewAsyncCheckWithPeriodTime creates a new AsyncCheck with a period time between relaunch checker.Step
+func NewAsyncCheckWithPeriodTime(checker L1BlockChecker, periodTime time.Duration) *AsyncCheck {
+	return &AsyncCheck{
+		checker:    checker,
+		periodTime: periodTime,
+	}
+}
```
tclemos: Refactoring suggestion:
```go
func (a *AsyncCheck) WithPeriodTime(periodTime time.Duration) *AsyncCheck {
	a.periodTime = periodTime
        return a
}
```

Then you use this way:
```go
ac := NewAsyncCheck().WithPeriodTime()
```
or this way:
```go
ac := NewAsyncCheck()
...
ac.WithPeriodTime()
```

synchronizer/l1_check_block/async.go
```diff
@@ -0,0 +1,186 @@
...
+const (
+	defaultPeriodTime = time.Second
+)
+
+// AsyncCheck is a wrapper for L1BlockChecker to become asynchronous
+type AsyncCheck struct {
+	checker     L1BlockChecker
+	mutex       sync.Mutex
+	lastResult  *syncinterfaces.IterationResult
+	onFnishCall func()
```
tclemos: Typo, it is missing an `i` on `Finish` word

synchronizer/l1_check_block/async.go
```diff
@@ -0,0 +1,186 @@
...
+		log.Infof("%s L1BlockChecker: finished background process", logPrefix)
+		a.Wg.Done()
+		a.mutex.Lock()
+		onFinishCall := a.onFnishCall
+		a.isRunning = false
+		a.mutex.Unlock()
+		// call onFinish function with no mutex
+		if onFinishCall != nil {
+			onFinishCall()
+		}
```
tclemos: I recommend deferring these before starting the for loop, just in case the for loop gets interrupted by a runtime error triggered by lines `128` or `130`.

synchronizer/l1_check_block/async.go
```diff
@@ -0,0 +1,186 @@
...
+	}()
+}
+
+// step is a method that executes a until executeItertion
+// returns an error or a reorg
```
tclemos: looks like we have an extra `a`: "... executes ~~a~~ until..."

synchronizer/l1_check_block/check_l1block.go
```diff
@@ -0,0 +1,146 @@
...
+)
+
+// This object check old L1block to double-check that the L1block hash is correct
+// - Get first not checked block
+// - Get last block on L1 (safe/finalized/ or minus -n)
+
+// L1Requester is an interface for GETH client
+type L1Requester interface {
+	HeaderByNumber(ctx context.Context, number *big.Int) (*types.Header, error)
+}
```
tclemos, line 24: This can be considered with the proposed Etherman refactoring about not exposing the EthClient. 

Etherman has these methods already exposed that might help:
- https://github.com/0xPolygonHermez/zkevm-node/blob/1f1550814b356034cd9b1a926126762966dc0e94/etherman/etherman.go#L1623
- https://github.com/0xPolygonHermez/zkevm-node/blob/1f1550814b356034cd9b1a926126762966dc0e94/etherman/etherman.go#L1656
- https://github.com/0xPolygonHermez/zkevm-node/blob/1f1550814b356034cd9b1a926126762966dc0e94/etherman/etherman.go#L1665-L1677

synchronizer/l1_check_block/check_l1block.go
```diff
@@ -0,0 +1,146 @@
...
+	return p.doAllBlocks(ctx, stateBlock, safeBlockNumber)
+}
+
+func (p *CheckL1BlockHash) doAllBlocks(ctx context.Context, stateBlock *state.Block, safeBlockNumber uint64) error {
+	var err error
+	startTime := time.Now()
+
+	if stateBlock == nil {
+		return fmt.Errorf("%s: DoAllBlocks stateBlock is nil! ", p.Name())
+	}
```
tclemos: If the function doesn't know what to do with a nil pointer and requires an instance to work, why not have a parameter that is not a pointer and force it by design? IMO, the struct in-memory copy is not that expensive compared to avoiding repeating this check in all the methods.

synchronizer/l1_check_block/check_l1block.go
```diff
@@ -0,0 +1,146 @@
...
+	return &CheckL1BlockHash{
+		L1Client:               l1Client,
+		State:                  state,
+		SafeBlockNumberFetcher: safeBlockNumberFetcher,
+	}
+}
+
+// Name is a method that returns the name of the checker
+func (p *CheckL1BlockHash) Name() string {
+	return logPrefix + " CheckL1Block: "
```
tclemos: this is going to return `checkL1block: CheckL1Block: `, it this the intended behavior?

tclemos (changes requested):
I don't know the scope of this task, but the PR changes seem like overkill.

I expected it to be just a single method running concurrently to check for blocks related to the consolidated point on Ethereum.

I have added many comments, but I confess I got exhausted while reviewing it because I always thought, "Is this necessary?"

I understand this PR can achieve the goal, but assuming the price we are paying to maintain this implementation in the future, I prefer to drop everything and come up with a more straightforward solution that gets the consolidated point from Ethereum and checks how we are from there.

## Proposal:

Assuming the Synchronizer must be able to recover by itself when a reorg is detected and this mechanism is a secondary protection to help the Synchronizer identify the reorg with a different strategy, I would suggest implementing the following in the simplest way possible, like:
- at the start of the application, get the current consolidated point block from Ethereum
- get all the blocks we have from the consolidated point and check all of them for a reorg
  - if reorg is detected, flag it and wait for the synchronizer to fix it. Once fixed
- stores the last block we checked in memory
- keep monitoring the consolidated point on Ethereum until the last block we checked matches
- repeat

Integration with the Synchronizer can be done via a simple channel, which can be appended to the current synchronization process.

Advantages of this approach:
no changes in the DB
faster, runs only on memory
takes advantage of the network consolidation point instead of verifying everything
with less checks, it can load all the blocks in a single shot instead of one by one and check them concurrently
way less code to maintain
way less changes in the real code due to this extra protection

### Reasoning:

We assume the consolidation point on Ethereum is where we trust a reorg will never happen; from this point, we do our own check to make sure all the blocks not consolidated yet are matching with Ethereum; once we guarantee this, the synchronizer takes place and continues his job synchronizing block by block until all the blocks we have checked are now part of the consolidated part of Ethereum, then we start over. If, for some reason, the synchronizer is not able to detect a reorg in the regular synchronization process, our next check will start from the last time we checked until the latest synchronized block, and we will find it, flagging it to the synchronizer to allow the reorg process to be executed in the next synchronizer reorg check.

### Conclusion:

I don't feel comfortable merging this whole PR, and I'm open to discussing it if you consider it worth it. Otherwise, you can check my comments in the PR and follow with this implementation.

synchronizer/l1_check_block/async.go
joanestebanr: What kind of errors? if there are a runtime error is not going to stop all process?

synchronizer/l1_check_block/integration.go
```diff
@@ -80,9 +88,46 @@ func (v *L1BlockCheckerIntegration) OnStartL2Sync(ctx context.Context) bool {
...
+
+// CheckReorgWrapper is a wrapper over reorg function of synchronizer.
+// it checks the result of the function and the result of background process and decides which return
+func (v *L1BlockCheckerIntegration) CheckReorgWrapper(ctx context.Context, reorgFirstBlockOk *state.Block, errReportedByReorgFunc error) (*state.Block, error) {
+	resultBackground := v.getMergedResults()
+	if resultBackground != nil && resultBackground.ReorgDetected {
+		// Background process detected a reorg, decide which return
+		firstOkBlockBackgroundCheck, err := v.state.GetPreviousBlockToBlockNumber(ctx, resultBackground.BlockNumber, nil)
+		if err != nil {
+			log.Warnf("%s Error getting previous block to block number where a reorg have been detected %d: %s. So we reorgFunc values", logPrefix, resultBackground.BlockNumber, err)
```
ARR552, line 105: This seems to be an error
```suggestion
			log.Errorf("%s Error getting previous block to block number where a reorg have been detected %d: %s. So we reorgFunc values", logPrefix, resultBackground.BlockNumber, err)
```

synchronizer/l1_check_block/integration.go
```diff
@@ -80,9 +88,46 @@ func (v *L1BlockCheckerIntegration) OnStartL2Sync(ctx context.Context) bool {
...
+			return firstOkBlockBackgroundCheck, nil
+		} else {
+			// Regular reorg function detected a reorg at oldest block
+			log.Warnf("%s Background checker detects bad block  at block %d (first block ok %d) and regular reorg function first block ok: %d. Executing from %d",
+				logPrefix, resultBackground.BlockNumber, firstOkBlockBackgroundCheck.BlockNumber, reorgFirstBlockOk.BlockNumber, reorgFirstBlockOk.BlockNumber)
+			return reorgFirstBlockOk, errReportedByReorgFunc
+		}
+	}
+	if resultBackground != nil && !resultBackground.ReorgDetected {
+		// Relaunch checker, if there are a reorg is going to be relaunched after (OnResetState)
```
ARR552:
```suggestion
		// Relaunch checker, if there is a reorg, It is going to be relaunched after (OnResetState)
```

synchronizer/synchronizer.go
```diff
@@ -408,19 +409,19 @@ func (s *ClientSynchronizer) Sync() error {
...
-						if s.asyncL1BlockChecker.OnStartL2Sync(s.ctx) {
-							log.Warn("s.asyncL1BlockChecker.OnStartL2Sync -> continue")
-							continue
-						}
+					//Sync Trusted State
+					resetDone, lastEthBlockSynced, err = s.checkReorgAndExecuteReset(lastEthBlockSynced)
+					if resetDone || err != nil {
+						log.Infof("Reset done before L2 sync")
+						continue
 					}
```
ARR552: I think err should be handled. At least the error must be printed even if it is not returned
```suggestion
					if resetDone {
						log.Infof("Reset done before L2 sync")
						continue
					} else if || err != nil {
					       log.Error("error in checkReorgAndExecuteReset function. Error: ", err)
					       return err
```

synchronizer/synchronizer.go
```diff
@@ -450,24 +451,23 @@ func (s *ClientSynchronizer) Sync() error {
...
 			}
 			//Sync L1Blocks
-			if s.asyncL1BlockChecker != nil {
-				if s.asyncL1BlockChecker.OnStartL1Sync(s.ctx) {
-					log.Warn("s.asyncL1BlockChecker.OnStartL1Sync -> continue")
-					continue
-				}
+			resetDone, lastEthBlockSynced, err := s.checkReorgAndExecuteReset(lastEthBlockSynced)
+			if resetDone || err != nil {
+				continue
```
ARR552, line 455: same here

synchronizer/synchronizer.go
```diff
@@ -540,22 +540,7 @@ func sanityCheckForGenesisBlockRollupInfo(blocks []etherman.Block, order map[com
...
-		log.Infof("reorg detected. Resetting the state from block %v to block %v", lastEthBlockSynced.BlockNumber, block.BlockNumber)
-		err = s.resetState(block.BlockNumber)
-		if err != nil {
-			log.Errorf("error resetting the state to a previous block. Retrying... Err: %v", err)
-			s.l1SyncOrchestration.Reset(lastEthBlockSynced.BlockNumber)
-			return lastEthBlockSynced, fmt.Errorf("error resetting the state to a previous block")
-		}
-		return block, nil
-	}
+	// MOVED checkReorg to main sync function
```
ARR552: this comment is not needed
```suggestion
```

synchronizer/synchronizer.go
```diff
@@ -827,6 +801,43 @@ func (s *ClientSynchronizer) ExecuteReorg(blockNumber uint64, reason string) err
 	return s.resetState(blockNumber - 1)
 }
 
+// checkReorgAndExecuteReset function will check if there is a reorg and execute the reset
+// returns true is reset have been done
+func (s *ClientSynchronizer) checkReorgAndExecuteReset(lastEthBlockSynced *state.Block) (bool, *state.Block, error) {
+	var err error
+	if lastEthBlockSynced == nil {
```
ARR552: this can't be nil. If this happens Fatal

synchronizer/synchronizer.go
```diff
@@ -827,6 +801,43 @@ func (s *ClientSynchronizer) ExecuteReorg(blockNumber uint64, reason string) err
...
+		lastEthBlockSynced, err = s.state.GetLastBlock(s.ctx, nil)
+		if err != nil {
+			log.Errorf("error getting last block synced from db. Error: %v", err)
+			return true, nil, err
```
ARR552: This only must be do it when the synchronizer starts

ARR552 (changes requested):

synchronizer/l1_check_block/check_l1block.go
```diff
@@ -0,0 +1,143 @@
...
+	l1Block, err := L1Client.HeaderByNumber(ctx, big.NewInt(int64(stateBlock.BlockNumber)))
+	if err != nil {
+		return err
+	}
+	if l1Block == nil {
+		err = fmt.Errorf("%s request of block: %d to L1 returns a nil", checkerName, stateBlock.BlockNumber)
+		log.Error(err.Error())
+		return err
+	}
+	if l1Block.Hash() != stateBlock.BlockHash {
```
ARR552, line 136: I know this shouldn't be necessary but What about checking also the ParentHash? You already have the information. It is just another if

synchronizer/l1_check_block/common.go
```diff
@@ -0,0 +1,5 @@
+package l1_check_block
+
+const (
+	logPrefix = "checkL1block:"
+)
```
ARR552, line 5: Another way of doing this would be:
```
checkerLog := log.WithFields("checkL1block:", block.BlockNumber)
checkerLog.Errorf("error doing something %v", err) 
```

synchronizer/synchronizer.go
```diff
@@ -123,6 +125,37 @@ func NewSynchronizer(
 		syncBlockProtection:           syncBlockProtection,
 		halter:                        syncCommon.NewCriticalErrorHalt(eventLog, 5*time.Second), //nolint:gomnd
 	}
+	if cfg.L1BlockCheck.Enable {
+		log.Infof("L1BlockChecker enabled: %s", cfg.L1BlockCheck.String())
+		l1requester, ok := res.etherMan.(l1_check_block.L1Requester)
```
ARR552: I think this wrapper does not make any sense. You could just use the etherman and specify just the method that you eed in the interface

synchronizer/synchronizer.go
```diff
@@ -372,13 +409,20 @@ func (s *ClientSynchronizer) Sync() error {
...
 			if latestSyncedBatch >= latestSequencedBatchNumber {
 				startTrusted := time.Now()
 				if s.syncTrustedStateExecutor != nil && !s.isTrustedSequencer {
 					log.Info("Syncing trusted state (permissionless)")
+					//Sync Trusted State
+					resetDone, lastEthBlockSynced, err = s.checkReorgAndExecuteReset(lastEthBlockSynced)
+					if resetDone || err != nil {
+						log.Infof("Reset done before L2 sync")
+						continue
+					}
```
ARR552, line 420: same here

synchronizer/synchronizer.go
```diff
@@ -407,18 +451,23 @@ func (s *ClientSynchronizer) Sync() error {
...
+			resetDone, lastEthBlockSynced, err := s.checkReorgAndExecuteReset(lastEthBlockSynced)
+			if resetDone || err != nil {
+				continue
+			}
+
 			startL1 := time.Now()
 			if s.l1SyncOrchestration != nil && (latestSyncedBatch < latestSequencedBatchNumber || !s.cfg.L1ParallelSynchronization.FallbackToSequentialModeOnSynchronized) {
 				log.Infof("Syncing L1 blocks in parallel lastEthBlockSynced=%d", lastEthBlockSynced.BlockNumber)
-				lastEthBlockSynced, err = s.syncBlocksParallel(lastEthBlockSynced)
+				_, err = s.syncBlocksParallel(lastEthBlockSynced)
```
ARR552: Why this change? I think we need that variable

ARR552 (changes requested):

joanestebanr:
> I don't know the scope of this task, but the PR changes seem like overkill.
> 
> I expected it to be just a single method running concurrently to check for blocks related to the consolidated point on Ethereum.
> 
> I have added many comments, but I confess I got exhausted while reviewing it because I always thought, "Is this necessary?"
> 
> I understand this PR can achieve the goal, but assuming the price we are paying to maintain this implementation in the future, I prefer to drop everything and come up with a more straightforward solution that gets the consolidated point from Ethereum and checks how we are from there.
> 
> ## Proposal:
> Assuming the Synchronizer must be able to recover by itself when a reorg is detected and this mechanism is a secondary protection to help the Synchronizer identify the reorg with a different strategy, I would suggest implementing the following in the simplest way possible, like:
> 
> * at the start of the application, get the current consolidated point block from Ethereum
> * get all the blocks we have from the consolidated point and check all of them for a reorg
>   
>   * if reorg is detected, flag it and wait for the synchronizer to fix it. Once fixed
> * stores the last block we checked in memory
> * keep monitoring the consolidated point on Ethereum until the last block we checked matches
> * repeat
> 
> Integration with the Synchronizer can be done via a simple channel, which can be appended to the current synchronization process.
> 
> Advantages of this approach: no changes in the DB faster, runs only on memory takes advantage of the network consolidation point instead of verifying everything with less checks, it can load all the blocks in a single shot instead of one by one and check them concurrently way less code to maintain way less changes in the real code due to this extra protection
> 
> ### Reasoning:
> We assume the consolidation point on Ethereum is where we trust a reorg will never happen; from this point, we do our own check to make sure all the blocks not consolidated yet are matching with Ethereum; once we guarantee this, the synchronizer takes place and continues his job synchronizing block by block until all the blocks we have checked are now part of the consolidated part of Ethereum, then we start over. If, for some reason, the synchronizer is not able to detect a reorg in the regular synchronization process, our next check will start from the last time we checked until the latest synchronized block, and we will find it, flagging it to the synchronizer to allow the reorg process to be executed in the next synchronizer reorg check.
> 
> ### Conclusion:
> I don't feel comfortable merging this whole PR, and I'm open to discussing it if you consider it worth it. Otherwise, you can check my comments in the PR and follow with this implementation.

I double-check the scope of the task and it's the expected behaviour

synchronizer/l1_check_block/integration.go
joanestebanr: It's not an error, is just because is  asynchronuous process and have missed some reset done in some point, it need to resync with stateDB

synchronizer/l1_check_block/check_l1block.go
joanestebanr: BlockHash includes ParentHash, isnt?

synchronizer/l1_check_block/common.go
joanestebanr: yes, you mention sometimes, but I need time to learn log object to understand how to do it that way

synchronizer/synchronizer.go
joanestebanr: totally agree! good catch!

synchronizer/synchronizer.go
joanestebanr: The same as before, are logged in the function `checkReorgAndExecuteReset`

synchronizer/l1_check_block/check_l1block.go
ARR552: Yes, it is not mandatory but as you have the parentHash, it is easy to check. Again, it is not mandatory so up to you

synchronizer/synchronizer.go
joanestebanr: you are right, I got an error of lint `SA4006: this value of `lastEthBlockSynced` is never used (staticcheck)` but I was because a previous error!

synchronizer/synchronizer.go
joanestebanr: I'll change for a error

synchronizer/synchronizer.go
```diff
@@ -825,8 +818,14 @@ func (s *ClientSynchronizer) checkReorgAndExecuteReset(lastEthBlockSynced *state
 			log.Errorf("error resetting the state to a previous block. Retrying... Err: %s", err.Error())
 			return false, lastEthBlockSynced, fmt.Errorf("error resetting the state to a previous block. Err: %w", err)
 		}
-		return true, block, nil
+		newLastBlock, err := s.state.GetLastBlock(s.ctx, nil)
+		if err != nil {
+			log.Warnf("error getting last block synced from db, returning expected block %d. Error: %v", block.BlockNumber, err)
+			return false, block, err
+		}
+		return true, newLastBlock, nil
```
ARR552: I don't understand this change. I think it is redundant

synchronizer/config.go
```diff
@@ -33,6 +33,10 @@ type Config struct {
 	L1ParallelSynchronization L1ParallelSynchronizationConfig
 	// L2Synchronization Configuration for L2 synchronization
 	L2Synchronization l2_sync.Config `mapstructure:"L2Synchronization"`
+
+	// dontDoReorgCheckBeforeL2Sync if is true then the reorg check is not done before the L2 sync
+	// this is a private field, can not be configured
+	dontDoReorgCheckBeforeL2Sync bool
```
ARR552: why is this param needed?

synchronizer/synchronizer.go
joanestebanr: yes, it's redudant, but it a double-check that we after reset we got the expected block

synchronizer/config.go
joanestebanr: The unittest is failing because there is a extra call to CheckReorg before starting L2 sync. With this flag we skip this

synchronizer/l1_check_block/check_l1block.go
joanestebanr: added parentHash on log

synchronizer/synchronizer.go
```diff
@@ -428,10 +428,14 @@ func (s *ClientSynchronizer) Sync() error {
...
-							err = s.resetState(l1BlockNumber - 1)
-							if err != nil {
-								log.Errorf("error resetting the state to a discrepancy block. Retrying... Err: %v", err)
-								continue
+							for {
+								resetDone, lastEthBlockSynced, err = s.detectedReorgBadBlockExecuteReset(lastEthBlockSynced, syncCommon.GetReorgErrorBlockNumber(err))
+								if resetDone {
+									break
+								} else {
+									log.Errorf("reorg doesnt done, retrying...")
```
ARR552:
```suggestion
									log.Errorf("reorg isn't done, retrying...")
```

synchronizer/synchronizer.go
```diff
@@ -428,10 +428,14 @@ func (s *ClientSynchronizer) Sync() error {
 						if errors.Is(err, syncinterfaces.ErrFatalDesyncFromL1) {
 							l1BlockNumber := err.(*l2_shared.DeSyncPermissionlessAndTrustedNodeError).L1BlockNumber
 							log.Error("Trusted and permissionless desync! reseting to last common point: L1Block (%d-1)", l1BlockNumber)
-							err = s.resetState(l1BlockNumber - 1)
-							if err != nil {
-								log.Errorf("error resetting the state to a discrepancy block. Retrying... Err: %v", err)
-								continue
+							for {
```
ARR552, line 429: Is it good to block it forever?? Maybe 10 retries or something like that

PR #3553, opened 2024-04-10
Feature/#3549 reorgs improvement

### LINKED ISSUE #3549: Check reorg refactor
Instead of checking the reorg in a different query, it should be done in the same query that reads the new data.
Old way: 
- check reorg for latest synced block
- get new data from lastes synced block + 1

New way: 
- get new data from latest synced block

The initial block should contain the same info that the node synced previously. If not, reorg detected

### LINKED ISSUE #3557: Fix e2e test "with gasPrice set and without from address" in the rpc
Fix and uncomment rpc e2e test case called "with gasPrice set and without from address".
For this specific test, new geth version returns no error. However, the zkevm node is returning the next error:
```
&rpc.jsonError{Code:-32000, Message:"gas required exceeds allowance (0)", Data:interface {}(nil)}
```

ARR552:
Closes #3549
Closes #3557

### What does this PR do?

This PR does a refactor in the way the reorgs are checked to be protected against inconsistent responses from the L1 providers

### Reviewers

- @joanestebanr 
- @tclemos


test/e2e/jsonrpc2_test.go
```diff
@@ -666,12 +676,14 @@ func TestEstimateGas(t *testing.T) {
...
 			},
+			/*
 			{
 				name:          "with gasPrice set and without from address",
 				address:       nil,
 				setGasPrice:   true,
-				expectedError: types.NewRPCError(-32000, "gas required exceeds allowance"),
+				expectedError: nil,
 			},
+			*/
```
ARR552: In order to not block this PR more time, I have created this issue to fix this test case https://github.com/0xPolygonHermez/zkevm-node/issues/3557

tclemos (approved):

synchronizer/synchronizer.go
```diff
@@ -835,6 +858,89 @@ func (s *ClientSynchronizer) checkReorg(latestBlock *state.Block) (*state.Block,
...
+			log.Debug("[checkReorg function] => BlockHashParent: ", block.ParentHash)
+			depth++
+			log.Debug("REORG: Looking for the latest correct ethereum block. Depth: ", depth)
+			// Reorg detected. Getting previous block
+			dbTx, err := s.state.BeginStateTransaction(s.ctx)
+			if err != nil {
+				log.Errorf("error creating db transaction to get prevoius blocks")
+				return nil, err
+			}
+			lb, err := s.state.GetPreviousBlock(s.ctx, depth, dbTx)
```
joanestebanr, line 908: I suggest to move this line down closest to the err check

synchronizer/synchronizer.go
```diff
@@ -835,6 +858,89 @@ func (s *ClientSynchronizer) checkReorg(latestBlock *state.Block) (*state.Block,
...
+				if rollbackErr != nil {
+					log.Errorf("error rolling back state. RollbackErr: %v", rollbackErr)
+					return nil, rollbackErr
+				}
+				log.Errorf("error committing dbTx, err: %v", errC)
+				return nil, errC
+			}
+			if errors.Is(err, state.ErrNotFound) {
+				log.Warn("error checking reorg: previous block not found in db: ", err)
+				return &state.Block{}, nil
```
joanestebanr: Have you check that doing a resetState to block 0 works?

ARR552: yes, this has been tested many times. The initial genesis block would be this situation where this is no previous block

synchronizer/synchronizer.go
ARR552: if I do that I need to duplicate the commit twice. Inside the if error and after it

PR #3560, opened 2024-04-15
Synchronizer support Feijoa Part 1

### LINKED ISSUE #3389: Synchronizer must deal with discrepancies in StateRoot when syncing from L2 a closed batch
# Rationale

Currently if `synchronizer` synchronize a batch from TrustedNode and doesn't match StateRoot but is closed it keep syncing from L2, maybe make sense to stop trying and update from L1 to get the Virtualized Batch

It produce next log: 
```
syncTrustedState: batch[57074/57076] mode nothing: error processing trusted batch. Error: when closing the batch, the batch is already close, but  the data on state doesnt match the expected
```

joanestebanr:
Partial #3389 

### General
- Updated Geth to v1.13.14 to support EIP-4844
- Add a consensus client for the retrieve the Blobs (`./beacon_client`)

### etherman
- Add etherman new structure for Feija events (to be able to adapt zkevm-l1-sync library)
- Add support for EIP-4844 (using beacon client)
- Add support to SequenceBlobs event
-Add new config params for consesus client: 
```
[Etherman]
ConsensusL1URL = ""
```
You can use a free quiknode.pro node 

### state
- Add support to blob_sequences (under develpoment)

### synchronizer
- Add initial support to synchronizer to SequenceBlobs (write sequence)

**NOTE**: If initializing the consensus client fails (because some is not configured for instance) the Feijoa code is disable on etherman to keep compability
**NOTE**: Migration 0021.sql still not finish (that is the reason that still doesnt have a test)

### Reviewers

Main reviewers:
- @ARR552 
- @tclemos 
- @agnusmor 
- @ToniRamirezM 

Codeowner reviewers:

- @-Alice
- @-Bob


beacon_client/beacon_client.go
```diff
@@ -0,0 +1,80 @@
...
+
+func unserializeGenericResponse2[T any](response BeaconAPIResponse) (T, error) {
+	var result T
+	err := json.Unmarshal(response.Result, &result)
+	if err != nil {
+		var zero T
+		return zero, err
+	}
+	return result, nil
+}
```
ARR552: Why do we need these two functions?

etherman/feijoa_contracts.go
```diff
@@ -0,0 +1,38 @@
...
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	"github.com/ethereum/go-ethereum/common"
+)
+
+// FeijoaContracts represents the contracts of the Feijoa upgrade
+type FeijoaContracts struct {
+	FeijoaZKEVMAddress common.Address
+	FeijoaZKEVM        *feijoapolygonzkevm.Feijoapolygonzkevm
+	//FeijoaRollupManager         *feijoapolygonrollupmanager.Feijoapolygonrollupmanager
+	//FeijoaGlobalExitRootManager *feijoapolygonzkevmglobalexitroot.Feijoapolygonzkevmglobalexitroot
```
ARR552, line 15: What happens with this contracts?? As far as I know the have also changed

etherman/feijoa_contracts.go
```diff
@@ -0,0 +1,38 @@
...
+}
+
+// NewFeijoaContracts creates a new FeijoaContracts
+func NewFeijoaContracts(ethClient bind.ContractBackend, l1Config L1Config) (*FeijoaContracts, error) {
+	FeijoaZKEVMAddress := l1Config.RollupManagerAddr
+	FeijoaZKEVM, err := feijoapolygonzkevm.NewFeijoapolygonzkevm(FeijoaZKEVMAddress, ethClient)
```
ARR552: I'm not sure about this. Seems it is the wrong address

etherman/feijoa_event_sequence_blobs.go
```diff
@@ -0,0 +1,162 @@
...
+	"github.com/0xPolygonHermez/zkevm-node/log"
+	"github.com/ethereum/go-ethereum/accounts/abi"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/crypto"
+)
+
+const (
+	// SequenceBlobsOrder identifies a SequenceBlobs order
+	SequenceBlobsOrder EventOrder = "SequenceBlobs"
```
ARR552, line 21: Shouldn't we add more new events here?

etherman/feijoa_contracts.go
joanestebanr: You are right, I'll remove the old one

beacon_client/beacon_client.go
joanestebanr: You're right, I remove the usage of old one

etherman/feijoa_contracts.go
joanestebanr: Yes, but is still not done. I put only the ones that I need

etherman/feijoa_event_sequence_blobs.go
joanestebanr: Yes, still pending

etherman/feijoa_contracts.go
joanestebanr: Good catch!, I'll change it

ARR552 (approved):

PR #3631, opened 2024-05-15
eth_call enable ooc check

giskook:
Closes <https://github.com/0xPolygonHermez/zkevm-node/issues/3632>.

### Reviewers

Main reviewers:

- @John
- @Doe

Codeowner reviewers:

- @-Alice
- @-Bob


tclemos:
Hi @giskook, thanks for your contribution.

The `eth_call` doesn't report OOC errors because we understand this method is used to read information from the chain and not to change its state.

We have a custom endpoint called `zkevm_estimateCounters` to help check for OOC; it uses the same parameters as the `eth_call`, but the response is focused on the counters.

PR #3641, opened 2024-05-22
Feature/3640 synchronizer choose to sync from l2

### LINKED ISSUE #3640: Add a paramater to synchronizer to choose if sync from L2 or not
# Rationale

The synchronizer usually sync from L1 and when reach the last sequence batch start syncing from L2 the trusted State. In some case we want to avoid to sync the trusted state

# Implementation

Add a configuration param to choose the behaviour

```
	[Synchronizer.L2Synchronization]
		Enable = true
```

joanestebanr:
Closes #3640

### What does this PR do?

Add a new param `Synchronizer.L2Synchronization.Enable` to choose if the sync from L2 is active

```
	[Synchronizer.L2Synchronization]
		Enable = true
```

### Reviewers

Main reviewers:

- @ARR552 
- @tclemos 
- @agnusmor


synchronizer/synchronizer.go
```diff
@@ -413,7 +413,7 @@ func (s *ClientSynchronizer) Sync() error {
 			// latestSequencedBatchNumber -> last batch on SMC
 			if latestSyncedBatch >= latestSequencedBatchNumber {
 				startTrusted := time.Now()
-				if s.syncTrustedStateExecutor != nil && !s.isTrustedSequencer {
+				if s.syncTrustedStateExecutor != nil && !s.isTrustedSequencer && s.cfg.L2Synchronization.Enable {
```
ARR552: Instead of adding a new condition in the if you can just leave s.syncTrustedStateExecutor as nil because it is not needed. Doing this, you will implement the requirement feature without wasting memory allocating something that it is not needed.

ARR552 (approved):

PR #3659, opened 2024-05-29
Remove sync with virtual state (synchronizer). Add L1 block confirmat…

agnusmor:
### What does this PR do?

- Removes sync with virtual state (synchronizer)
- Adds L1 block confirmations to consider sequence final

New config parameter:
> [SequenceSender]
> SequenceL1BlockConfirmations = 4

### Reviewers

Main reviewers:

@tclemos 
@ToniRamirezM


sequencesender/sequencesender.go
```diff
@@ -96,19 +100,58 @@ func (s *SequenceSender) tryToSendSequence(ctx context.Context) {
...
 		}
+
+		if len(result.Txs) == 1 {
+			var txL1BlockNumber uint64
+			var txHash common.Hash
+			for _, tx := range result.Txs {
+				txL1BlockNumber = tx.Receipt.BlockNumber.Uint64()
+				txHash = tx.Tx.Hash()
+				break
+			}
```
tclemos: I think this code should be like this:
```go
if len(result.Txs) > 0 {
    var txL1BlockNumber uint64
    var txHash common.Hash
    for _, tx := range result.Txs {
        if tx.Receipt != nil {
	        txL1BlockNumber = tx.Receipt.BlockNumber.Uint64()
	        txHash = tx.Tx.Hash()
        }
    }
```

The monitored TX can have multiple TXs. Also, only one of the TXs will have a receipt; the others were discarded in the pool because a new one with a better fee was mined.

sequencesender/sequencesender.go
```diff
@@ -96,19 +100,58 @@ func (s *SequenceSender) tryToSendSequence(ctx context.Context) {
...
+				log.Warnf("failed to get from the SC last sequenced batch number, err: %v", err)
+				return
+			}
+
+			if lastSCBatchNum != s.lastSequenceEndBatch {
+				s.halt(ctx, fmt.Errorf("last sequenced batch from SC %d doesn't match last sequenced batch sent %d", lastSCBatchNum, s.lastSequenceEndBatch))
+			}
+		} else {
+			s.halt(ctx, fmt.Errorf("monitored txs length is %d and should be <= 1", len(result.Txs)))
+		}
```
tclemos: following up the suggestion above, this can also be changed to:

```go
} else {
    s.halt(ctx, fmt.Errorf("monitored tx %v doesn't have transactions to be checked", result.ID))
}
```

sequencesender/sequencesender.go
```diff
@@ -126,7 +169,7 @@ func (s *SequenceSender) tryToSendSequence(ctx context.Context) {
 		return
 	}
 
-	lastVirtualBatchNum, err := s.state.GetLastVirtualBatchNum(ctx, nil)
+	lastVirtualBatchNum, err := s.etherman.GetLatestBatchNumber()
```
tclemos, line 180: It looks like this call is the same as line 131. Maybe we can avoid calling the SC again.

tclemos (approved):

sequencesender/sequencesender.go
agnusmor: Ok, I misunderstand the Txs field, I thought this was a map with the monitored tx for the sequencer-sender, but I have reviewed the code and the ProcessPendingMonitoredTxs is called for each monitored tx of the owner and Txs is the list of L1 txs retries

agnusmor: I have added a "break" when a tx with receipt is found (it should have only one tx with receipt). Also I have put the code that waits for L1 block confirmations inside a "else" block, since that code must be only executed if the monitored tx is confirmed (not failed). Before the code was executed always, also when the tx had failed status. Finally I have added a fatal in case the monitored tx is confirmed and doesn't have any tx with receipt (that shouldn't happen). Please review the update PR

agnusmor: I have just changed the if..else order. The "if" block is executed when the tx is confirmed, the "else" block when it's failed. I think is more "clean" to do in this way now

ToniRamirezM (approved):

PR #3744, opened 2024-07-31
Fix tx OOC (node level) when first empty L2 block in batch

agnusmor:
### What does this PR do?

Fix tx OOC (node level) when first empty L2 block in batch

### Reviewers

Main reviewers:

@tclemos 
@joanestebanr

### DIFF
--- a/pool/config_test.go
+++ b/pool/config_test.go
@@ -1,6 +1,7 @@
 package pool
 
 import (
+	"fmt"
 	"testing"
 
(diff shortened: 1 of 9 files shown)


tclemos (approved):
The solution to the problem seems ok, but the PR is lacking tests.

It would be good to have these scenarios:
- OOC for the first empty block
- OOC for any other block than the first

both checking the TX was invalidated and removed from the worker

I'll keep it approved in case we need this, but if the fix is not urgent, I recommend adding those tests.

PR #3750, opened 2024-08-05
Fix timestamp when querying zkevm_getBatchByNumer and the batch is only in trusted state

agnusmor:
### What does this PR do?

When querying zkevm_getBatchByNumber endpoint and the batch requested was only in trusted state (not sequenced/virtualized) the timestamp returned in the response for the batch was wrong. The function assigned a "nil" value for the timestamp and this was intrepreted as a very high timestamp value in the json response (future timestamp).

Now the issue is fixed and in the case the batch is only in trusted state it returns the timestamp of the last L2 block in the batch.

### Reviewers

Main reviewers:

@ToniRamirezM 
@tclemos 
@ARR552

### DIFF
--- a/state/batch.go
+++ b/state/batch.go
@@ -568,14 +568,11 @@ func (s *State) GetLastBatch(ctx context.Context, dbTx pgx.Tx) (*Batch, error) {
 	return batches[0], nil
 }
 
-// GetBatchTimestamp returns the batch timestamp.
-// if batch >= etrog
-//
-//	if the batch is virtualized it will return virtual_batch.timestamp value
-//	if the batch is not virtualized (only in trusted state) and the batch doesn't have L2 blocks it will return batch.timestamp value
-//	if the batch has blocks it will return the timestamp of the last L2 block of the batch
-//
-// if batch < etrog it will return batch.timestamp value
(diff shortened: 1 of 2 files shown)


ARR552:
What would happened if the batch is already virtualized?? In this case if you use the timestamp of the last l2 block in the batch the value would be lower than the virtualization time

ARR552 (approved):
LGTM

ARR552 (approved):

Booss13 (commented):
[]()

Booss13:
> ### What does this PR do?
> 
> 
> 
> When querying zkevm_getBatchByNumber endpoint and the batch requested was only in trusted state (not sequenced/virtualized) the timestamp returned in the response for the batch was wrong. The function assigned a "nil" value for the timestamp and this was intrepreted as a very high timestamp value in the json response (future timestamp).
> 
> 
> 
> Now the issue is fixed and in the case the batch is only in trusted state it returns the timestamp of the last L2 block in the batch.
> 
> 
> 
> ### Reviewers
> 
> 
> 
> Main reviewers:
> 
> 
> 
> @ToniRamirezM 
> 
> @tclemos 
> 
> @ARR552

PR #3751, opened 2024-08-06
add l2blockend to DS

ToniRamirezM:
### What does this PR do?

Adds l2blockend to DS

### Reviewers

Main reviewers:

- @agnusmor


state/datastream.go
```diff
@@ -531,6 +588,10 @@ func GenerateDataStreamFile(ctx context.Context, streamServer *datastreamer.Stre
 						BlockGasLimit:   l2Block.BlockGasLimit,
 					}
 
+					streamL2BlockEnd := &datastream.L2BlockEnd{
```
agnusmor: Maybe this code can be added inside the if statement in line 691, since streamL2BlockEnd is used only if ds version >= 4

agnusmor (approved):

PR #3762, opened 2024-08-12
Fix getfilterchanges return value

### LINKED ISSUE #3686: eth_getFilterLogs returns empty may cause NullPointerException (NPE) in web3j
# Rationale

From user's feedback, we have found that there is a difference between Erigon and zkevm-node when querying events with `eth_getFilterLogs` RPC method. When the RPC cannot find the filter ID from user in its own memory, Erigon will return a `[]` while the zkevm-node will return `null`.  But `null` leads to a NullPointerException (NPE) in web3j and thus causes a crash on user's program.  So maybe zkevm-node should change its return from null to an empty slice when nothing returns and no error occors, and the RPC query response will be like the following: 

```
{
    "jsonrpc": "2.0",
    "id": 73,
    "result": []
}
```

# Implementation

We modified `jsonrpc/endpoints_eth.go` Line 493 & 499 into `return []types.Log{}, nil`, and it works.

giskook:
Closes https://github.com/0xPolygonHermez/zkevm-node/issues/3686

### Reviewers

Main reviewers:

- @John
- @Doe

Codeowner reviewers:

- @-Alice
- @-Bob

### DIFF
--- a/jsonrpc/endpoints_eth.go
+++ b/jsonrpc/endpoints_eth.go
@@ -434,7 +434,7 @@ func (e *EthEndpoints) GetFilterChanges(filterID string) (interface{}, types.Err
 				return nil, rpcErr
 			}
 			if len(res) == 0 {
-				return nil, nil
+				return []common.Hash{}, nil
 			}
 			return res, nil
 		}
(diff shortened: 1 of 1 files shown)


tclemos:
hi @giskook, thanks for your contribution.
It looks like one of the tests is failing due to the change you proposed. Would you mind checking it?

PR #3774, opened 2024-08-22
Synchronization of L1InfoTree events before genesis allow to resume

joanestebanr:
Closes #<issue number>.

### What does this PR do?

 - Synchronization of L1InfoTree events before genesis allow to resume 
 - If genesis `RollupManagerBlockNumber` param is set 0 autodiscover LxLy starting block

### Reviewers

Main reviewers:

- @arnaubennassar


joanestebanr:
This PR is for cdk-validium-node

PR #3824, opened 2025-01-17
Add archive notice with replacement information

ToniRamirezM:
Adds deprecation notice with replacement information


eduadiez (approved):
LGTM

